PR #71, opened 2025-03-08
update prism-react-renderer version

badmonster0:


docs/docusaurus.config.ts
```diff
@@ -18,7 +18,7 @@ const config: Config = {
   // GitHub pages deployment config.
   // If you aren't using GitHub pages, you don't need these.
   organizationName: 'cocoindex-io', // Usually your GitHub org/user name.
-  projectName: 'docs', // Usually your repo name.
+  projectName: 'cocoindex', // Usually your repo name.
```
badmonster0: what's the purpose of changing this?

we need to deploy through the "docs" project.

PR #182, opened 2025-03-20
chore: Clippy fixes

Anush008:
## Description

This PR applies a bunch a bunch of Clippy fixes I found when working with the project. No change in behaviour.


badmonster0:
Thanks for the fix!!

PR #183, opened 2025-03-20
Fix more clippy warnings.

badmonster0:


badmonster0:
Thanks for the PR Anush 🤗! @Anush008

PR #205, opened 2025-03-24
feat: Qdrant storage support

Anush008:
## Descrption

This PR adds support to export data to a Qdrant - https://qdrant.tech/ collection.

To run Qdrant, 
```
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

6333 - REST API, 6334 - gRPC API.
The Rust client calls the gRPC API.

You can find a dashboard at `https://localhost:6333/dashboard`.


src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        self.client
+            .upsert_points(UpsertPointsBuilder::new(&self.collection_name, points))
+            .await?;
+        Ok(())
+    }
+}
+
+fn parse_key_fields(
```
anush008: I am testing out with different field types.

docs/docs/ops/storages.md
```diff
@@ -11,6 +11,16 @@ description: CocoIndex Built-in Storages
 
 The spec takes the following fields:
 
-*   `database_url` (type: `str`, optional): The URL of the Postgres database to use as the internal storage, e.g. `postgres://cocoindex:cocoindex@localhost/cocoindex`. If unspecified, will use the same database as the [internal storage](/docs/core/basics#internal-storage).
+* `database_url` (type: `str`, optional): The URL of the Postgres database to use as the internal storage, e.g. `postgres://cocoindex:cocoindex@localhost/cocoindex`. If unspecified, will use the same database as the [internal storage](/docs/core/basics#internal-storage).
```
badmonster0: Prefer undoing this indent change. 

We're following this style guide and doing 4-space indent consistently: https://google.github.io/styleguide/docguide/style.html#nested-list-spacing

docs/docs/ops/storages.md
```diff
@@ -11,6 +11,16 @@ description: CocoIndex Built-in Storages
...
-*   `table_name` (type: `str`, optional): The name of the table to store to. If unspecified, will generate a new automatically. We recommend specifying a name explicitly if you want to directly query the table. It can be omitted if you want to use CocoIndex's query handlers to query the table.
+* `table_name` (type: `str`, optional): The name of the table to store to. If unspecified, will generate a new automatically. We recommend specifying a name explicitly if you want to directly query the table. It can be omitted if you want to use CocoIndex's query handlers to query the table.
+
+## Qdrant
+
+`Qdrant` exports data to a [Qdrant](https://qdrant.tech/) collection.
+
+The spec takes the following fields:
+
+* `qdrant_url` (type: `str`, required): The [gRPC URL](https://qdrant.tech/documentation/interfaces/#grpc-interface) of the Qdrant instance. Defaults to <http://localhost:6334/>.
```
badmonster0: There's a build failure for this doc:

https://github.com/cocoindex-io/cocoindex/actions/runs/14029768057/job/39274718806?pr=205

Seems caused by this line. `<http://localhost:6334/>` isn't valid MDX.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+                    .get(field_name)
+                    .and_then(|v| v.as_double().map(BasicValue::Float64)),
+
+                BasicValueType::Json => point
+                    .payload
+                    .get(field_name)
+                    .map(|v| BasicValue::Json(Arc::from(v.clone().into_json()))),
+
+                BasicValueType::Vector(_) => {
+                    let vectors_options = point.vectors.clone().unwrap().vectors_options.unwrap();
```
badmonster0: `unwrap()` may panic. Prefer returning an error (try `.ok_or_else(|| ...)?` for `Option`) (unless a failure indicates some very bad unrecoverable error)

src/ops/storages/qdrant.rs
badmonster0: The function name isn't very clear about what it's doing. Maybe something like `key_values_to_qdrant_fields`?

Similar to `parse_value_fields` below.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+}
+
+#[async_trait]
+impl ExportTargetExecutor for Executor {
+    async fn apply_mutation(&self, mutation: ExportTargetMutation) -> Result<()> {
+        let mut points: Vec<PointStruct> = Vec::with_capacity(mutation.upserts.len());
+        for upsert in mutation.upserts.iter() {
+            let key_fields = key_value_fields_iter(&self.key_fields_schema, &upsert.key)?
+                .iter()
+                .collect();
```
badmonster0: This `.iter().collect()` seems redundant. We can directly pass a slice (in type `&[KeyValue]`) to `parse_key_fields`.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+            let key_fields = parse_key_fields(&key_fields, &self.key_fields_schema)?;
+            let (mut payload, vectors) =
+                parse_value_fields(&upsert.value.fields, &self.value_fields_schema)?;
+            payload.extend(key_fields);
```
badmonster0: Here the key fields is treated as a regular value fields, so they're no longer keys.

We distinguished key/value because keys can be used to uniquely identify a row ("Point" in Qdrant's concept) in the storage - and it's important for update and delete.

After going through Qdrant's documents, it looks like Qdrant is using Point ID for this purpose. Qdrant allows a point ID to be 64-bit unsigned integers or UUID.

So here's probably one way to handle it:

- For Qdrant storage, only allows users to specify field with compatible types as key (they're specified by users like [this](https://github.com/cocoindex-io/cocoindex/blob/c3a7e50b5aadcbacacaf16db76ff5fd7882ad8b4/examples/text_embedding/main.py#L37))
- The `StorageFactoryBase::build` method should return an error if the primary key has incompatible type.
- We will add the following support to the cocoindex framework (soon):
  - Support UUID as a basic type (Postgres also supports it as a type, so it makes sense)
  - Support auto-generated UUID, i.e. in places like [this](https://github.com/cocoindex-io/cocoindex/blob/c3a7e50b5aadcbacacaf16db76ff5fd7882ad8b4/examples/text_embedding/main.py#L37)) users can ask the framework to auto-generate an UUID as key, and the framework will remember it and use it during updates/deletion.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+    payload.insert(field_name.to_string(), QdrantValue { kind: Some(kind) });
+}
+
+fn convert_to_vector(v: Vec<BasicValue>) -> Vec<f32> {
+    v.iter()
+        .filter_map(|elem| match elem {
+            BasicValue::Float32(f) => Some(*f),
+            BasicValue::Float64(f) => Some(*f as f32),
+            BasicValue::Int64(i) => Some(*i as f32),
+            _ => None,
```
badmonster0, line 161: Instead of silently dropping non-number elements (which may hide real issues), some alternatives are:

-   We can decide this qdrant storage only supports vector with these types. Hence we can raise an error (by `api_bail!` or `api_error!` which is intended for errors should be resolved by users) in `StorageFactoryBase()` if we see any type unsupported by qdrant storage. Then, here we can raise an error (by `bail!`, as it's no longer expected here unless bug of the cocoindex framework)

-   We can also decide to support it, then we can encode it in some way similar to how we deal with JSON. We did the similar thing for the `Postgres` storage.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+                        VectorsOptions::Vector(vector) => {
+                            let x = vector
+                                .data
+                                .into_iter()
+                                .map(BasicValue::Float32)
+                                .collect::<Vec<_>>();
+                            Some(BasicValue::Vector(Arc::from(x)))
+                        }
+                        VectorsOptions::Vectors(vectors) => {
+                            let vector = vectors.vectors[field_name].clone();
```
badmonster0: Can we avoid a `clone` here? (change `into_iter()` to `iter()` below?)

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,432 @@
...
+
+                BasicValueType::Float64 => point
+                    .payload
+                    .get(field_name)
+                    .and_then(|v| v.as_double().map(BasicValue::Float64)),
+
+                BasicValueType::Json => point
+                    .payload
+                    .get(field_name)
+                    .map(|v| BasicValue::Json(Arc::from(v.clone().into_json()))),
```
badmonster0, line 199: nit: If this function takes `ScoredPoint` instead of `&ScoredPoint` as argument type, I think we should be able to avoid a `clone` here.

badmonster0 (commented):
Thanks for adding Qdrant storage support!

src/ops/storages/qdrant.rs
badmonster0: #206 for UUID

src/ops/storages/qdrant.rs
anush008: `cannot move out of index of `HashMap<std::string::String, VectorOutput>`
move occurs because value has type `VectorOutput`, which does not implement the `Copy` traitrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/anush/Desktop/cocoindex/src/ops/storages/qdrant.rs)`

anush008:
Payload conversion is simpler now.
Setup status check is WIP.

anush008:
Hey @badmonster0.

The fields specified under `primary_key_fields` are only sent as key fields and the rest as value fields. Correct?
So in case of Qdrant, we try to parse the key fields as UUIDs or Integers to be used as point IDs?

badmonster0:
> Hey @badmonster0.
> 
> The fields specified under `primary_key_fields` are only sent as key fields and the rest as value fields. Correct? So in case of Qdrant, we try to parse the key fields as UUIDs or Integers to be used as point IDs?

Correct. "key fields" and "value fields" are exclusive.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,400 @@
...
+        (CollectionId, SetupState),
+        ExecutorFuture<'static, (Arc<dyn ExportTargetExecutor>, Option<Arc<dyn QueryTarget>>)>,
+    )> {
+        // TODO(Anush008): Add as a field to the Spec
+        let url = "http://localhost:6334/";
+        let collection_name = spec.collection_name;
+        let table_id = CollectionId {
+            collection_name: collection_name.to_owned(),
+        };
+        let setup_state = SetupState {};
```
badmonster0: Hint: When building `SetupState`, you may put schema information closer to Qdrant's native representation there.

When you convert `key_fields_schema` and `value_fields_schema` to `SetupState`, feel free to raise an error if a type cannot be supported by Qdrant.

Then in `SetupStatusCheck` you'll more easily create/update Qdrant collections using these states.

badmonster0:
Just a warm checkin to see if there's any questions during implementing. Feel free to ask if there's any :)

It looks like the executor logic is almost there. The setup status checker is usually harder. We're also considering making this part optional (e.g. users can manage the backend resources themselves outside cocoindex, as long as the schema they created matches certain expectations). If there're indeed difficulty in this part and you prefer, I can make this change.

Thanks!

badmonster0 (commented):
Just a warm checkin to see if there's any questions during implementing. Feel free to ask if there's any :)

It looks like the executor logic is almost there. The setup status checker is usually harder. We're also considering making this part optional (e.g. users can manage the backend resources themselves outside cocoindex, as long as the schema they created matches certain expectations). If there're indeed difficulty in this part and you prefer, I can make this change.

Thanks!

anush008:
Hey. I'm on a vacation.

> I can make this change

Sure. Thank you 🙏

badmonster0:
> Hey. I'm on a vacation.
> 
> > I can make this change
> 
> Sure. Thank you 🙏

Added a `setup_by_user` option for this purpose: https://cocoindex.io/docs/core/flow_def#export

Then when implementing [`check_setup_status`](https://github.com/cocoindex-io/cocoindex/blob/390332cc112d81faf2ca5233be3201646161f069/src/ops/factory_bases.rs#L291-L298) you can directly return an error (also just implemented `Infallible` for `ResourceSetupStatusCheck` which provides a concrete type that you can return)

Hope this makes the integration of Qdrant easier by separating into different steps :)

anush008:
`ResourceSetupStatusCheck` is handled now.
With https://github.com/cocoindex-io/cocoindex/pull/205/commits/4a0bdf5b066945fdf1a5d0f9ccb8cc4bf1698e3c and https://github.com/cocoindex-io/cocoindex/pull/205/commits/a4c6cae1fed7a2774bded749eaa068a74d5cf9d5

anush008:
Hey @badmonster0! Am I right in my understanding that `primary_key_fields` are passed as `key_fields`? If so, would it be sufficient to treat the first key field as the ID when using Qdrant?

https://github.com/cocoindex-io/cocoindex/pull/205/commits/67f1e714daaeb5659a9b54de84c72be94af433c9

badmonster0:
> Hey @badmonster0! Am I right in my understanding that `primary_key_fields` are passed as `key_fields`? 

Yes, that's correct.


> If so, would it be sufficient to treat the first key field as the ID when using Qdrant?

Yes, this is the right approach in my mind.
`primary_key_fields` is allowed to have multiple elements as storages like Postgres support primary key on multiple fields. 
Since qdrant only support a single key part, the size should always be 1. And we can raise an error (like `api_bail!`) when `StorageFactoryBase::build` is called.

Anush008:
Hi @badmonster0 👋. This should be it.

You can now try the example at https://github.com/Anush008/cocoindex/tree/qdrant/examples/text_embedding.

anush008:
Hi @badmonster0 👋. This should be it.

You can now try the example at https://github.com/Anush008/cocoindex/tree/qdrant/examples/text_embedding_qdrant.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,393 @@
...
+fn key_to_point_id(key_values: &[KeyValue]) -> Result<PointId> {
+    let point_id = if let Some(key_value) = key_values.first() {
+        match key_value {
+            KeyValue::Str(v) => PointId::from(v.to_string()),
+            KeyValue::Int64(v) => PointId::from(*v as u64),
+            KeyValue::Uuid(v) => PointId::from(v.to_string()),
+            _ => bail!("Unsupported Qdrant Point ID key type"),
+        }
+    } else {
+        let uuid = Uuid::new_v4().to_string();
```
badmonster0: This will have a problem though: the cocoindex framework is unable to track which rows written to Qdrant is based on each source row, so that once the source row gets updated/deleted, cocoindex is unable to update/delete the corresponding rows in Qdrant, which will leak the data there.

I think it's good to assume primary key fields exist, and raise an error in the factory otherwise. (when we defined the interface we made it optional, as we thought about the framework should automatically create a unique key. It's not implemented yet. After if it's implemented, the framework should automatically add this before calling the factory, so a specific storage's implementation should always see at least one key field)

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,393 @@
...
+            let point_id = key_to_point_id(key_fields)?;
+            let (payload, vectors) =
+                values_to_payload(&upsert.value.fields, &self.value_fields_schema)?;
+
+            points.push(PointStruct::new(point_id, vectors, payload));
+        }
+
+        self.client
+            .upsert_points(UpsertPointsBuilder::new(&self.collection_name, points).wait(true))
+            .await?;
```
badmonster0, line 79: We also want to delete points as specified by `mutation.delete_keys`

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,393 @@
...
+                        }
+                    }),
+
+                BasicValueType::Date
+                | BasicValueType::LocalDateTime
+                | BasicValueType::OffsetDateTime
+                | BasicValueType::Time
+                | BasicValueType::Uuid => point.payload.get(field_name).and_then(|v| {
+                    v.as_str()
+                        .map(|s| BasicValue::Str(Arc::from(s.to_string())))
```
badmonster0: These should have specific value types as indicated by the enum of `BasicValueType`.

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,393 @@
...
+use qdrant_client::qdrant::{
+    NamedVectors, PointId, PointStruct, UpsertPointsBuilder, Value as QdrantValue,
+};
+use qdrant_client::qdrant::{Query, QueryPointsBuilder, ScoredPoint};
+use qdrant_client::Qdrant;
+use serde::Serialize;
+use serde_json::json;
+use uuid::Uuid;
+
+fn key_value_fields_iter<'a>(
```
badmonster0: Recently we added a new method, which can replace this one:

https://github.com/cocoindex-io/cocoindex/blob/e8a6290d5bbd0430d24760c6cfa69508a2659ac7/src/base/value.rs#L177

examples/text_embedding/main.py
badmonster0, line 1: How about creating a new example directory, e.g. `text_embedding_qdrant`. We can keep the original example, and also a dedicated example for qdrant makes the support for qdrant more obvious.

src/ops/storages/qdrant.rs
anush008: Qdrant accepts payload as JSON, so there aren't any other types.

badmonster0: This function's input is a value from Qdrant, and output is a value for cocoindex.

I think what we need here is the opposite of these cocoindex value -> string conversions:

https://github.com/Anush008/cocoindex/blob/aa2acda58b1779e844fdb51c3b966fa170766c1b/src/ops/storages/qdrant.rs#L138-L142

Similar to the branch below for `Range` which parses JSON values to cocoindex `Range` value:

https://github.com/Anush008/cocoindex/blob/aa2acda58b1779e844fdb51c3b966fa170766c1b/src/ops/storages/qdrant.rs#L241-L254

anush008: Got it. Will do.

examples/text_embedding/main.py
anush008: Done with https://github.com/cocoindex-io/cocoindex/pull/205/commits/86fc14c5368a726e8b88614362e5ac7cc5e0622d and https://github.com/cocoindex-io/cocoindex/pull/205/commits/f749375b90ace2e932263a9e92e3264b5a66c49b.

src/ops/storages/qdrant.rs
anush008: Didn't need that too. Since we allow only one key field for Qdrant.

https://github.com/cocoindex-io/cocoindex/pull/205/commits/67e3608248ae37e5e5d2f4e9b212a53382d9b2d6

src/ops/storages/qdrant.rs
anush008: Resolved with https://github.com/cocoindex-io/cocoindex/pull/205/commits/67e3608248ae37e5e5d2f4e9b212a53382d9b2d6

src/ops/storages/qdrant.rs
anush008: Done with https://github.com/cocoindex-io/cocoindex/pull/205/commits/d9c93d4d54cffa3d97dcc3aa05ddb66fa27ec208.

src/ops/storages/qdrant.rs
anush008: Done with https://github.com/cocoindex-io/cocoindex/pull/205/commits/32d94f74d0ccb43387c92d1d10e246221f792d85

src/ops/storages/qdrant.rs
```diff
@@ -0,0 +1,430 @@
...
+            Value::Basic(basic_value) => {
+                let json_value: serde_json::Value = match basic_value {
+                    BasicValue::Bytes(v) => String::from_utf8_lossy(v).into(),
+                    BasicValue::Str(v) => v.clone().to_string().into(),
+                    BasicValue::Bool(v) => (*v).into(),
+                    BasicValue::Int64(v) => (*v).into(),
+                    BasicValue::Float32(v) => (*v as f64).into(),
+                    BasicValue::Float64(v) => (*v).into(),
+                    BasicValue::Range(v) => json!({ "start": v.start, "end": v.end }),
+                    BasicValue::Uuid(v) => json!({ "uuid": v.to_string() }),
```
badmonster0: Personally I think we don't have to have this wrapper. Directly putting the string for the uuid should be good enough - it's the more natural way to represent a uuid in a database, and will make the downstream easier to consume.

The same for `Date`, `LocalDateTime`, `Time` and `OffsetDateTime` below.

(`Range` is different as it naturally has two parts)

anush008: Addressed in https://github.com/cocoindex-io/cocoindex/pull/205/commits/19dee3e1180180efb71cafa4dc8ccc3ef23ad82b.

badmonster0 (approved):
Awesome! Thank you very much!

PR #222, opened 2025-03-27
Update contributing.md with instructions to build cocoindex

cocoindex-dev:


docs/docs/about/contributing.md
```diff
@@ -22,7 +22,7 @@ We love contributions from our community! This guide explains how to get involve
     pip install maturin
     ```
 
--   Build the library
+-   Build the library. In the same virtual env, run at the root of cocoindex directory
```
badmonster0: "In the same virtual env" is also applied to "Install maturin". 

Personally I think using the consistent virtual env is the basic. If we really need to emphasize it, maybe we can put it under the bullet above for creating virtual env (also note that virtual env is optional any way), e.g. we can say something like "make sure buildings / runnings all happen in the same virtual env"

docs/docs/about/contributing.md
```diff
@@ -11,7 +11,7 @@ We love contributions from our community! This guide explains how to get involve
 
 -   Install Rust toolchain: [docs](https://rust-lang.org/tools/install)
 
--   (Optional) Setup and activate python virtual environment
+-   (Optional) Setup and activate python virtual environment. Make sure buildings / runnings all happen in the same virtual env
```
badmonster0: The second sentence is specifically for the second command (`activate`). So I think probably we can make it more clear like this:

-   (Optional) Setup Python virtual environment.

    ```bash
    virtualenv --python=$(which python3.12) .venv
    ```

    Activate the virtual environment, before any installings / buildings / runnings:

    ```bash
    . .venv/bin/activate
    ```

docs/docs/about/contributing.md
```diff
@@ -22,7 +22,7 @@ We love contributions from our community! This guide explains how to get involve
     pip install maturin
     ```
 
--   Build the library
+-   Build the library. Run at the root of cocoindex directory.
```
badmonster0: nit: s/./:/ for the end period.

badmonster0 (approved):

PR #242, opened 2025-04-01
configure algolia doc search

badmonster0:


docs/docusaurus.config.ts
```diff
@@ -4,6 +4,8 @@ import type { Config } from '@docusaurus/types';
 import type * as Preset from '@docusaurus/preset-classic';
 
 // This runs in Node.js - Don't use client-side code here (browser APIs, JSX...)
+const COCOINDEX_DOCS_ALGOLIA_API_KEY = process.env.COCOINDEX_DOCS_ALGOLIA_API_KEY;
+const COCOINDEX_DOCS_ALGOLIA_APP_ID = process.env.COCOINDEX_DOCS_ALGOLIA_APP_ID;
```
badmonster0: Can we move these closer to where they're actually used?

(BTW, I'll also even consider removing these two local variables)

docs/docusaurus.config.ts
```diff
@@ -164,4 +166,15 @@ if (!!process.env.COCOINDEX_DOCS_POSTHOG_API_KEY) {
   ]);
 }
 
+
+if (!!COCOINDEX_DOCS_ALGOLIA_API_KEY) {
```
badmonster0: We should also check `COCOINDEX_DOCS_ALGOLIA_APP_ID` here.

PR #305, opened 2025-04-14
chore: Some Clippy fixes

Anush008:
## Description

As the title goes.

Doesn't include fixes for `warning: very complex type used. Consider factoring parts into `type` definitions`.


badmonster0:
Thanks a lot!

Issue #184, opened 2025-03-21
Implement Unit Tests for Python Native to Engine Format Conversion


badmonster0:
## Enhancement
Add Unit test to [convert.py](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/convert.py) for `to_engine_value`

## Objective: 
Develop unit tests to validate the functionality of the to_engine_value function, ensuring it accurately converts Python native data types into the engine-compatible format.​

## Background:
- This module provide conversions between two data formats:
  - Values in Python native data types
  - Values in data structure exchangeable with cocoindex core engine (in Rust)
- Related [issue](https://github.com/cocoindex-io/cocoindex/issues/185) for conversion from Engine to Python Format

The key difference is about how they handle [Struct type](https://cocoindex.io/docs/core/data_types#struct-type)
- In Python native data formats, it's represented by a `dataclass` (see [the doc](https://cocoindex.io/docs/core/data_types#struct-type)). Example: `Order(order_id='O123', name='mixed nuts', price=25.0)`
 
- In the exchangeable format with the core engine, it's a list of all fields. Example: `['O123', 'mixed nuts', 25.0]`

## Scope:
- The test need to cover different types:
  - Various basic types, struct types and collection types as listed in [the doc](https://cocoindex.io/docs/core/data_types#basic-types)
  - Composition of various types, e.g. list of struct, struct with list fields, struct with struct fields

badmonster0:
thanks @par4m for the contribution! closing

PR #315, opened 2025-04-16
Add unit tests for Python to Engine format conversion

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/184


python/cocoindex/tests/test_convert.py
```diff
@@ -0,0 +1,62 @@
...
+    u = uuid.uuid4()
+    assert to_engine_value(u) == u.bytes
+
+def test_to_engine_value_date_time():
+    d = datetime.date(2024, 1, 1)
+    t = datetime.time(12, 30)
+    dt = datetime.datetime(2024, 1, 1, 12, 30)
+    assert to_engine_value(d) == d
+    assert to_engine_value(t) == t
+    assert to_engine_value(dt) == dt
```
badmonster0, line 39: nit: We may interleave each assignments with assert (i.e. `assert to_engine_value(d) == d` follows `d = datetime.date(2024, 1, 1)` directly). It makes define/use of each variables closer.

python/cocoindex/tests/test_convert.py
```diff
@@ -0,0 +1,62 @@
...
+    assert to_engine_value(basket) == [["apple", "banana"]]
+
+def test_to_engine_value_nested_struct():
+    customer = Customer(name="Alice", order=Order("O1", "item1", 10.0))
+    assert to_engine_value(customer) == ["Alice", ["O1", "item1", 10.0]]
+
+def test_to_engine_value_empty_list():
+    assert to_engine_value([]) == []
+    assert to_engine_value([[]]) == [[]]
+
```
badmonster0, line 60: Will be great if also adding a test case for tuple, including empty tuple (result should be similar to list)

badmonster0 (commented):
Thank you!

python/cocoindex/tests/test_convert.py
par4m: Done! Let me know if there’s anything else you'd like me to adjust. Thanks!

badmonster0 (approved):
Awesome! Thanks you very much!

Issue #185, opened 2025-03-21
Implement Unit Tests for Engine to Python Native Format Conversion


badmonster0:
## Enhancement
Add Unit test to [convert.py](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/convert.py) for `make_engine_value_converter`

## Objective: 
Develop unit tests to validate the functionality of the make_engine_value_converter function, ensuring it accurately converts engine-compatible data structures back into Python native data types.

## Background:
This module provide conversions between two data formats:
  - Values in Python native data types
  - Values in data structure exchangeable with cocoindex core engine (in Rust)
- Related [issue](https://github.com/cocoindex-io/cocoindex/issues/184) for conversion from Python to Engine Format

The key difference is about how they handle [Struct type](https://cocoindex.io/docs/core/data_types#struct-type)
- In Python native data formats, it's represented by a `dataclass` (see [the doc](https://cocoindex.io/docs/core/data_types#struct-type)). Example: `Order(order_id='O123', name='mixed nuts', price=25.0)`
- In the exchangeable format with the core engine, it's a list of all fields. Example: `['O123', 'mixed nuts', 25.0]`

- In the engine->Python native converter, it's possible that a Struct type in engine and Python native have different fields (can be: Python dataclass has more fields, has less fields, has fields in different orders). We should initialize all matched fields (by field names, not order), and leave unmatched fields with their default values.

## Scope:
- The test need to cover different types:
  - Various basic types, struct types and collection types as listed in [the doc](https://cocoindex.io/docs/core/data_types#basic-types)
  - Composition of various types, e.g. list of struct, struct with list fields, struct with struct fields

Hint: `make_engine_value_converter()` also needs to take an engine data type (in `dict[str, Any]` type) as argument. A convenient way to construct it is calling [`typing.encode_enriched_type()`](https://github.com/cocoindex-io/cocoindex/blob/ffcd21522537227d35430666e0fa3fb3c2001cfa/python/cocoindex/typing.py#L193), and get the `'type'` in the result (i.e. `['type']`).

badmonster0:
Thanks @par4m for this improvement!

PR #328, opened 2025-04-17
Add unit tests for Engine to Python format conversion

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/185

### DIFF
--- a/python/cocoindex/tests/test_convert.py
+++ b/python/cocoindex/tests/test_convert.py
@@ -3,13 +3,19 @@ import uuid
 import datetime
 from dataclasses import dataclass
 import pytest
-from cocoindex.convert import to_engine_value
+from cocoindex.convert import to_engine_value, make_engine_value_converter
+from cocoindex.typing import encode_enriched_type
 
 @dataclass
 class Order:
     order_id: str
     name: str
     price: float
+    extra_field: str = "default_extra"
+
+@dataclass
+class Tag:
+    name: str
 
 @dataclass
 class Basket:
@@ -19,6 +25,13 @@ class Basket:
 class Customer:
     name: str
     order: Order
+    tags: list[Tag] = None
+
+@dataclass
+class NestedStruct:
+    customer: Customer
+    orders: list[Order]
+    count: int = 0
 
 def test_to_engine_value_basic_types():
     assert to_engine_value(123) == 123
@@ -40,11 +53,11 @@ def test_to_engine_value_date_time_types():
 
 def test_to_engine_value_struct():
     order = Order(order_id="O123", name="mixed nuts", price=25.0)
-    assert to_engine_value(order) == ["O123", "mixed nuts", 25.0]
+    assert to_engine_value(order) == ["O123", "mixed nuts", 25.0, "default_extra"]
 
 def test_to_engine_value_list_of_structs():
     orders = [Order("O1", "item1", 10.0), Order("O2", "item2", 20.0)]
-    assert to_engine_value(orders) == [["O1", "item1", 10.0], ["O2", "item2", 20.0]]
+    assert to_engine_value(orders) == [["O1", "item1", 10.0, "default_extra"], ["O2", "item2", 20.0, "default_extra"]]
 
 def test_to_engine_value_struct_with_list():
     basket = Basket(items=["apple", "banana"])
@@ -52,7 +65,7 @@ def test_to_engine_value_struct_with_list():
 
 def test_to_engine_value_nested_struct():
     customer = Customer(name="Alice", order=Order("O1", "item1", 10.0))
-    assert to_engine_value(customer) == ["Alice", ["O1", "item1", 10.0]]
+    assert to_engine_value(customer) == ["Alice", ["O1", "item1", 10.0, "default_extra"], None]
 
 def test_to_engine_value_empty_list():
     assert to_engine_value([]) == []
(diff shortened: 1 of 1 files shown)


python/cocoindex/tests/test_convert.py
```diff
@@ -67,3 +80,81 @@ def test_to_engine_value_tuple():
...
+        (bool, True),
+        # (type(None), None),  # Removed unsupported NoneType
+    ]:
+        engine_type = encode_enriched_type(py_type)["type"]
+        converter = make_engine_value_converter([], engine_type, py_type)
+        assert converter(value) == value
+
+def test_make_engine_value_converter_struct():
+    engine_type = encode_enriched_type(Order)["type"]
+    converter = make_engine_value_converter([], engine_type, Order)
```
badmonster0: For all these tests, we're calling `encode_enriched_type()` and `make_engine_value_converter()` in sequence. It'll be great if we wrap these two to a function and call this to build converter for test. It will make each test slightly easier to read.

python/cocoindex/tests/test_convert.py
```diff
@@ -19,6 +25,21 @@ class Basket:
...
     order: Order
+    tags: list[Tag] = None
+
+@dataclass
+class NestedStruct:
+    customer: Customer
+    orders: list[Order]
+    count: int = 0
+
+def build_converter(py_type, target_type=None):
```
badmonster0: The first argument `py_type` is actually used to simulate type on the engine side, so something like `engine_type_in_py` maybe more clear.

python/cocoindex/tests/test_convert.py
```diff
@@ -67,3 +88,74 @@ def test_to_engine_value_tuple():
...
+        (float, 3.14),
+        (str, "hello"),
+        (bool, True),
+        # (type(None), None),  # Removed unsupported NoneType
+    ]:
+        converter = build_converter(py_type)
+        assert converter(value) == value
+
+def test_make_engine_value_converter_struct():
+    converter = build_converter(Order)
```
badmonster0: For several cases, like extra fields, missing fields, order difference, the actual reason to trigger these case is: engine side has a slightly different schema from Python type definitions (the schema from engine is determined based on source/function output types + flow definitions).

To simulate such situations in this test, we may need to define separate python types and pass two different types to `build_converter` (one to simulate type on engine side, one is for type on Python side, e.g. to cover the case of different field orders, they can be two different dataclasses with different field orders).

python/cocoindex/tests/test_convert.py
badmonster0: The name `build_engine_value_converter` may be more clear.

python/cocoindex/tests/test_convert.py
```diff
@@ -153,9 +154,66 @@ def test_make_engine_value_converter_collections():
...
+    """
+    @dataclass
+    class EngineOrder:
+        id: str
+        name: str
+    @dataclass
+    class PythonOrder:
+        id: str
+        name: str
+        price: float = 0.0
```
badmonster0: The extra field and missing field cases are not limited to the fields at last. If we can make the extra / missing field in the middle, the test will catch more potential problems.

par4m: There's a decorator in pytest which can be used for this :- 
```python

@pytest.mark.parametrize(
    "engine_fields, python_fields, python_defaults, engine_val, expected_python_val",
    [
        # Extra field - middle
        (
            [("id", str), ("name", str)],
            [("id", str), ("price", float), ("name", str)],
            {"price": 0.0},
            ["O123", "mixed nuts"],
            ("O123", 0.0, "mixed nuts"),
        ),
        # Missing field - middle
        (
            [("id", str), ("price", float), ("name", str)],
            [("id", str), ("name", str)],
            {},
            ["O123", 25.0, "mixed nuts"],
            ("O123", "mixed nuts"),
        ),
        # Extra field - start
        (
            [("name", str), ("price", float)],
            [("extra", str), ("name", str), ("price", float)],
            {"extra": "default"},
            ["mixed nuts", 25.0],
            ("default", "mixed nuts", 25.0),
        ),
        # Missing field - start
        (
            [("extra", str), ("name", str), ("price", float)],
            [("name", str), ("price", float)],
            {},
            ["unexpected", "mixed nuts", 25.0],
            ("mixed nuts", 25.0),
        ),
    ]
)
```
Does this cover all the edge cases ?

badmonster0: Yes! This decorator sounds great!

python/cocoindex/tests/test_convert.py
```diff
@@ -67,3 +89,193 @@ def test_to_engine_value_tuple():
...
+def test_make_engine_value_converter_defaults_and_missing_fields():
+    # Missing optional field in engine value
+    converter = build_engine_value_converter(Customer)
+    engine_val = ["Alice", ["O1", "item1", 10.0, "default_extra"], None]  # tags explicitly None
+    assert converter(engine_val) == Customer("Alice", Order("O1", "item1", 10.0, "default_extra"), None)
+    # Extra field in engine value (should ignore)
+    engine_val = ["Alice", ["O1", "item1", 10.0, "default_extra"], [["vip"]], "extra"]
+    assert converter(engine_val) == Customer("Alice", Order("O1", "item1", 10.0, "default_extra"), [Tag("vip")])
+
+def test_engine_python_schema_field_order():
```
badmonster0: Since we have the generic parameterized `test_field_position_cases`, seems this case can be folded into that one as another parameter set? (the same for `test_engine_python_schema_extra_field` and `test_engine_python_schema_missing_field`) This will reduce code for these tests and make them more easily understandable and maintainable.

par4m: Following this, scenarios from `test_make_engine_value_converter_struct` and `test_make_engine_value_converter_defaults_and_missing_fields` are also parameterized in a single test: `test_struct_conversion_cases`. Is there anything else that needs refactoring ?

badmonster0:
Thanks a lot for making this test!

Issue #317, opened 2025-04-16
Use `bytes::Bytes` as underlying representation for bytes value


badmonster0:
Currently we're using `Arc[u8]` for bytes.
*   As `KeyValue`: https://github.com/cocoindex-io/cocoindex/blob/79072fbd94b27a87a671bac766c7a3346dcd6e54/src/base/value.rs#L76
*   As `Value`: https://github.com/cocoindex-io/cocoindex/blob/79072fbd94b27a87a671bac766c7a3346dcd6e54/src/base/value.rs#L345

[Crate `bytes`](https://docs.rs/bytes/latest/bytes/) provides a more flexible and efficient way to represent bytes (e.g. if we want a slice of it, we don't need to make a copy). It's widely used by other libraries. We need to do a conversion to interchange with them, e.g. conversion for `neo4rs`: https://github.com/cocoindex-io/cocoindex/blob/762031a26a1334b3b16099095bdfa31e944e4a52/src/ops/storages/neo4j.rs#L285

We want to switch to use `bytes::Bytes` for cocoindex's values.

badmonster0:
closing, thanks @par4m for working on this!

PR #330, opened 2025-04-17
Refactor: migrate from Arc<[u8]> to bytes::Bytes

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/317
This PR migrates all usages of `Arc<[u8]>` to `bytes::Bytes` for bytes representation.

### DIFF
--- a/src/py/convert.rs
+++ b/src/py/convert.rs
@@ -6,7 +6,8 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::collections::BTreeMap;
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::Arc; // Added missing Arc import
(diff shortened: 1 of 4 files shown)


src/py/convert.rs
```diff
@@ -6,7 +6,8 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::collections::BTreeMap;
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::Arc; // Added missing Arc import
```
badmonster0: This `// Added missing Arc import` comment  is unnecessary.

badmonster0 (approved):

badmonster0 (approved):

PR #348, opened 2025-04-19
feat: Add unit tests for SplitRecursively function

### LINKED ISSUE #34: [FEATURE] [FlowCapability] Multiple sources export to the same storage
**What is the use case?**

Users may have documents from multiple different sources and they want to export to the same storage after processing

**Additional context**

FranFigueroa:
Adds unit tests covering basic functionality, overlap, whitespace trimming,
        and discarding empty chunks for the SplitRecursively function.

        Includes TODOs noting potential inconsistencies observed during testing
        regarding overlap and whitespace handling for further investigation.

        Closes #34

### DIFF
--- a/src/ops/functions/split_recursively.rs
+++ b/src/ops/functions/split_recursively.rs
@@ -614,3 +614,165 @@ impl SimpleFunctionFactoryBase for Factory {
         Ok(Box::new(Executor::new(args)?))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*; // Importa todo del módulo padre (el código de split_recursively)
+                 // Nota: Es posible que necesitemos ajustar las importaciones si create_test_value no está donde esperamos.
+                 // Intentaremos importar desde `crate::base::test_utils` si existe, de lo contrario, comentaremos la línea.
+                 // use crate::base::test_utils::create_test_value;
+
+    // Helper para crear un RecursiveChunker de prueba simple
+    fn create_test_chunker(text: &str, chunk_size: usize, chunk_overlap: usize) -> RecursiveChunker {
+        RecursiveChunker {
+            full_text: text,
+            lang_config: None, // Empezar sin lenguaje específico
+            chunk_size,
+            chunk_overlap,
+        }
+    }
+
+    #[test]
+    fn test_translate_bytes_to_chars_simple() {
+        let text = "abc😄def"; // Incluye un caracter multi-byte
+        let mut start1 = 0; // Inicio
+        let mut end1 = 3;   // 'c' (3 bytes)
+        let mut start2 = 3; // Inicio de 😄 (byte 3)
+        let mut end2 = 7;   // Fin de 😄 (byte 7)
+        let mut start3 = 7; // Inicio de 'd' (byte 7)
+        let mut end3 = 10;  // Fin de 'f' (byte 10)
+        let mut end_full = text.len(); // Longitud total en bytes
+
+        let offsets = vec![
+            &mut start1,
+            &mut end1,
+            &mut start2,
+            &mut end2,
+            &mut start3,
+            &mut end3,
+            &mut end_full,
+        ];
+
+        translate_bytes_to_chars(text, offsets.into_iter());
+
+        assert_eq!(start1, 0); // 'a' está en el índice de char 0
+        assert_eq!(end1, 3);   // 'c' está en el índice de char 3
+        assert_eq!(start2, 3); // 😄 empieza en el índice de char 3
+        assert_eq!(end2, 4);   // 😄 termina en el índice de char 4 (ocupa 1 char)
+        assert_eq!(start3, 4); // 'd' empieza en el índice de char 4
+        assert_eq!(end3, 7);   // 'f' termina en el índice de char 7
+        assert_eq!(end_full, 7); // La longitud total en chars es 7
+    }
+
+    // --- Próximos Tests ---
+    #[test]
+    fn test_basic_split_no_overlap() {
+        let text = "Linea 1.\nLinea 2.\n\nLinea 3.";
+        let chunker = create_test_chunker(text, 15, 0); 
+        
+        let result = chunker.split_root_chunk(ChunkKind::RegexpSepChunk { next_regexp_sep_id: 0 });
+        
+        assert!(result.is_ok());
+        let chunks = result.unwrap();
+
+        assert_eq!(chunks.len(), 3);
+        assert_eq!(chunks[0].1, "Linea 1.");
+        assert_eq!(chunks[1].1, "Linea 2.");
+        assert_eq!(chunks[2].1, "Linea 3.");
+
+        let text2 = "Una frase muy larga que definitivamente necesita ser dividida.";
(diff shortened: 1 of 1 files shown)


src/ops/functions/split_recursively.rs
```diff
@@ -614,3 +614,165 @@ impl SimpleFunctionFactoryBase for Factory {
...
+        
+        let result = chunker.split_root_chunk(ChunkKind::RegexpSepChunk { next_regexp_sep_id: 0 });
+        
+        assert!(result.is_ok());
+        let chunks = result.unwrap();
+
+        assert_eq!(chunks.len(), 3);
+        assert_eq!(chunks[0].1, "Linea 1.");
+        assert_eq!(chunks[1].1, "Linea 2.");
+        assert_eq!(chunks[2].1, "Linea 3.");
```
badmonster0: Will be great if we can also validate the range (`.0`). One way to do it is to have a util function, such as `assert_chunk(chunks[0], "Linea 1.")`, which validates both the text and the range automatically (by getting the substring using the range and comparing with the text).

src/ops/functions/split_recursively.rs
```diff
@@ -614,3 +614,165 @@ impl SimpleFunctionFactoryBase for Factory {
...
+        assert_eq!(start2, 3); // 😄 empieza en el índice de char 3
+        assert_eq!(end2, 4);   // 😄 termina en el índice de char 4 (ocupa 1 char)
+        assert_eq!(start3, 4); // 'd' empieza en el índice de char 4
+        assert_eq!(end3, 7);   // 'f' termina en el índice de char 7
+        assert_eq!(end_full, 7); // La longitud total en chars es 7
+    }
+
+    // --- Próximos Tests ---
+    #[test]
+    fn test_basic_split_no_overlap() {
```
badmonster0, line 680: These tests are great! And it'll be even better if we can test on the `Executor` logic, which will cover more code (they call into `split_root_chunk` so this specific logic will also be covered).

Ideally there should be some util to streamline tests for `Factory` and `Executor` (FYI, just filed #352 for this).  So it's totally OK to not to cover `Executor` for now, before we have the util.

badmonster0 (commented):
Thanks a lot for making this PR! Really appreciate the effort!

Would you mind translating the comments to English? Since English is used throughout codebase of this project. Thanks!

FranFigueroa:
Hi!

    Thanks a lot for the review and the feedback! I really appreciate you taking the time.

    I've updated the PR to translate the comments into English as requested. The latest commit reflects this change.

    Let me know if any other changes are needed!

src/ops/functions/split_recursively.rs
```diff
@@ -614,3 +614,156 @@ impl SimpleFunctionFactoryBase for Factory {
         Ok(Box::new(Executor::new(args)?))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*; // Importa todo del módulo padre (el código de split_recursively)
```
badmonster0: Can we translate these comments too? The same for other comments in Spanish below. Thanks!

FranFigueroa:
My apologies for missing those earlier!

I've removed non-English comments. The code should now only contains comments in english, as requested.

Please let me know if everything looks good now or if there's anything else needed.

Thanks!

src/ops/functions/split_recursively.rs
badmonster0: (this comment is optional, mostly FYI, not blocking this PR)

src/ops/functions/split_recursively.rs
badmonster0: Do you think it's possible to make the test cover range?

FranFigueroa: I've implemented the helper function (`assert_chunk_eq`) and updated the tests (`test_basic_split_no_overlap`, `test_split_trims_whitespace`, `test_split_discards_empty_chunks`) to use it for validating both the chunk text and its `RangeValue`.

It took a couple of tries to get the expected ranges correct for the existing behavior in `test_split_discards_empty_chunks`, but the latest push should have the correct assertions now.

For `test_split_trims_whitespace`, as we discussed regarding the inconsistent behavior, I've kept the assertions for chunks 1 and 2 commented out but used the helper for the stable chunk 0 assertion. The TODO comment explains the situation.

I left `test_basic_split_with_overlap` validating only the chunk size for now, as the overlap logic itself seems potentially inconsistent, and validating precise ranges there might also become flaky.

The latest push includes these changes. Please let me know what you think!

src/ops/functions/split_recursively.rs
```diff
@@ -614,3 +614,148 @@ impl SimpleFunctionFactoryBase for Factory {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*; 
+
+    // Helper function to assert both range and text content of a chunk.
+    fn assert_chunk_eq(
+        actual_chunk: &(RangeValue, &str),
+        expected_range: RangeValue,
```
badmonster0: Actually in my mind this util function doesn't need to take the `expected_range` - it's not easy to come up with the right range manually when writing code.

We can do this instead: let this util function gets the full string as input too, and within this util function, get the substring of it using `actual_chunk.0`, and asserting the substring equals to `expected_text`.  i.e. these three should equal to each other: `actual_chunk.1`, `expected_text`, this substring extracted using the range of `actual_chunk.0`.

I feel this will make the tests easier to write and maintain, and also automatically validate an invariance (the range should be consistent with the text).

Let me know what you think about this. Thanks!

src/ops/functions/split_recursively.rs
badmonster0: Almost there! latest looks great, just two small comments

src/ops/functions/split_recursively.rs
badmonster0: Almost there, this is the only thing that needs to address before we can merge, please let me know if there's any question or needs clarification! thanks!

FranFigueroa: Almost got it! I'm working on it

FranFigueroa: Hi!


I've updated the PR:

*   Renamed the helper to `assert_chunk_text_consistency`.
*   It now takes the `full_text` as input and internally extracts the substring using the provided chunk's range, asserting that the expected text, the chunk's text, and the extracted text all match.
*   Updated the calls in `test_basic_split_no_overlap` and `test_split_discards_empty_chunks` to use this new helper.

For `test_split_trims_whitespace`, I've updated the call for the first (stable) chunk to use the new helper. The assertions for the potentially flaky chunks 1 and 2 remain commented out, as documented in the `TODO`, since the core inconsistency issue persists.

The latest push includes these changes, and the tests should now be passing.

Please let me know if this looks better or if any further adjustments are needed. Thanks again for your guidance!

badmonster0: this is exactly what I was thinking about. thanks for the change!

badmonster0:
@FranFigueroa we just published release note and made a section for you https://cocoindex.io/blogs/cocoindex-changelog-2025-04-30#franfigueroa. we love your contribution and welcome to the community!

Issue #350, opened 2025-04-19
[FEATURE] Support Gemini in the DataFlow (Native Support in Rust)


badmonster0:
CocoIndex  🥥 currently supports OpenAI and Ollama ✨ , to add LLM as part of the data pipeline. https://cocoindex.io/docs/ai/llm

Here is an example of how Cocoindex uses Ollama to extract structured information from PDF.
https://cocoindex.io/blogs/cocoindex-ollama-structured-extraction-from-pdf/

We would like to add support for Gemini API (native support in Rust)
- Investigate if there's an official RUST client; or
- Make native HTTP request 

Related code to support [OpenAI](https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex+OpenAI++%28language%3Apython+OR+language%3Arust%29&type=code)

Steps:
1. Update Rust code:
    - Add Enum to LlmApiType in Rust: https://github.com/cocoindex-io/cocoindex/blob/801ae8f421611e9f1359cd2e4158cc161019a104/src/llm/mod.rs#L9-L12
    - Create Gemini client similar to OpenAI client: https://github.com/cocoindex-io/cocoindex/blob/main/src/llm/openai.rs
    - Update the router to connect to the Gemini client: https://github.com/cocoindex-io/cocoindex/blob/801ae8f421611e9f1359cd2e4158cc161019a104/src/llm/mod.rs#L53-L60
2. Add Enum to LlmApiType in Python]: https://github.com/cocoindex-io/cocoindex/blob/0fbfd3f9836ce94d9be65ac82b67f080b2fa7d08/python/cocoindex/llm.py#L4-L7
3. Test with the existing `manuals_llm_extraction` example. You can add a few lines similar to OpenAI: https://github.com/cocoindex-io/cocoindex/blob/555d3286c02f4f7887145418dbd866bfcb91d243/examples/manuals_llm_extraction/main.py#L93-L95
4. Update documentation: https://cocoindex.io/docs/ai/llm

badmonster0:
hey there contributors ❤️- our community member @par4m has been working diligently on this topic
please consider join our https://discord.com/invite/zpA9S2DR7s to check the status with him before you start working on it. thanks!

badmonster0:
Thanks @par4m for adding this feature!

PR #360, opened 2025-04-21
Feat: Add Gemini Support to DataFlow (Rust)

par4m:
This PR adds native Gemini support in DataFlow, implementing [#350](https://github.com/cocoindex-io/cocoindex/issues/350).

**Summary of changes:**
- Added Gemini client and integration.
- Updated example and documentation to include Gemini usage.
- No breaking changes.

**Other changes:**
- Minor logger handling improvement in `lib_context.rs` (unrelated).

### DIFF
--- a/src/llm/gemini.rs
+++ b/src/llm/gemini.rs
@@ -1,21 +1,26 @@
 use async_trait::async_trait;
 use crate::llm::{LlmGenerationClient, LlmSpec, LlmGenerateRequest, LlmGenerateResponse, ToJsonSchemaOptions, OutputFormat};
-use anyhow::{Result, anyhow};
-use serde_json;
-use reqwest::Client as HttpClient;
+use anyhow::{Result, bail};
 use serde_json::Value;
+use crate::api_bail;
+use urlencoding::encode;
 
 pub struct Client {
     model: String,
+    api_key: String,
+    client: reqwest::Client,
 }
 
 impl Client {
     pub async fn new(spec: LlmSpec) -> Result<Self> {
-        if std::env::var("GEMINI_API_KEY").is_err() {
-            anyhow::bail!("GEMINI_API_KEY environment variable must be set");
-        }
+        let api_key = match std::env::var("GEMINI_API_KEY") {
+            Ok(val) => val,
+            Err(_) => api_bail!("GEMINI_API_KEY environment variable must be set"),
+        };
         Ok(Self {
             model: spec.model,
+            api_key,
+            client: reqwest::Client::new(),
         })
     }
 }
@@ -51,12 +56,11 @@ impl LlmGenerationClient for Client {
         })];
 
         // Optionally add system prompt
-        let mut system_instruction = None;
-        if let Some(system) = request.system_prompt {
-            system_instruction = Some(serde_json::json!({
-                "parts": [{ "text": system }]
-            }));
-        }
+        let system_instruction = request.system_prompt.map(|system|
+            serde_json::json!({
+                "parts": [ { "text": system } ]
+            })
+        );
 
         // Prepare payload
         let mut payload = serde_json::json!({ "contents": contents });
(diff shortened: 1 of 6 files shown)


src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+use serde_json::Value;
+
+pub struct Client {
+    model: String,
+}
+
+impl Client {
+    pub async fn new(spec: LlmSpec) -> Result<Self> {
+        if std::env::var("GEMINI_API_KEY").is_err() {
+            anyhow::bail!("GEMINI_API_KEY environment variable must be set");
```
badmonster0: Let's use `api_bail!` for this. It's for problems caused by users. (now it has the same behavior as `bail!` but we may make the way to report this user-caused error more concise/prettier later)

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+            "parts": [{ "text": request.user_prompt }]
+        })];
+
+        // Optionally add system prompt
+        let mut system_instruction = None;
+        if let Some(system) = request.system_prompt {
+            system_instruction = Some(serde_json::json!({
+                "parts": [{ "text": system }]
+            }));
+        }
```
badmonster0: You can try this, which is more concise and avoids mutable variables:

```
let system_instruction = request.system_prompt.map(|system| 
    serde_json::json!({
        "parts": [{ "text": system }]
    }))
```

src/llm/gemini.rs
badmonster0: Consider keeping this env var as a member of `Client`, so we don't need to read/check it again each time during `generate()`.

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+                "responseMimeType": "application/json",
+                "responseSchema": schema_json
+            });
+        }
+
+        let api_key = std::env::var("GEMINI_API_KEY")
+            .map_err(|_| anyhow!("GEMINI_API_KEY environment variable must be set"))?;
+        let url = format!(
+            "https://generativelanguage.googleapis.com/v1beta/models/{}:generateContent?key={}",
+            self.model, api_key
```
badmonster0: Let's encode the URL component to avoid it has special characters in case. e.g. using something like https://docs.rs/urlencoding/latest/urlencoding/

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
+use async_trait::async_trait;
+use crate::llm::{LlmGenerationClient, LlmSpec, LlmGenerateRequest, LlmGenerateResponse, ToJsonSchemaOptions, OutputFormat};
+use anyhow::{Result, anyhow};
+use serde_json;
+use reqwest::Client as HttpClient;
```
badmonster0: nit: I will avoid renaming a class if possible. I'll avoid this `use` and just directly use `reqwest::Client` across the code. Because `reqwest::Client` is commonly known by people and when people see it they will immediately know which specific type it's referring to, when reading the code.

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        let api_key = std::env::var("GEMINI_API_KEY")
+            .map_err(|_| anyhow!("GEMINI_API_KEY environment variable must be set"))?;
+        let url = format!(
+            "https://generativelanguage.googleapis.com/v1beta/models/{}:generateContent?key={}",
+            self.model, api_key
+        );
+
+        let client = HttpClient::new();
```
badmonster0: This client should be created in `new` and reused across different `generate()` call - it manages a connection pool internally and the pool should be shared.

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+        let resp = client.post(&url)
+            .json(&payload)
+            .send()
+            .await
+            .map_err(|e| anyhow!("HTTP error: {e}"))?;
+
+        let resp_json: Value = resp.json().await.map_err(|e| anyhow!("Invalid JSON: {e}"))?;
+
+        if let Some(error) = resp_json.get("error") {
+            return Err(anyhow!("Gemini API error: {:?}", error));
```
badmonster0: We can directly use `bail!(....)` here

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+        let text = resp_json["candidates"][0]["content"]["parts"][0]["text"]
```
badmonster0: (optional optimization) It's possible to avoid a string copy here by something like:

```
let mut resp_json = ...
...
let text = match &mut resp_json["candidates"][0]["content"]["parts"][0]["text"] {
  Some(serde_json::Value::String(s)) => std::mem::take(s),
  _ => bail!("No text in response"),
};
```

src/llm/gemini.rs
```diff
@@ -0,0 +1,112 @@
...
+            anyhow::bail!("GEMINI_API_KEY environment variable must be set");
+        }
+        Ok(Self {
+            model: spec.model,
+        })
+    }
+}
+
+// Recursively remove all `additionalProperties` fields from a JSON value
+fn remove_additional_properties(value: &mut Value) {
```
badmonster0, line 29: It'll be nice if we avoid setting this at the first place.

FYI, the code to set it is here:

https://github.com/cocoindex-io/cocoindex/blob/f530811b7f6c66d353dc66f76215e1d6eba14f39/src/base/json_schema.rs#L208

We can add a new boolean field to `ToJsonSchemaOptions`, and check the value of this new field there.

As we discussed in the Discord, feel free to do this as a follow-up PR.

par4m: I can work on it after this is merged since it requires changes in generator logic, created an issue for the same here https://github.com/cocoindex-io/cocoindex/issues/364

badmonster0: Sounds good. Thanks!

src/llm/gemini.rs
```diff
@@ -0,0 +1,120 @@
...
+        let api_key = &self.api_key;
+        let url = format!(
+            "https://generativelanguage.googleapis.com/v1beta/models/{}:generateContent?key={}",
+            encode(&self.model), encode(api_key)
+        );
+
+        let resp = match self.client.post(&url)
+            .json(&payload)
+            .send()
+            .await {
```
badmonster0: `anyhow` crates provided a shortcut `context()`, so we can do `...await.context("HTTP error")?` here. It carries the previous error with the string as additional context.

Same below.

src/llm/gemini.rs
```diff
@@ -0,0 +1,120 @@
...
+        // Optionally add system prompt
+        let system_instruction = request.system_prompt.map(|system|
+            serde_json::json!({
+                "parts": [ { "text": system } ]
+            })
+        );
+
+        // Prepare payload
+        let mut payload = serde_json::json!({ "contents": contents });
+        if let Some(system) = system_instruction {
```
badmonster0: Actually you can also merge L59-63 to here, to things even simpler:

```
if let Some(system) = request.system_prompt {
  payload["systemInstruction"] = serde_json::json!({
                "parts": [ { "text": system } ]
            });
}
```

badmonster0:
thanks a lot for adding the feature!

Issue #346, opened 2025-04-19
Create a Colab Note book example to run CocoIndex end to end


badmonster0:
Create a Colab note book example for [CocoIndex text embedding](https://github.com/cocoindex-io/cocoindex/tree/main/examples/text_embedding).

An great [example note book flow] (https://colab.research.google.com/github/qdrant/examples/blob/master/qdrant_101_image_data/04_qdrant_101_cv.ipynb#scrollTo=_pVKD77ZcYQ6) as reference

par4m:
I'll work on this!

PR #362, opened 2025-04-21
Add Colab Notebook for Text Embedding Example

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/346


examples/text_embedding/Text_Embedding.ipynb
```diff
@@ -0,0 +1,1189 @@
...
...xIDI2LjY1MTZMMzQuOTg0NCAyNi42NDM2QzM0LjgzNDYgMjYuNjEzNyAzNC42NzkxIDI2LjYzNDUgMzQuNTQyNSAyNi43MDI4TDM0LjUwOSAyNi43MTk2QzM0LjM5NzggMjYuNzc1MiAzNC4zMDc2IDI2Ljg2NTMgMzQuMjUyIDI2Ljk3NjVDMzQuMjM5OSAyNy4wMDA4IDM0LjIyOTUgMjcuMDI1OSAzNC4yMjA5IDI3LjA1MTdMMzQuMTg1OSAyNy4xNTY1QzM0LjE0ODggMjcuMjY3OSAzNC4xMjk5IDI3LjM4NDUgMzQuMTI5OSAyNy41MDE5QzM0LjEyOTkgMjcuNTkxMiAzNC4xMTg5IDI3LjY4MDIgMzQuMDk3MyAyNy43NjY4TDM0LjA4OTkgMjcuNzk2MUMzNC4wNDc3IDI3Ljk2NSAzMy45NzIyIDI4LjEyMzcgMzMuODY3OCAyOC4yNjI5QzMzLjc2NjQgMjguMzk4IDMzLjYzOTggMjguNTEyMiAzMy40OTUgMjguNTk5MUwzMy4zODE0IDI4LjY2NzNDMzMuMjU3NCAyOC43NDE3IDMzLjEyNTYgMjguODAyMyAzMi45ODg0IDI4Ljg0OEwzMi43ODggMjguOTE0OEMzMi42NDM4IDI4Ljk2MjkgMzIuNDk2MiAyOC45OTk2IDMyLjM0NjMgMjkuMDI0NUMzMi4wODIxIDI5LjA2ODYgMzEuODEzIDI5LjA3NTkgMzEuNTQ2OCAyOS4wNDYzTDMxLjMzNTMgMjkuMDIyOEMzMS4xMjEzIDI4Ljk5OTEgMzAuOTExIDI4Ljk0ODggMzAuNzA5MyAyOC44NzMxTDMwLjYzMTEgMjguODQzOEMzMC4zMzUgMjguNzMyOCAzMC4wNjYxIDI4LjU1OTYgMjkuODQyNSAyOC4zMzZDMjkuNjU0NyAyOC4xNDgyIDI5LjUwMjEgMjcuOTI4MSAyOS4zOTIyIDI3LjY4NjJMMjkuMDc4OCAyNi45OTY4QzI5LjA1MjUgMjYuOTM4OCAyOS4wMTY0IDI2Ljg4NDkgMjguOTY4NyAyNi44NDI3WiIgZmlsbD0iIzQwMUUyQiIvPgo8L2c+CjxkZWZzPgo8L2RlZnM+Cjwvc3ZnPgo=) Welcome to [Cocoindex](https://cocoindex.io/)\n",
```
badmonster0: For the image, we can use the URL `https://cocoindex.io/icon.svg` instead of the inlined data.

examples/text_embedding/Text_Embedding.ipynb
```diff
@@ -0,0 +1,1189 @@
...
+        },
+        "collapsed": true,
+        "id": "rQcJanCi-W3I",
+        "outputId": "a5362319-f7ca-45f0-9754-e8a825ac1696"
+      },
+      "execution_count": null,
+      "outputs": [
+        {
+          "output_type": "stream",
+          "name": "stdout",
```
badmonster0: We shouldn't put output in the file. I think before exporting the file, you can do a "Edit -> Clear all outputs".

examples/text_embedding/Text_Embedding.ipynb
```diff
@@ -0,0 +1,498 @@
...
+        "# Grab some markdown files for demo"
+      ],
+      "metadata": {
+        "id": "Xh2sMemiA7_N"
+      }
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "!mkdir markdown_files && cd markdown_files && wget https://raw.githubusercontent.com/cocoindex-io/cocoindex/refs/heads/main/examples/text_embedding/markdown_files/1706.03762v7.md && wget https://raw.githubusercontent.com/cocoindex-io/cocoindex/refs/heads/main/examples/text_embedding/markdown_files/1810.04805v2.md && wget https://raw.githubusercontent.com/cocoindex-io/cocoindex/refs/heads/main/examples/text_embedding/markdown_files/rfc8259.md && cd .."
```
badmonster0: This is too long and maybe better if can be split into multiple lines. Will using trailing `\` for each line just work?

examples/text_embedding/README.md
badmonster0, line 1: FYI, I noticed the color of the `cocoindex setup` doesn't show well under light background.  This can be handled separately. I filed #369

examples/text_embedding/Text_Embedding.ipynb
par4m: Yup! that'll work. I was thinking maybe like we can just write a script on https://gist.github.com/ and run it. Since a user wouldn't really mind since they're just trying out an example. What do you think ?

badmonster0: So the colab can directly use the script in gist?

Actually I still prefer putting the script in this notebook file, so it's self contained and under version control. We don't need to manage a separate script on another location.

par4m: Got it. Also for the last part about exposing CocoInsight using Cloudflared, should I leave it there ? It doesn't work right now for some reason

badmonster0: I think we can remove it for now. Thanks!

badmonster0:
Thanks a lot for adding this!

PR #371, opened 2025-04-22
Update Colab Link to Match Github

par4m:
README of text embedding example now points to https://colab.research.google.com/github/cocoindex-io/cocoindex/blob/main/examples/text_embedding/Text_Embedding.ipynb


badmonster0:
super neat! thanks a lot for making this @par4m !

Issue #369, opened 2025-04-22
[BUG] "cocoindex setup" title colors don't show well under bright background


badmonster0:
Under Colab:

![Image](https://github.com/user-attachments/assets/c1620453-e479-4b3c-a9c3-dcf8f5acb6d1)

Under terminal with bright background:

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/18a1d312-3657-4e59-a12c-65cd4ce19bb6" />

badmonster0:
It seems hard to detect background color in tty. To solve this, one way is to pick colors working for both light and dark themes.

For now, probably leaving default colors to titles will just work.

par4m:
I guess there are a few workarounds for this, 
1. "guessing" if $TERM or $COLORTERM looks like it belongs to a light theme
2. Ask the user to set an environment variable (e.g., TERMINAL_THEME=dark) and adapt based on that.
3. Provide a CLI flag like --dark-theme or --light-theme. (if they want, otherwise default to current setup)

I personally think a cli flag is the most optimal solution for this, we can add multiple themes this way since owo-colors supports setting XTerm colors

par4m:
For now a theme that's neutral to both light and dark themes can be used. Any suggestions ?

badmonster0:
I feel we should keep the control minimum. It's tedious for users to provide another flag or environment variable, especially new users don't know how to set it. The default appearance should be good enough for most users.

The current color themes:
1. The status like `[ READY ]` - this parts work well for both light and dark.
2. Headings (the line to the right of the status) like `CocoIndex Metadata Table` and `Flow: DocToKG`. This should be more obvious than details (TODOs below), but under light background it's hard to be seen.
3. Details (the TODOs) - this also seems good. Under both light and dark, it's less eminent than others.

So overall I think maybe it's good enough to leave default colors to "2" - then it'll be more eminent than "3" under both dark and light.

par4m:
Also I realized that the colors that are enforced just take the shade that the terminal set for that color,
e.g

![Image](https://github.com/user-attachments/assets/a7848338-3286-4c01-94f7-8412c4058253)

Is this the default behavior that you expect or would you want to enforce a specific e.g Cyan shade ? (setting XTerm color)

badmonster0:
This looks pretty good. Thanks for trying out!

par4m:
[...]

PR #372, opened 2025-04-22
Fix Title Visibility for bright backgrounds

par4m:
Fixes https://github.com/cocoindex-io/cocoindex/issues/369
Remove color from description in cli so that it uses the foreground color of the terminal, making it visible regardless of the theme.


badmonster0:
thanks for the fix!

PR #381, opened 2025-04-24
Update contributing.md

par4m:
- Use a more standard way of creating virtual environment
- Ensure latest version of rust

### DIFF
--- a/docs/docs/about/contributing.md
+++ b/docs/docs/about/contributing.md
@@ -24,15 +24,17 @@ We tag issues with the ["good first issue"](https://github.com/cocoindex-io/coco
 Following the steps below to get cocoindex build on latest codebase locally - if you are making changes to cocoindex funcionality and want to test it out.
 
 -   Install Rust toolchain: [docs](https://rust-lang.org/tools/install)
+    
+    Run `rustup update` to ensure you're on the latest version.
 
--   (Optional) Setup Python virtual environment:
+-   (Recommended) Setup Python virtual environment:
     ```bash
-    virtualenv --python=$(which python3.12) .venv
+    python3 -m venv .venv
     ```
     Activate the virtual environment, before any installings / buildings / runnings:
 
     ```bash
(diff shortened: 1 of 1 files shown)


docs/docs/about/contributing.md
```diff
@@ -24,15 +24,17 @@ We tag issues with the ["good first issue"](https://github.com/cocoindex-io/coco
...
+-   (Recommended) Setup Python virtual environment:
     ```bash
-    virtualenv --python=$(which python3.12) .venv
+    python3 -m venv .venv
     ```
     Activate the virtual environment, before any installings / buildings / runnings:
 
     ```bash
-    . .venv/bin/activate
+    source .venv/bin/activate
```
badmonster0: I think `.` is POSIX standard, and `source` is only supported by a subset of shells like `bash`. e.g. `dash` doesn't really support `source`: https://unix.stackexchange.com/questions/700562/sourcing-file-in-dash-with-file-gives-not-found-error

So probably we should keep `.` here.

badmonster0 (commented):
thanks for polishing the docs!

docs/docs/about/contributing.md
par4m: I didn't know this, thanks for the info! I'll change it right away. Also, would you prefer to include the Rust install command `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` directly instead of linking to the docs, since it's unlikely to change?

badmonster0: Sounds good. We can directly put this command, and keep the link to the original doc for reference. Thanks!

par4m: This is how it looks, what do you think ?
image

badmonster0 (approved):

docs/docs/about/contributing.md
badmonster0: Looks great. Thanks!

Issue #176, opened 2025-03-19
[FEATURE] Integrate with Antropic API


badmonster0:
CocoIndex  🥥 currently supports OpenAI and Ollama ✨ , to add LLM as part of the data pipeline. https://cocoindex.io/docs/ai/llm

Here is an example of how Cocoindex uses Ollama to extract structured information from PDF.
https://cocoindex.io/blogs/cocoindex-ollama-structured-extraction-from-pdf/

We would like to add support for Anthropic API - https://docs.anthropic.com/en/api/getting-started

Related code to support [OpenAI](https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex+OpenAI++%28language%3Apython+OR+language%3Arust%29&type=code)

Steps:
1. Update Rust code:
    - Add Enum to LlmApiType in Rust: https://github.com/cocoindex-io/cocoindex/blob/801ae8f421611e9f1359cd2e4158cc161019a104/src/llm/mod.rs#L9-L12
    - Create Anthropic client similar to OpenAI client: https://github.com/cocoindex-io/cocoindex/blob/main/src/llm/openai.rs
    - Update the router to connect to the Anthropic client: https://github.com/cocoindex-io/cocoindex/blob/801ae8f421611e9f1359cd2e4158cc161019a104/src/llm/mod.rs#L53-L60
2. Add Enum to LlmApiType in Python]: https://github.com/cocoindex-io/cocoindex/blob/0fbfd3f9836ce94d9be65ac82b67f080b2fa7d08/python/cocoindex/llm.py#L4-L7
3. Test with the existing `manuals_llm_extraction` example. You can add a few lines similar to OpenAI: https://github.com/cocoindex-io/cocoindex/blob/555d3286c02f4f7887145418dbd866bfcb91d243/examples/manuals_llm_extraction/main.py#L93-L95
4. Update documentation: https://cocoindex.io/docs/ai/llm

par4m:
I can work on this, please assign!

badmonster0:
SG! thanks a lot @par4m

iamsb97:
Hi @par4m , have you started working on this?

par4m:
> Hi [@par4m](https://github.com/par4m) , have you started working on this?

Yep @iamsb97 , just doing some final testing before the PR.

badmonster0:
Thanks a lot @par4m for getting this feature in!

PR #395, opened 2025-04-25
feat: Add Anthropic Support

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/176
- Used `json5` crate  for fallback
- No breaking changes


src/llm/anthropic.rs
```diff
@@ -0,0 +1,132 @@
...
+            "messages": messages,
+            "max_tokens": 4096
+        });
+
+        // Add system prompt as top-level field if present (required)
+        if let Some(system) = request.system_prompt {
+            payload["system"] = serde_json::json!(system);
+        }
+
+        let OutputFormat::JsonSchema { schema, .. } = request.output_format.as_ref().expect("Anthropic client expects OutputFormat::JsonSchema for all requests");
```
badmonster0: This may panic. We usually prefer raising an error instead of panicking.

We can use something like `ok_or_else(|| api_error!("..."))?` for this case.

src/llm/anthropic.rs
```diff
@@ -0,0 +1,132 @@
...
+        let schema_json = serde_json::to_value(schema)?;
+        payload["tools"] = serde_json::json!([
+            { "type": "custom", "name": "extraction", "input_schema": schema_json }
```
badmonster0: For the LLM, purpose of the tool is to deliver result. So the name "extraction" seems confusing for the LLM. Maybe use something similar to "report_result".

src/llm/anthropic.rs
```diff
@@ -0,0 +1,132 @@
...
+                    if let Some(input) = item.get("input") {
+                        extracted_json = Some(input.clone());
+                        break;
+                    }
+                }
+            }
+        }
+        let text = if let Some(json) = extracted_json {
+            // Try strict JSON serialization first
+            serde_json::to_string(&json)?
```
badmonster0, line 100: I think we may change the trait to allow `generate()` directly return a JSON (the return type can be an enum, with either a JSON or a String, depending on if JSON schema is provided in request).

I can also file a issue for this follow-up if you prefer keeping this PR smaller.

src/llm/anthropic.rs
```diff
@@ -0,0 +1,132 @@
...
+                Value::String(s) => {
+                    // Try strict JSON parsing first
+                    match serde_json::from_str::<serde_json::Value>(s) {
+                        Ok(_) => s.clone(),
+                        Err(e) => {
+                            // Try permissive json5 parsing as fallback
+                            match json5::from_str::<serde_json::Value>(s) {
+                                Ok(_) => {
+                                    println!("[Anthropic] Used permissive JSON5 parser for output");
+                                    s.clone()
```
badmonster0: Note that if it doesn't fit into strict JSON, it'll still fails if we return the original string (because the caller will also try to parse it as strict JSON).

I think what we should do here is to serialize the value parsed as JSON5 as JSON.

src/llm/anthropic.rs
```diff
@@ -0,0 +1,132 @@
...
+        let text = if let Some(json) = extracted_json {
+            // Try strict JSON serialization first
+            serde_json::to_string(&json)?
+        } else {
+            // Fallback: try text if no tool output found
+            match &resp_json["content"][0]["text"] {
+                Value::String(s) => {
+                    // Try strict JSON parsing first
+                    match serde_json::from_str::<serde_json::Value>(s) {
+                        Ok(_) => s.clone(),
```
badmonster0: (optional) Seems this `clone()` can be avoided, if we move/take the text from `resp_json`.

src/llm/anthropic.rs
par4m: A new issue for this will be great, Thanks!

badmonster0: created #400. thanks!

badmonster0:
super neat! thanks for making it happen!

badmonster0:
@par4m - new release blog is out and we made a section for you! https://cocoindex.io/blogs/cocoindex-changelog-2025-04-30#par4m  we love your contribution and thank you!

par4m:
> @par4m - new release blog is out and we made a section for you! https://cocoindex.io/blogs/cocoindex-changelog-2025-04-30#par4m we love your contribution and thank you!

Thank you so much!

PR #404, opened 2025-04-28
feat: beautify show cli command with colorful structured output

### LINKED ISSUE #324: [FEATURE] Improve "cocoindex show" to make it more useful
The current `cocoindex show` CLI is simply printing JSON for a flow spec.

We want to improve to make it more useful:

* Show information in more cleared way, e.g. source by source, target by target.
* Show schema of data scopes.

---
To contributors  ❤️:  Please leave a comment `I'm working on this`  if you decide to take it  -  to better communicate with other contributors :)

lemorage:
As the title states, this PR creates a helper method to make the output spec info more structured in a way that users can appreciate, and also modifies the show command to use colorful output mode by default.

Resolves #324

before:
```
{
  "name": "TextEmbedding",
  "import_ops": [
    {...}
  ],
  "reactive_ops": [
    {...}
  ],
  "export_ops": [
    {...}
  ]
}
```

after:
\
show-cli

### DIFF
--- a/python/cocoindex/cli.py
+++ b/python/cocoindex/cli.py
@@ -52,11 +52,31 @@ def ls(show_all: bool):
 
 @cli.command()
 @click.argument("flow_name", type=str, required=False)
-def show(flow_name: str | None):
+@click.option("--no-color", is_flag=True, help="Disable colored output.")
+def show(flow_name: str | None, no_color: bool):
     """
-    Show the flow spec.
+    Show the flow spec in a readable format with colored output.
     """
-    click.echo(str(_flow_by_name(flow_name)))
+    flow = _flow_by_name(flow_name)
+    flow_str = str(flow)
+
+    for line in flow_str.splitlines():
+        line = line.rstrip()
+        if not line:
+            click.echo(line, color=not no_color)
+            continue
+
+        if line.startswith("Flow:") or line in ("Sources:", "Processing:", "Targets:"):
+            click.secho(line, fg="cyan", bold=True, color=not no_color)
+            continue
+
+        if ":" in line:
+            key, value = line.split(":", 1)
+            click.secho(f"{key}:", fg="green", nl=False, color=not no_color)
+            click.secho(value, fg="yellow", color=not no_color)
+            continue
+
+        click.secho(line, fg="yellow", color=not no_color)
 
 @cli.command()
 def setup():
--- a/python/cocoindex/cli.py
+++ b/python/cocoindex/cli.py
@@ -1,6 +1,7 @@
 import asyncio
 import click
 import datetime
+from rich.console import Console
 
 from . import flow, lib
 from .setup import sync_setup, drop_setup, flow_names_with_setup, apply_setup_changes
@@ -52,31 +53,14 @@ def ls(show_all: bool):
 
 @cli.command()
 @click.argument("flow_name", type=str, required=False)
-@click.option("--no-color", is_flag=True, help="Disable colored output.")
-def show(flow_name: str | None, no_color: bool):
+@click.option("--color/--no-color", default=True)
+def show(flow_name: str | None, color: bool):
     """
     Show the flow spec in a readable format with colored output.
     """
     flow = _flow_by_name(flow_name)
-    flow_str = str(flow)
-
-    for line in flow_str.splitlines():
-        line = line.rstrip()
-        if not line:
-            click.echo(line, color=not no_color)
-            continue
-
-        if line.startswith("Flow:") or line in ("Sources:", "Processing:", "Targets:"):
-            click.secho(line, fg="cyan", bold=True, color=not no_color)
-            continue
-
-        if ":" in line:
-            key, value = line.split(":", 1)
-            click.secho(f"{key}:", fg="green", nl=False, color=not no_color)
-            click.secho(value, fg="yellow", color=not no_color)
-            continue
-
-        click.secho(line, fg="yellow", color=not no_color)
+    console = Console(no_color=not color)
+    console.print(flow._render_text())
 
 @cli.command()
 def setup():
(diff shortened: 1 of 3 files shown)


python/cocoindex/cli.py
```diff
@@ -52,11 +52,31 @@ def ls(show_all: bool):
 
 @cli.command()
 @click.argument("flow_name", type=str, required=False)
-def show(flow_name: str | None):
+@click.option("--no-color", is_flag=True, help="Disable colored output.")
```
badmonster0: We may consider define a pair of flags `--color`/`--no-color` (https://click.palletsprojects.com/en/stable/options/#boolean-flags) and default it to `True` (i.e. show with color)

python/cocoindex/flow.py
```diff
@@ -452,7 +453,59 @@ def _lazy_engine_flow() -> _engine.Flow:
         self._lazy_engine_flow = _lazy_engine_flow
 
     def __str__(self):
-        return str(self._lazy_engine_flow())
+        try:
+            flow_spec_str = str(self._lazy_engine_flow())
```
badmonster0: It's clearer to put this  line before the `try` (it should never fail, and the `except` branch uses its value)

python/cocoindex/flow.py
```diff
@@ -452,7 +453,59 @@ def _lazy_engine_flow() -> _engine.Flow:
...
+
+            if isinstance(data, dict):
+                for key, value in data.items():
+                    full_key = f"{parent_key}.{key}" if parent_key else key
+                    build_hierarchy(value, full_key, hierarchy)
+            elif isinstance(data, list):
+                for index, item in enumerate(data):
+                    build_hierarchy(item, f"{parent_key}[{index}]", hierarchy)
+            else:
+                keys = parent_key.split('.')
```
badmonster0: Looks like this logic can be simplify a little bit. `build_hierarchy` is already a recursive function, so we don't need another loop to traverse from the root to the leaf field.

We can change `build_hierarchy` to only return the converted object for the current input `data` (instead of the root hierarchy), so that it doesn't even need to take the root hierarchy and `parent_key` as argument.  Something like this:

```
def build_hierarchy(data):
  if data is None:
    return None
  elif isinstance(data, dict):
    return {key: build_hierarchy(value) for (key, value) in data.items()}
  elif isinstance(data, list):
    return [f"[{index}]": build_hierarchy(item) for (index, item) in data.enumerate()}
  else:
    return data
```

python/cocoindex/cli.py
```diff
@@ -52,11 +52,31 @@ def ls(show_all: bool):
...
+        if not line:
+            click.echo(line, color=not no_color)
+            continue
+
+        if line.startswith("Flow:") or line in ("Sources:", "Processing:", "Targets:"):
+            click.secho(line, fg="cyan", bold=True, color=not no_color)
+            continue
+
+        if ":" in line:
+            key, value = line.split(":", 1)
```
badmonster0: this may not be super reliable, as `:` doesn't always mean a separator between key and value.

For example, sometimes there can be a long string passed as part of the function spec (like [this](https://github.com/cocoindex-io/cocoindex/blob/08b428d42dc658384b6fc2293eedd14264e149e7/examples/manuals_llm_extraction/main.py#L106)) and this can have multi lines.

The root of the difficulty here is: we convert structured data to unstructured (lines), then here we have to do the reverse side: parse some structure from unstructured data. I feel it will make things simpler if we can directly add color/style before losing the structure.

I searched a little bit, and the `rich` library can be helpful for this purpose. [`rich.Text`](https://rich.readthedocs.io/en/stable/text.html) can represent text with colors. So maybe we can change `_format_flow` to return `Text` (which already have color information), and call `rich.print(text)` here to print (we can call `str(text)` if we don't want color, e.g. in `__str__()` or here when `no_color` is set)

badmonster0 (commented):
very nicely looked printout! thanks for making this!

python/cocoindex/cli.py
lemorage: This will be resolved, but I will take a grain of salt. Since we default it to colorful output, I did not see any sense in making it a boolean flag pair.

python/cocoindex/cli.py
lemorage: reasonable to me

python/cocoindex/flow.py
```diff
@@ -451,8 +454,60 @@ def _lazy_engine_flow() -> _engine.Flow:
...
+
+        def format_data(data, indent=0):
+            if isinstance(data, dict):
+                for key, value in data.items():
+                    if isinstance(value, (dict, list)):
+                        add_line(f"- {key}:", indent, style="green")
+                        format_data(value, indent + 2)
+                    else:
+                        add_line(f"- {key}:", indent, style="green", end="")
+                        add_line(f" {value}", style="yellow")
```
badmonster0: we can consider extracting L472-477 into another function, taking the prefix string (`f"- {key}:"` part) and the value as arguments. Then L480-485 can call this function too.

python/cocoindex/cli.py
badmonster0: Appreciate sharing the thoughts!

To share the rational in my mind: it'll simplify both users expectation and our code if consistently having a positive flag pairing with the negative one.  users don't always remember the default and natural for them to think there's a `--color`, and in our code `color = True` is a little straightforward than `no_color = False` (double negative)

badmonster0:
Thanks for this change! super neat!

FYI, I filed two additional issues #410 and #411 as further enhancement in related area (needs to touch Rust codebase).

lemorage:
Will look at them, thanks!

badmonster0:
@lemorage We just published release note and made a section for your beautiful work  https://cocoindex.io/blogs/cocoindex-changelog-2025-04-30/#lemorage. we love your contribution and thank you!

PR #427, opened 2025-05-03
feat(cli): enhance flow display with schema rendering support

### LINKED ISSUE #411: [FEATURE] Also show schema information in `cocoindex show`
We want `cocoindex show` also show schema information of a flow.

- Example of existing code to get schema: https://github.com/cocoindex-io/cocoindex/blob/7fef101c7308bd16522b0979fa47698ae52ddd41/src/service/flows.rs#L36-L37
- We want to pretty format the schema on Rust side and expose it to Python by some methods

---

lemorage:
This PR tries to resolve #411.

We implement a flow schema function on Rust side, and expose it to Python, so the `show` command can directly use. Aside, we use `rich.Table` to beautify the schema information of each flow.

Now, when using `cocoindex show`, aside from spec info, schema info will also print out at last as follows:

image

### DIFF
--- a/python/cocoindex/cli.py
+++ b/python/cocoindex/cli.py
@@ -1,7 +1,9 @@
+import asyncio
 import click
 import datetime
 
 from rich.console import Console
+from rich.table import Table
 
 from . import flow, lib, setting
 from .setup import sync_setup, drop_setup, flow_names_with_setup, apply_setup_changes
@@ -54,26 +56,25 @@ def ls(show_all: bool):
 @cli.command()
 @click.argument("flow_name", type=str, required=False)
 @click.option("--color/--no-color", default=True)
-@click.option(
-    "--schema", is_flag=True, show_default=True, default=False,
-    help="Also show the schema of the flow.")
-def show(flow_name: str | None, color: bool, schema: bool):
+def show(flow_name: str | None, color: bool):
     """
-    Show the flow spec in a readable format with colored output.
-    Optionally display the schema if --schema is provided.
+    Show the flow spec in a readable format with colored output,
+    including the schema.
     """
-    from rich.table import Table
     flow = _flow_by_name(flow_name)
     console = Console(no_color=not color)
     console.print(flow._render_text())
-    if schema:
+
+    async def render_schema_and_print():
         table = Table(title=f"Schema for Flow: {flow.name}", show_header=True, header_style="bold magenta")
         table.add_column("Field", style="cyan")
         table.add_column("Type", style="green")
         table.add_column("Attributes", style="yellow")
-        for field_name, field_type, attr_str in flow._render_schema():
+        for field_name, field_type, attr_str in await flow._render_schema():
             table.add_row(field_name, field_type, attr_str)
         console.print(table)
+
+    asyncio.run(render_schema_and_print())
 
 @cli.command()
 def setup():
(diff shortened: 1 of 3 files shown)


src/py/mod.rs
```diff
@@ -365,6 +369,76 @@ fn add_auth_entry(key: String, value: Pythonized<serde_json::Value>) -> PyResult
     Ok(())
 }
 
+#[pyfunction]
+fn format_flow_schema<'py>(py: Python<'py>, flow_name: String) -> PyResult<Bound<'py, PyAny>> {
```
badmonster0: We may make this a method of `Flow` https://github.com/cocoindex-io/cocoindex/blob/bac81a5bc6ab51577c3db6f75e6ae0cbe26d4eb9/src/py/mod.rs#L162

`Flow` already has the schema in its context, so we won't need to get the information from the global `lib_context` by name. Also we don't need another async function.

python/cocoindex/cli.py
```diff
@@ -56,11 +58,23 @@ def ls(show_all: bool):
...
+    async def render_schema_and_print():
+        table = Table(title=f"Schema for Flow: {flow.name}", show_header=True, header_style="bold magenta")
+        table.add_column("Field", style="cyan")
+        table.add_column("Type", style="green")
+        table.add_column("Attributes", style="yellow")
+        for field_name, field_type, attr_str in await flow._render_schema():
+            table.add_row(field_name, field_type, attr_str)
+        console.print(table)
+
+    asyncio.run(render_schema_and_print())
```
badmonster0: to execute an async function, let's use `execution_context.run()` (it uses a dedicated event loop specific to cocoindex library, so in case users call this function from another async function, things won't be broken)

badmonster0 (commented):
That's really beautiful! Love this change!

src/py/mod.rs
```diff
@@ -193,6 +195,69 @@ impl Flow {
...
+        fn process_fields(
+            fields: &[FieldSchema],
+            prefix: &str,
+            result: &mut Vec<(String, String, String)>,
+        ) {
+            for field in fields {
+                let field_name = format!("{}{}", prefix, field.name);
+
+                let mut field_type = match &field.value_type.typ {
+                    ValueType::Basic(basic) => match basic {
```
badmonster0: Here we can directly use `format!("{}", basic)` for basic types as `Display` is implemented for it.

src/py/mod.rs
```diff
@@ -193,6 +195,69 @@ impl Flow {
...
+                            let elem = match *v.element_type {
+                                BasicValueType::Float32 => "Float32",
+                                BasicValueType::Float64 => "Float64",
+                                _ => "Unknown",
+                            };
+                            format!("Vector[{}, {}]", dim, elem)
+                        }
+                        other => format!("{:?}", other),
+                    },
+                    ValueType::Table(t) => format!("{:?}", t.kind),
```
badmonster0: Prefer `"{}"` to `"{:?}"`. `"{:?}"` uses `Debug` which is supposed to be more verbose in general (although in this case they're the same for now).

src/py/mod.rs
```diff
@@ -193,6 +195,69 @@ impl Flow {
...
+                };
+
+                result.push((field_name.clone(), field_type, attr_str));
+
+                match &field.value_type.typ {
+                    ValueType::Struct(s) => {
+                        process_fields(&s.fields, &format!("{}.", field_name), result);
+                    }
+                    ValueType::Table(t) => {
+                        process_fields(&t.row.fields, &format!("{}.", field_name), result);
```
badmonster0: For tables, maybe we can use `"{}[]."` (or `"{}[*]."`) here, so that it's more distinguishable from structs.

badmonster0 (approved):

PR #432, opened 2025-05-05
Added new fastapi_server_docker example

AbdelRahmanYaghi:
This example contains a simple docker compose which runs pg17 along with a simple fastapi endpoint for querying on code.

### DIFF
--- a/examples/fastapi_server_docker/main.py
+++ b/examples/fastapi_server_docker/main.py
@@ -4,18 +4,15 @@ import uvicorn
 from fastapi import FastAPI
 from dotenv import load_dotenv
 
-from src.cocoindex_funs import code_embedding_flow
+from src.cocoindex_funs import code_embedding_flow, code_to_embedding
 
 fastapi_app = FastAPI()
-
     
 query_handler = cocoindex.query.SimpleSemanticsQueryHandler(
     name="SemanticsSearch",
     flow=code_embedding_flow,
     target_name="code_embeddings",
-    query_transform_flow=lambda text: text.transform(
-        cocoindex.functions.SentenceTransformerEmbed(
-            model="sentence-transformers/all-MiniLM-L6-v2")),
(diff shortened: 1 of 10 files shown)


examples/fastapi_server_docker/main.py
```diff
@@ -0,0 +1,33 @@
...
+    
+query_handler = cocoindex.query.SimpleSemanticsQueryHandler(
+    name="SemanticsSearch",
+    flow=code_embedding_flow,
+    target_name="code_embeddings",
+    query_transform_flow=lambda text: text.transform(
+        cocoindex.functions.SentenceTransformerEmbed(
+            model="sentence-transformers/all-MiniLM-L6-v2")),
+    default_similarity_metric=cocoindex.VectorSimilarityMetric.COSINE_SIMILARITY
+)
```
badmonster0, line 17: We can replace this lambda by `cocoindex_funs.code_to_embedding` (in the original example, sharing the embedding model setup between the query handler and the flow definition is the main motivation of having `code_to_embedding` as a separate function)

badmonster0 (approved):
Thanks a lot for creating this!

examples/fastapi_server_docker/main.py
AbdelRahmanYaghi: Ah, I was wondering why these were separate in the original example, but didn't want to change it because I did not know the reason behind it. Please check whether I satisfied what you are requiring and update me.

Also, I've added the example to the README. And fixed something in the README file for the example.

badmonster0: looks good! thanks!

badmonster0 (approved):

badmonster0:
@AbdelRahmanYaghi we appreciate your contribution and made a section for you on the latest release note https://cocoindex.io/blogs/cocoindex-changelog-2025-05-31/#abdelrahmanyaghi

PR #442, opened 2025-05-07
feat(pyo3): implement flow spec pretty print and add verbose mode

### LINKED ISSUE #410: [FEATURE] Implement flow spec pretty print in Rust
In #404 we started to pretty print flow spec in `cocoindex show`, which is nice.

We want to further consolidate the related logic into Rust, so that
- The logic can be shared by more components (e.g. we'll have JS and Rust SDK in the future)
- It has direct access to the original struct of `FlowSpec`, so potentially do even more customized improvements.

---

lemorage:
Following #427 and #404, this PR implements the flow spec pretty print in Rust and adds `--verbose` mode to enable more detailed output.

#### Changes on Rust side
1. We first port the code from #404, using Rust, and then rewrite it using idiomatic Rust style, largely via `Display` trait.
2. We add verbose mode to the spec getter method.

#### Changes on Python side
1. We change it to use `get_spec` from Rust side.
2. We use `rich.Tree` instead of `rich.Console` for clearer readability.

#### Preview
`python main.py cocoindex show DocsToKG`

image

After adding `--verbose` flag, it will then produce a more detailed output with full operation specs, including all fields and JSON-like formatting, similar to preview in #404.

Resolves #410.

### DIFF
--- a/python/cocoindex/flow.py
+++ b/python/cocoindex/flow.py
@@ -458,38 +458,19 @@ class Flow:
         """
         Render the flow spec as a styled rich Tree with hierarchical structure.
         """
+        spec = self._get_spec(verbose=verbose)
         tree = Tree(f"Flow: {self.name}", style="cyan")
-        current_section = None
-        section_node = None
-        indent_stack = []
 
-        for i, (section, content, indent) in enumerate(self._get_spec(verbose=verbose)):
-            # Skip "Scope" entries (see ReactiveOpScope in spec.rs)
-            if content.startswith("Scope:"):
-                continue
+        def build_tree(label: str, lines: list):
+            node = Tree(label, style="bold magenta" if lines else "cyan")
+            for line in lines:
+                child_node = node.add(Text(line.content, style="yellow"))
+                child_node.children = build_tree("", line.children).children
+            return node
 
-            if section != current_section:
-                current_section = section
-                section_node = tree.add(f"{section}:", style="bold magenta")
-                indent_stack = [(0, section_node)]
-
-            while indent_stack and indent_stack[-1][0] >= indent:
-                indent_stack.pop()
-
-            parent = indent_stack[-1][1] if indent_stack else section_node
-            styled_content = Text(content, style="yellow")
-            is_parent = any(
-                next_indent > indent
-                for _, next_content, next_indent in self._get_spec(verbose=verbose)[i + 1:]
-                if not next_content.startswith("Scope:")
-            )
-
-            if is_parent:
-                node = parent.add(styled_content, style=None)
-                indent_stack.append((indent, node))
-            else:
-                parent.add(styled_content, style=None)
-
+        for section, lines in spec.sections:
+            section_node = build_tree(f"{section}:", lines)
+            tree.children.append(section_node)
         return tree
 
     def _get_spec(self, verbose: bool = False) -> list[tuple[str, str, int]]:
--- a/python/cocoindex/flow.py
+++ b/python/cocoindex/flow.py
@@ -15,7 +15,7 @@ from threading import Lock
 from enum import Enum
 from dataclasses import dataclass
 from rich.text import Text
-from rich.console import Console
+from rich.tree import Tree
 
 from . import _engine
 from . import index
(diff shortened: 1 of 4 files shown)


src/py/mod.rs
```diff
@@ -211,7 +316,7 @@ impl Flow {
                 let mut field_type = match &field.value_type.typ {
                     ValueType::Basic(basic) => format!("{}", basic),
                     ValueType::Table(t) => format!("{}", t.kind),
-                    ValueType::Struct(_) => "Struct".to_string(),
+                    ValueType::Struct(s) => format!("{}", s),
```
badmonster0: seems this change is unexpected? We don't want to show `Struct(title: Str, summary: Str)` given the sub fields will be shown in separate lines below.

src/py/mod.rs
```diff
@@ -196,6 +197,110 @@ impl Flow {
...
+
+        // Header
+        result.push(("Header".into(), format!("Flow: {}", spec.name), 0));
+
+        // Sources
+        extend_with(&mut result, "Sources", spec.import_ops.iter(), |op| {
+            format!(
+                "Import: name={}, {}",
+                op.name,
+                if verbose {
```
badmonster0: Maybe we can pass this as an argument down to the `format()` method for each one, so that code branching only need to happen in a deeper level when needed.

(for the argument, maybe consider define an enum, for ease of extension in the future and better type safety)

src/base/spec.rs
```diff
@@ -155,18 +199,94 @@ pub struct OpArgBinding {
...
+                        .iter()
+                        .filter_map(|v| v.as_str())
+                        .collect::<Vec<_>>()
+                        .join(",");
+                    if !items.is_empty() {
+                        parts.push(format!("{}={}", key, items));
+                    }
+                }
+                Value::Object(obj) => {
+                    if let Some(model) = obj.get("model").and_then(|v| v.as_str()) {
```
badmonster0: This is a little bit ad-hoc and hard to maintain.

How about this:
- In concise mode, don't show the specific content of the spec.
- In verbose mode, always show the full spec.

badmonster0 (commented):
Beautiful! Thanks a lot!

src/py/mod.rs
lemorage: ahh, likely forgot to clean this up

src/base/spec.rs
lemorage: thanks, sounds good to me

src/py/mod.rs
```diff
@@ -196,6 +198,78 @@ impl Flow {
         })
     }
 
+    #[pyo3(signature = (output_mode="concise"))]
+    pub fn get_spec(&self, output_mode: &str) -> PyResult<Vec<(String, String, u32)>> {
```
badmonster0: For `output_mode`, we may use type `Pythonized<OutputMode>` (and also add a `Deserialize` for its derive), so we won't need to manually parse it from string.

python/cocoindex/flow.py
```diff
@@ -454,61 +454,52 @@ def _lazy_engine_flow() -> _engine.Flow:
...
-    def _format_flow(self, flow_dict: dict) -> Text:
-        output = Text()
+    def _render_spec(self, verbose: bool = False) -> Tree:
+        """
+        Render the flow spec as a styled rich Tree with hierarchical structure.
+        """
+        tree = Tree(f"Flow: {self.name}", style="cyan")
+        current_section = None
+        section_node = None
+        indent_stack = []
```
badmonster0: Now on the Rust side we're walking the tree recursively and output the flattened result, but here we're using this stack trying to restore the hierarchies, which seems a little bit detour.

I'm wondering if we can simply build a recursive data structure like the following from Rust side:

```
struct RenderedSpecLine {
  content: String,
  children: Vec<RenderedSpecLine>,
}

struct RenderedSpec {
  sections: Vec<(String, Vec<RenderedSpecLine>)>
}
```

Then on Python side we can traverse this data structure recursively. I think the logic here will be more straightforward and easier to maintain, e.g. we don't need to compare indent, matching against string `"Scope:"` etc.

What do you think?

python/cocoindex/flow.py
```diff
@@ -454,61 +454,52 @@ def _lazy_engine_flow() -> _engine.Flow:
...
 
-        def add_line(content, indent=0, style=None, end="\n"):
-            output.append(" " * indent)
-            output.append(content, style=style)
-            output.append(end)
+        for i, (section, content, indent) in enumerate(self._get_spec(verbose=verbose)):
+            # Skip "Scope" entries (see ReactiveOpScope in spec.rs)
+            if content.startswith("Scope:"):
```
badmonster0: Note that the scope name is actually quite useful (ops under this scope will use it to reference the current row, e.g. `documents_1.content` in your screenshot).

src/base/spec.rs
```diff
@@ -1,8 +1,35 @@
 use crate::prelude::*;
 
 use super::schema::{EnrichedValueType, FieldSchema};
+use serde::{Deserialize, Serialize};
+use std::fmt;
 use std::ops::Deref;
 
+/// OutputMode enum for displaying spec info in different granularity
+#[derive(Debug, Clone, Copy, PartialEq)]
```
badmonster0: We can also add `Eq` to it.

python/cocoindex/flow.py
lemorage: umm, that would be nicer. I did not remove the old python traversing for simplicity, but your suggested way would make more sense, I'll try that later!

lemorage:
Thanks for all the reviews from @badmonster0. We made a lot of changes since my first comment, so I will try to give a whole summary of them here. 

1. We made an enum for distinct output modes in Rust: verbose or concise.
2. We made a trait `SpecFormatter` to enforce the spec formatting logic per output mode for each `opSpec`.
3. We made two structs `RenderedSpecLine` and `RenderedSpec`, to handle the spec line rendering logic on Rust side, so we can grab those and structure the content via `rich.Tree` simply on Python side.

So now this is what `cocoindex show` looks like on the terminal. Basically, for verbose mode, all the nested contents will be unfolded (e.g., a JSON struct), but in concise mode, not.

> The following output can be acquired by running `python main.py cocoindex show` under [`cocoindex/examples/text_embedding`](https://github.com/cocoindex-io/cocoindex/tree/main/examples/text_embedding).
cocoindex show in different modes

src/py/mod.rs
```diff
@@ -196,6 +198,96 @@ impl Flow {
...
+            "Header".to_string(),
+            vec![RenderedSpecLine {
+                content: format!("Flow: {}", spec.name),
+                scope: None,
+                children: vec![],
+            }],
+        );
+        for key in ["Sources", "Processing", "Targets", "Declarations"] {
+            sections.insert(key.to_string(), Vec::new());
+        }
```
badmonster0: This seems unnecessary as we already use `or_default()` below everywhere.

src/py/mod.rs
```diff
@@ -196,6 +198,96 @@ impl Flow {
...
+        for op in &spec.import_ops {
+            sections
+                .entry("Sources".to_string())
+                .or_default()
+                .push(RenderedSpecLine {
+                    content: format!("Import: name={}, {}", op.name, op.spec.format(mode)),
+                    scope: None,
+                    children: vec![],
+                });
+        }
```
badmonster0: This might be a little bit preferable in Rust, as we don't introduce mutable variables and avoids to access the map entry with a new string each time in the loop:

```
sections.insert(
  "Source".to_string(),
  spec.import_ops.iter().map(|op| {
    RenderedSpecLine {
      content: format!("Import: name={}, {}", op.name, op.spec.format(mode)),
      scope: None,
      children: vec![],
    }
  }).collect()
)
```
 
Similar below.

src/base/spec.rs
```diff
@@ -1,8 +1,45 @@
...
+    Verbose,
+}
+
+/// Formatting spec per output mode
+pub trait SpecFormatter {
+    fn format(&self, mode: OutputMode) -> String;
+}
+
+/// A single line in the rendered spec, with optional scope and children
+#[pyclass(get_all, set_all)]
```
badmonster0: Since this is python specific, maybe move these to the `py` mod.

We may add a new file there, e.g. `py/render.rs` (together with other rendering logic)

src/base/spec.rs
```diff
@@ -1,8 +1,45 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RenderedSpecLine {
+    /// The formatted content of the line (e.g., "Import: name=documents, source=LocalFile")
+    pub content: String,
+    /// The scope name, if applicable (e.g., "documents_1" for ForEach scopes)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub scope: Option<String>,
```
badmonster0: Seems we can put this information into `content` just like all others? This field seems not used by the python code any way.

src/py/mod.rs
lemorage: Thanks, makes sense!

src/base/spec.rs
lemorage: Thanks! I moved all the renderedSpec-related pyclass into `mod.rs` but did not create a new file. I found that the rendering logic was in the `get_spec` function, which is relatively compact. So I did not find many things that could be moved to a new file. If you have any other views on this, I'm glad to hear!

badmonster0 (approved):

badmonster0:
Thanks for the wonderful PR!

PR #444, opened 2025-05-07
feat(yaml): store variant name in tuple and struct variants

### LINKED ISSUE #443: [FEATURE] Keep Variant Name in YAML Serialization
**What is the use case?**
For the current implementation of `YamlSerializer::serialize_tuple_variant` and `YamlSerializer::serialize_struct_variant`, **only** the data of the variant will be outputted. The name of the variant itself is lost in the YAML.

For example:

```rust
enum MyStuff {
    OptionA(String, i32),         // Tuple Variant
    OptionB { user: String, id: u64 } // Struct Variant
}
```

Right now, if we serialize `MyStuff::OptionA("hey", 123)`, the YAML just looks like:

```yaml
- hey
- 123
```

And for `MyStuff::OptionB { user: "me", id: 42 }`, it's like:

```yaml
user: me
id: 42
```

All the actual variant names ("OptionA", "OptionB") are missing from the YAML.

Most YAML serializers for enums make sure to include the variant's name so the data makes sense on its own, and I think here we should do the same.

We probably need to tweak our `SeqSerializer` and `MapSerializer` to *remember* that variant name. Then, when they're done serializing all the data bits, they should wrap it all up, like:

For `OptionA("hey", 123)`:
```yaml
OptionA:
  - hey
  - 123
```

And for `OptionB { user: "me", id: 42 }`:
```yaml
OptionB:
  user: me
  id: 42
```

---

lemorage:
Previously, in our implementations of yaml serialization, the `serialize_tuple_variant` and `serialize_struct_variant` methods only serialized the data within the variant. They don't handle the variant's name.

https://github.com/cocoindex-io/cocoindex/blob/cf567ce8eb73053a8e539253ce125fb7f391f1ff/src/utils/yaml_ser.rs#L181-L191

https://github.com/cocoindex-io/cocoindex/blob/cf567ce8eb73053a8e539253ce125fb7f391f1ff/src/utils/yaml_ser.rs#L208-L216

This PR modifies them to store variant name when serialized, closes #443.

### DIFF
--- a/src/utils/yaml_ser.rs
+++ b/src/utils/yaml_ser.rs
@@ -44,10 +44,10 @@ impl ser::Serializer for YamlSerializer {
     type SerializeSeq = SeqSerializer;
     type SerializeTuple = SeqSerializer;
     type SerializeTupleStruct = SeqSerializer;
-    type SerializeTupleVariant = SeqSerializer;
+    type SerializeTupleVariant = VariantSeqSerializer;
     type SerializeMap = MapSerializer;
     type SerializeStruct = MapSerializer;
-    type SerializeStructVariant = MapSerializer;
+    type SerializeStructVariant = VariantMapSerializer;
 
     fn serialize_bool(self, v: bool) -> Result<Self::Ok, Self::Error> {
         Ok(Yaml::Boolean(v))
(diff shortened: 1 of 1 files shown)


badmonster0 (approved):

badmonster0:
Thanks for fixing this!

PR #452, opened 2025-05-09
fix: unit tests for python values

### LINKED ISSUE #438: Unittest for Value Conversion
Add Unittest for 
https://github.com/cocoindex-io/cocoindex/blob/main/src/py/convert.rs

Convert between Engine value and Rust Value. 

RoundTrip - test Property Rust value -> Python value -> Rust value should be the same

Good first issue if you are picking up rust.

---

TwistingTwists:
fixes #438

### DIFF
--- a/src/py/convert.rs
+++ b/src/py/convert.rs
@@ -226,3 +226,89 @@ pub fn value_from_py_object<'py>(
     };
     Ok(result)
 }
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::base::{schema, value};
+    use pyo3::Python;
+    use std::sync::Arc;
+
+    #[test]
+    fn test_roundtrip_basic_values() {
+        Python::with_gil(|py| {
+            // Test Int64
+            let int_value = value::Value::Basic(value::BasicValue::Int64(42));
+            let int_type = schema::ValueType::Basic(schema::BasicValueType::Int64);
+            let py_obj = value_to_py_object(py, &int_value).unwrap();
+            let roundtrip_value = value_from_py_object(&int_type, &py_obj).unwrap();
+            assert_eq!(int_value, roundtrip_value, "Int64 roundtrip failed");
+
+            // Test String
+            let str_value = value::Value::Basic(value::BasicValue::Str(Arc::from("test string")));
+            let str_type = schema::ValueType::Basic(schema::BasicValueType::Str);
+            let py_obj = value_to_py_object(py, &str_value).unwrap();
+            let roundtrip_value = value_from_py_object(&str_type, &py_obj).unwrap();
+            assert_eq!(str_value, roundtrip_value, "String roundtrip failed");
+
+            // Test Bool
+            let bool_value = value::Value::Basic(value::BasicValue::Bool(true));
+            let bool_type = schema::ValueType::Basic(schema::BasicValueType::Bool);
+            let py_obj = value_to_py_object(py, &bool_value).unwrap();
+            let roundtrip_value = value_from_py_object(&bool_type, &py_obj).unwrap();
+            assert_eq!(bool_value, roundtrip_value, "Bool roundtrip failed");
+        });
+    }
+
+    #[test]
+    fn test_roundtrip_struct() {
+        Python::with_gil(|py| {
+            // Create a struct schema with multiple fields
+            let struct_schema = schema::StructSchema {
+                description: Some(Arc::from("Test struct")),
+                fields: Arc::new(vec![
+                    schema::FieldSchema {
+                        name: "id".into(),
+                        value_type: schema::EnrichedValueType {
+                            typ: schema::ValueType::Basic(schema::BasicValueType::Int64),
+                            nullable: false,
+                            attrs: Default::default(),
+                        },
+                    },
+                    schema::FieldSchema {
+                        name: "name".into(),
+                        value_type: schema::EnrichedValueType {
+                            typ: schema::ValueType::Basic(schema::BasicValueType::Str),
+                            nullable: false,
+                            attrs: Default::default(),
+                        },
+                    },
+                    schema::FieldSchema {
+                        name: "active".into(),
+                        value_type: schema::EnrichedValueType {
+                            typ: schema::ValueType::Basic(schema::BasicValueType::Bool),
+                            nullable: false,
+                            attrs: Default::default(),
+                        },
+                    },
+                ]),
+            };
+
+            // Create a struct value matching the schema
+            let struct_value = value::Value::Struct(value::FieldValues {
+                fields: vec![
+                    value::Value::Basic(value::BasicValue::Int64(1)),
+                    value::Value::Basic(value::BasicValue::Str(Arc::from("test"))),
+                    value::Value::Basic(value::BasicValue::Bool(true)),
+                ],
+            });
+
+            // Perform roundtrip conversion
+            let struct_type = schema::ValueType::Struct(struct_schema);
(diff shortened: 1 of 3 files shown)


src/py/convert.rs
```diff
@@ -226,3 +226,89 @@ pub fn value_from_py_object<'py>(
...
+    use std::sync::Arc;
+
+    #[test]
+    fn test_roundtrip_basic_values() {
+        Python::with_gil(|py| {
+            // Test Int64
+            let int_value = value::Value::Basic(value::BasicValue::Int64(42));
+            let int_type = schema::ValueType::Basic(schema::BasicValueType::Int64);
+            let py_obj = value_to_py_object(py, &int_value).unwrap();
+            let roundtrip_value = value_from_py_object(&int_type, &py_obj).unwrap();
```
badmonster0: `value_from_py_object` and `value_to_py_object` are functions internal to this module. `Pythonized` is the API exposed by the module. Ideally we want to test on the behavior of the public API.

You can also consider create a helper function as testutil to do this roundtrip conversion and validation (e.g. take a type + a value as input, do the roundtrip conversion internally and assert on equality), so that there will be less boilerplates in each specific test.

src/py/convert.rs
```diff
@@ -226,3 +226,89 @@ pub fn value_from_py_object<'py>(
...
+                ],
+            });
+
+            // Perform roundtrip conversion
+            let struct_type = schema::ValueType::Struct(struct_schema);
+            let py_obj = value_to_py_object(py, &struct_value).unwrap();
+            let roundtrip_value = value_from_py_object(&struct_type, &py_obj).unwrap();
+            assert_eq!(struct_value, roundtrip_value, "Struct roundtrip failed");
+        });
+    }
```
badmonster0, line 523: Can we also cover Table types? Thanks!

badmonster0 (commented):
Thanks for adding this test! Appreciate the effort!

src/py/convert.rs
TwistingTwists: Yes. Surely!

src/py/convert.rs
TwistingTwists: Ah. Right. Thanks for pointing me in right direction!

TwistingTwists:
@badmonster0 can you please review this once?

src/py/convert.rs
```diff
@@ -226,3 +230,269 @@ pub fn value_from_py_object<'py>(
...
+    use crate::base::value;
+    use crate::base::value::ScopeValue;
+    use pyo3::Python;
+    use std::collections::BTreeMap;
+    use std::sync::Arc;
+
+    fn assert_roundtrip_conversion(original_value: &value::Value, value_type: &schema::ValueType) {
+        Python::with_gil(|py| {
+            // Convert Rust value to Python object
+            let pythonized_value = Pythonized(original_value.clone());
```
badmonster0: Actually I think I made a mistake in my last round of comments. Apologize for that..

There're actually two sets of conversion mechanism defined in this package:
- For conversion of some meta objects, e.g. configs (settings, specs, etc.), `Pythonized` is used.
- For conversion of data values in CocoIndex, `value_to_py_object()` and `value_from_py_object()` are used.

Really sorry for the confusion. Your previous code was actually correct in using `value_from_py_object()` and `value_to_py_object()`.

With these, we don't need to add `Deserialize` for existing types.

src/py/convert.rs
```diff
@@ -226,3 +230,269 @@ pub fn value_from_py_object<'py>(
...
+                value_from_py_object(value_type, &py_object).unwrap_or_else(|e| {
+                    panic!(
+                        "Failed to convert Python object back to Rust value: {:?}",
+                        e
+                    )
+                });
+
+            println!("Roundtripped value: {:?}", roundtripped_value);
+            // Compare values
+            match (&original_value, &roundtripped_value) {
```
badmonster0: Since `PartialEq` is already provided for value types, I think we can directly compare these by `==` or `assert_eq!`?

src/settings.rs
```diff
@@ -1,13 +1,13 @@
 use serde::Deserialize;
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Clone)]
```
badmonster0: is `Clone` here needed by any new code?

badmonster0 (commented):
Hi, sorry I missed your reply a few days ago, and just notice.

It's very close! Thanks for your patience.

src/py/convert.rs
TwistingTwists: Ah. Thanks for clarification. I'll revert related changes. to use `value_to_py_object` and `value_from_py_object`

src/py/convert.rs
TwistingTwists: Will look into it. Did not realise that there is `PartialEq` already.

badmonster0:
Hi @TwistingTwists, are you still on this? I noticed the existing changes are already very close. Just want to check if you're still on it - it's fine to take more time. Thanks!

TwistingTwists:
Yes. last week was busy. Will update over weekend.

TwistingTwists:
@badmonster0 ready for review. let me know if this looks good.

src/server.rs
```diff
@@ -8,7 +8,7 @@ use tower_http::{
     trace::TraceLayer,
 };
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Clone)]
```
badmonster0: Is this `Clone` needed by any new code?

src/settings.rs
badmonster0: Same here

src/py/convert.rs
```diff
@@ -226,3 +230,207 @@ pub fn value_from_py_object<'py>(
...
+    use crate::base::value;
+    use crate::base::value::ScopeValue;
+    use pyo3::Python;
+    use std::collections::BTreeMap;
+    use std::sync::Arc;
+
+    fn assert_roundtrip_conversion(original_value: &value::Value, value_type: &schema::ValueType) {
+        Python::with_gil(|py| {
+            // Convert Rust value to Python object using value_to_py_object
+            let py_object = value_to_py_object(py, original_value).unwrap_or_else(|e| {
```
badmonster0: It's easier to do this with `expect`:

```rs
let py_object = value_to_py_object(py, original_value)
    .expect("Failed to convert Rust value to Python object");
```

src/py/convert.rs
```diff
@@ -226,3 +230,207 @@ pub fn value_from_py_object<'py>(
...
+                panic!("Failed to convert Rust value to Python object: {:?}", e)
+            });
+
+            println!("Python object: {:?}", py_object);
+            let roundtripped_value =
+                value_from_py_object(value_type, &py_object).unwrap_or_else(|e| {
+                    panic!(
+                        "Failed to convert Python object back to Rust value: {:?}",
```
badmonster0: same here, we can use `expect`

TwistingTwists: updated.

src/py/convert.rs
TwistingTwists: updated.

src/server.rs
TwistingTwists: removed

src/settings.rs
TwistingTwists: removed

badmonster0:
hi @TwistingTwists 
Thank you for your contribution, we made a section for you in the latest release note
https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#twistingtwists

thanks!

PR #459, opened 2025-05-09
Enhance Image Search Example - Live Update

par4m:
Enhances Image Search Example by adding Live Update
- Threading: Runs the live updater in the background,separate from FastAPI’s main thread.
- Asyncio: Runs the async updater logic


examples/image_search_example/main.py
```diff
@@ -101,6 +105,25 @@ def startup_event():
...
+                async with updater:
+                    print("[LiveUpdater] Running updater.wait()...")
+                    await updater.wait()
+                print("[LiveUpdater] Exited updater context.")
+            except Exception as e:
+                print(f"[LiveUpdater] Exception: {e}")
+        print("[LiveUpdater] Thread starting...")
+        asyncio.run(updater_coroutine())
+        print("[LiveUpdater] Thread finished.")
+    thread = threading.Thread(target=run_updater, daemon=True)
```
badmonster0: We don't need to create a separate thread. I think this is sufficient:

```
app.state.live_updater = cocoindex.FlowLiveUpdater(image_object_embedding_flow)
```

It will use a background thread (with Rust runtime) to watch changes automatically
(You can call `app.state.live_updater.abort()` to stop it, but usually not necessary for a long-running process)

badmonster0 (commented):
For `get_image_caption`, we can consider enable cache:

https://cocoindex.io/docs/core/custom_function#parameters-for-custom-functions

badmonster0:
With this change, the `README` can also be simplified. We don't need to run `cocoindex update` separately:


https://github.com/cocoindex-io/cocoindex/blob/cbc199823b14309f536831408768674c5285cf4c/examples/image_search_example/README.md?plain=1#L50

examples/image_search_example/main.py
badmonster0: I just updated the API, trying to make it more clear. [The doc](https://cocoindex.io/docs/core/flow_methods#library-api-1) is updated accordingly. Now an explicit `start()` is needed to actually start the change capturing, i.e. need to do this:

```
app.state.live_updater = cocoindex.FlowLiveUpdater(image_object_embedding_flow)
app.state.live_updater.start()
```

par4m: I tried with the new API, this is the implementation. For some reason I keep getting 
`TypeError: No constructor defined for FlowLiveUpdater` . Not sure why, does it have anything to do with the Rust binding? 

```python

# 1. Extract caption from image using Ollama vision model
@cocoindex.op.function(cache=True, behavior_version=1)
def get_image_caption(img_bytes: bytes) -> str:
    ...
    return text


# 3. CocoIndex flow: Ingest images, extract captions, embed, export to Qdrant
@cocoindex.flow_def(name="ImageObjectEmbedding")
def image_object_embedding_flow(flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope):
    data_scope["images"] = flow_builder.add_source(
        cocoindex.sources.LocalFile(path="img", included_patterns=["*.jpg", "*.jpeg", "*.png"], binary=True),
        refresh_interval=datetime.timedelta(minutes=1)  # Poll for changes every 1 minute
    )
    img_embeddings = data_scope.add_collector()
    with data_scope["images"].row() as img:
        img["caption"] = img["content"].transform(get_image_caption)
        img["embedding"] = caption_to_embedding(img["caption"])
        img_embeddings.collect(
            id=cocoindex.GeneratedField.UUID,
            filename=img["filename"],
            caption=img["caption"],
            embedding=img["embedding"],
        )
    img_embeddings.export(
        "img_embeddings",
        cocoindex.storages.Qdrant(
            collection_name="image_search",
            grpc_url=os.getenv("QDRANT_GRPC_URL", "http://localhost:6334/"),
        ),
        primary_key_fields=["id"],
        setup_by_user=True,
    )


# --- CocoIndex initialization on startup ---
@app.on_event("startup")
def startup_event():
    settings = cocoindex.setting.Settings.from_env()
    cocoindex.init(settings)
    app.state.query_handler = cocoindex.query.SimpleSemanticsQueryHandler(
        name="ImageObjectSearch",
        flow=image_object_embedding_flow,
        target_name="img_embeddings",
        query_transform_flow=caption_to_embedding,
        default_similarity_metric=cocoindex.VectorSimilarityMetric.COSINE_SIMILARITY,
    )
    app.state.live_updater = cocoindex.FlowLiveUpdater(image_object_embedding_flow)
    app.state.live_updater.start()
    
```

badmonster0: Ah, a previous version indeed has a bug ([here](https://github.com/cocoindex-io/cocoindex/blob/445919229e1c85bde4faf092b8183a8bdb66cee1/python/cocoindex/flow.py#L393) we should call `.create` instead of the constructor).

Can you try on the latest version? The bug has been fixed in v0.1.35

par4m: It worked now! Should I change the README or description to include Live Update

badmonster0: Nice! Yes, it'll nice to also change README. Thanks!

badmonster0 (approved):

badmonster0:
Awesome! This example is more fantastics now! Thank you very much!!

PR #462, opened 2025-05-10
feat: support namedtuple for struct types

### LINKED ISSUE #437: [FEATURE] Support NamedTuple in addition to dataclass in Python SDK
Currently `dataclass` is the only way for users to define a  [`struct type`](https://cocoindex.io/docs/core/data_types#struct-type) in Python SDK.
[`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) is another possible way in Python to define structs and used by people and existing libraries. We want to support it too.

- [`typing.py`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/python/cocoindex/typing.py)
  - Extend [`AnalyzedTypeInfo`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/python/cocoindex/typing.py#L63) to be able to keep `NamedTuple` type info too.
  - Update [`analyze_type_info()`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/python/cocoindex/typing.py#L77) accordingly
- [`convert.py`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/python/cocoindex/convert.py). Both encode/decode logic needs to be updated, to support `NamedTuple` in addition to `dataclass`.
- Add test in [`test_convert.py`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/python/cocoindex/tests/test_convert.py) to cover `NamedTuple` caes.
- Update the document [`data_types.mdx`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/docs/docs/core/data_types.mdx)

---

lemorage:
This PR tries to resolve #437.

`NamedTuples` support is added, so now struct types can be defined either by a dataclass or a namedtuple. A bunch of unit tests added for verifying the functionality and the docs is updated correpondingly.

### DIFF
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -105,7 +105,7 @@ In Python, a KTable type is represented by `dict[K, V]`.
 The `V` should be a struct type, either a `dataclass` or `NamedTuple`, representing the value fields of each row.
 For example, you can use `dict[str, Person]` or `dict[str, PersonTuple]` to represent a KTable, with 4 columns: key (Str), `first_name` (Str), `last_name` (Str), `dob` (Date).
 
-Note that if you want to use a struct as the key, you need to annotate the struct with `@dataclass(frozen=True)`, For `NamedTuple`, immutability is built-in. so the values are immutable.
+Note that if you want to use a struct as the key, you need to ensure the struct is immutable. For `dataclass`, annotate it with `@dataclass(frozen=True)`. For `NamedTuple`, immutability is built-in.
 For example:
 
 ```python
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -53,22 +53,40 @@ The native Python type is always more permissive and can represent a superset of
     you can choose whatever to use.
     The native Python type is usually simpler.
 
-### Struct Type
+### Struct Types
 
 A Struct has a bunch of fields, each with a name and a type.
 
-In Python, a Struct type is represented by a [dataclass](https://docs.python.org/3/library/dataclasses.html),
-and all fields must be annotated with a specific type. For example:
+In Python, a Struct type is represented by either a [dataclass](https://docs.python.org/3/library/dataclasses.html)
+or a [NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple), with all fields annotated with a specific type.
+Both options define a structured type with named fields, but they differ slightly:
+
+- **Dataclass**: A flexible class-based structure, mutable by default, defined using the `@dataclass` decorator.
+- **NamedTuple**: An immutable tuple-based structure, defined using `typing.NamedTuple`.
+
+For example:
 
 ```python
 from dataclasses import dataclass
+from typing import NamedTuple
+import datetime
 
+# Using dataclass
 @dataclass
 class Person:
     first_name: str
-    last_name
+    last_name: str
+    dob: datetime.date
+
+# Using NamedTuple
+class PersonTuple(NamedTuple):
+    first_name: str
+    last_name: str
     dob: datetime.date
 ```
+
+Both `Person` and `PersonTuple` are valid Struct types in CocoIndex, with identical schemas (three fields: `first_name` (Str), `last_name` (Str), `dob` (Date)).
+Choose `dataclass` for mutable objects or when you need additional methods, and `NamedTuple` for immutable, lightweight structures.
 
 ### Table Types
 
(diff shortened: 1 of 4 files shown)


fivemeyestoreserver:
This is a well-structured codebase. I appreciate the clean separation of concerns.

eyestoreoye:
Impressive project structure! The architecture really demonstrates solid design principles.

ayestoro:
Interesting solution to this problem! I've been working on similar challenges recently.

eyestorefivemstore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

docs/docs/core/data_types.mdx
```diff
@@ -84,20 +102,24 @@ The row order of a KTable is not preserved.
 Type of the first column (key column) must be a [key type](#key-types).
 
 In Python, a KTable type is represented by `dict[K, V]`. 
-The `V` should be a dataclass, representing the value fields of each row.
-For example, you can use `dict[str, Person]` to represent a KTable, with 4 columns: key (Str), `first_name` (Str), `last_name` (Str), `dob` (Date).
+The `V` should be a struct type, either a `dataclass` or `NamedTuple`, representing the value fields of each row.
+For example, you can use `dict[str, Person]` or `dict[str, PersonTuple]` to represent a KTable, with 4 columns: key (Str), `first_name` (Str), `last_name` (Str), `dob` (Date).
 
-Note that if you want to use a struct as the key, you need to annotate the struct with `@dataclass(frozen=True)`, so the values are immutable.
+Note that if you want to use a struct as the key, you need to annotate the struct with `@dataclass(frozen=True)`, For `NamedTuple`, immutability is built-in. so the values are immutable.
```
badmonster0: nit: move the part "so the values are immutable." after "you need to annotate the struct with `@dataclass(frozen=True)`,"

python/cocoindex/typing.py
```diff
@@ -70,6 +73,7 @@ class AnalyzedTypeInfo:
 
     key_type: type | None           # For element of KTable
     dataclass_type: type | None     # For Struct
+    namedtuple_type: type | None    # For Struct
```
badmonster0: It looks like we always use `or` when we access `dataclass_type` / `namedtuple_type`. No field is accessed separately. So maybe make code slightly simpler by using one field instead of two, e.g.

```
struct_type: type | None  # For Struct, a dataclass or namedtuple
```

python/cocoindex/convert.py
```diff
@@ -96,11 +101,29 @@ def decode(value):
...
+    is_namedtuple = is_namedtuple_type(dst_struct_type)
+    
+    if is_dataclass:
+        parameters = inspect.signature(dst_struct_type).parameters
+    elif is_namedtuple:
+        parameters = {
+            name: inspect.Parameter(
+                name=name,
+                kind=inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                annotation=dst_struct_type.__annotations__.get(name, inspect.Parameter.empty)
```
badmonster0, line 119: We probably also want to unpack the default values from `_field_defaults` if exists:

https://docs.python.org/3/library/collections.html#collections.somenamedtuple._field_defaults

python/cocoindex/convert.py
```diff
@@ -119,10 +142,14 @@ def make_closure_for_value(name: str, param: inspect.Parameter) -> Callable[[lis
...
-    return lambda values: dst_dataclass_type(
-        *(decoder(values) for decoder in field_value_decoder))
+        for (name, param) in parameters.items()]
+
+    if is_dataclass:
+        return lambda values: dst_struct_type(
+            *(decoder(values) for decoder in field_value_decoder))
+    else:  # namedtuple
+        return lambda values: dst_struct_type(
+            *(decoder(values) for decoder in field_value_decoder))
```
badmonster0: Looks like these two branches are doing exactly the same thing.

badmonster0 (commented):
Thanks for making this PR! Really appreciate the speed and code quality!

docs/docs/core/data_types.mdx
lemorage: Ah!

python/cocoindex/typing.py
lemorage: Thanks for the suggestion! I was thinking the same when doing this part, but for consistency, did not take this jumping step. I'll do this refactor shortly.

PR #475, opened 2025-05-12
support s3 as native source

badmonster0:


src/ops/sources/s3.rs
```diff
@@ -0,0 +1,211 @@
...
+impl SourceExecutor for Executor {
+    fn list<'a>(
+        &'a self,
+        _options: &'a SourceExecutorListOptions,
+    ) -> BoxStream<'a, Result<Vec<SourceRowMetadata>>> {
+        let client = self.client.clone();
+        let bucket = self.bucket_name.clone();
+        let prefix = self.prefix.clone();
+        let included_glob_set = self.included_glob_set.clone();
+        let excluded_glob_set = self.excluded_glob_set.clone();
```
badmonster0: The returned BoxStream has the same lifetime `'a` as `self`, so seems we don't need to clone them.

src/ops/sources/s3.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+pub struct Factory;
+
+#[async_trait]
+impl SourceFactoryBase for Factory {
+    type Spec = Spec;
+
+    fn name(&self) -> &str {
+        "S3"
```
badmonster0: Maybe call it `AmazonS3`, which is the official name and avoids potential name conflict. The same for file name (`amazon_s3`).

Cargo.toml
```diff
@@ -109,3 +109,5 @@ rand = "0.9.0"
 indoc = "2.0.6"
 owo-colors = "4.2.0"
 json5 = "0.4.1"
+aws-config = "1"
+aws-sdk-s3 = "1"
```
badmonster0: 1.85.0

Cargo.toml
```diff
@@ -109,3 +109,5 @@ rand = "0.9.0"
 indoc = "2.0.6"
 owo-colors = "4.2.0"
 json5 = "0.4.1"
+aws-config = "1"
```
badmonster0: 1.6.2

(if we add it by `cargo add aws-config`, cargo will put the latest released version automatically)

badmonster0:
all done

PR #485, opened 2025-05-14
feat: create standalone CLI for cocoindex

### LINKED ISSUE #417: [FEATURE] Create standalone cocoindex CLI
The current CocoIndex CLI is provided as part of users' application by `cocoindex.main_fn()` (as described [here](https://cocoindex.io/docs/core/initialization#packaged-main)).

We may want to create a standalone CLI. It will provide the following benefits:

- A little bit easier to type. e.g. `cocoindex update main.py` instead of `python main.py cocoindex update`.
- Provide a more natural entrypoint for global tasks, e.g. list all flows by `cocoindex ls`.
- Less intrusive for users own application logic - they can more freely take CLI arguments upon their requirements, e.g. if their CLI takes a filename as argument like `main.py $FILENAME`, in the current way `cocoindex` cannot be the filename.

Existing example of standalone CLI taking users module as input: [FastAPI CLI](https://fastapi.tiangolo.com/fastapi-cli/)

---

lemorage:
This is a breaking change, and may resolve #417.

We create a standalone CLI for cocoindex. For compatibility reasons, we mark `main_fn` as deprecated. So, users can still use `@cocoindex.main_fn()` in their programs, but when running `python main.py cocoindex <command>`, the deprecation warning will be shown instead.

Now, in the new CLI, we will use `cocoindex <command> [options] --app main.py` to run the CLI program directly.

A lot of examples and docs will then be outdated and need to be updated accordingly. I will start on those if all the changes in this PR are confirmed.


python/cocoindex/lib.py
```diff
@@ -19,59 +15,36 @@ def init(settings: setting.Settings):
 
 def start_server(settings: setting.ServerSettings):
     """Start the cocoindex server."""
-    flow.ensure_all_flows_built()
-    query.ensure_all_handlers_built()
```
badmonster0: Maybe still keep these two in `start_server`.

Although not promoted now, this is supposed to be some library functions that users can call directly from their app.

python/cocoindex/cli.py
```diff
@@ -1,27 +1,90 @@
...
+    except Exception as e:
+        raise click.ClickException(f"Failed to initialize CocoIndex library: {e}")
 
 @cli.command()
+@click.option(
+    '--app', 'app_path', required=False,
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True, resolve_path=True),
+    help="Path to the Python file defining flows."
+)
 @click.option(
```
badmonster0: Probably we don't need this `-a` now. If `app_path` not provided, we show all.

python/cocoindex/cli.py
```diff
@@ -1,27 +1,90 @@
...
+# Create ServerSettings lazily upon first call, as environment variables may be loaded from files, etc.
+COCOINDEX_HOST = 'https://cocoindex.io'
+
+def _load_user_app(app_path: str):
+    """Loads the user's application file as a module. Exits on failure."""
+    if not app_path:
+        click.echo("Internal Error: Application path not provided.", err=True)
+        sys.exit(1)
+
+    app_path = os.path.abspath(app_path)
```
badmonster0: I'm thinking maybe we should also allow users directly pass in a module name if the module is already installed.  This will be helpful for users applications distributed by pypi packages.

python/cocoindex/cli.py
```diff
@@ -1,27 +1,90 @@
...
+        click.echo(f"\nError: Failed processing flows/handlers from application.", err=True)
+        click.echo(f"Reason: {e}", err=True)
+        sys.exit(1)
+
 @click.group()
+@click.version_option(package_name="cocoindex", message="%(prog)s version %(version)s")
 def cli():
     """
-    CLI for Cocoindex.
+    CLI for Cocoindex. Requires --app for most commands.
```
badmonster0: Actually how about directly passing in this path without `--app`? This is also a more common pattern used by other popular CLIs like fastapi.

Also another common patten is using `<module_path>:<app_name>`. For CocoIndex, I think we can accept `<module_path>:<flow_name>`

badmonster0:
Really awesome! Thanks for making this happen!

python/cocoindex/cli.py
lemorage: Sounds more concise! I can update to this approach later.

python/cocoindex/cli.py
```diff
@@ -1,89 +1,217 @@
...
+    app_ref = parts[0]
+
+    if not app_ref:
+        raise click.BadParameter(
+            f"Application module/path part is missing or invalid in specifier: '{specifier}'. "
+            "Expected format like 'myapp.py' or 'myapp:MyFlow'.",
+            param_hint="APP_SPECIFIER"
+        )
+
+    if len(parts) > 1:
```
badmonster0: nit: early return on the simpler branch will reduce the nesting level of code, slightly simplify code structure

```py
if len(parts) = 1:
  return app_ref

# The remaining logic to handle `>1` case below
...
```

python/cocoindex/cli.py
```diff
@@ -1,89 +1,217 @@
...
+                f"Invalid format for flow name part ('{flow_ref_part}') in specifier '{specifier}'. "
+                "If a colon separates the application from the flow name, the flow name should typically be "
+                "a valid identifier (e.g., alphanumeric with underscores, not starting with a number).",
+                param_hint="APP_SPECIFIER"
+            )
+        return app_ref, flow_ref_part
+    else:
+        return app_ref, None
+
+def _load_user_app(app_target: str):
```
badmonster0: nit: annotate the return type for clarity

python/cocoindex/cli.py
```diff
@@ -1,89 +1,217 @@
...
+    Exits on failure.
+    """
+    if not app_target:
+        raise click.ClickException("Application target not provided.")
+
+    looks_like_path = os.sep in app_target or app_target.lower().endswith(".py")
+
+    if looks_like_path or os.path.isfile(app_target):
+        if not os.path.isfile(app_target):
+            if looks_like_path and not os.path.exists(app_target):
```
badmonster0: `os.path.isfile()` already verifies if it the file exists: https://docs.python.org/3/library/os.path.html#os.path.isfile
(The main difference between it and `exists()` is: it also checks it's not a directory, special file like sockets, etc.)

So I think we can simplify the check above. Probably this is enough:

```python
if looks_like_path:
  if not os.path.isfile(app_target):
    raise ...
  # do the normal logic
```

python/cocoindex/cli.py
```diff
@@ -1,89 +1,217 @@
...
+        suffix = ''
+
+        if is_in_current_app:
+            if is_persisted:
+                suffix = ''  # Defined in app and setup exists
+            else:
+                suffix = ' [+]'  # Defined in app, setup missing
+                has_missing_setup = True
+        elif is_persisted:
+            suffix = ' [?]'  # Setup exists in backend, not defined in loaded app
```
badmonster0: This is what I was thinking originally:

1. If app_target is provided, only show flows in the current app_target (with or without a `[+]` to indicate if it exists in backend)
2. Otherwise, show all flows in the backend.

For "1", we should skip `[?]` case. For "2", there's no "current loaded app" so no need to have `[?]` too.

WDYT?

python/cocoindex/cli.py
```diff
@@ -1,89 +1,217 @@
...
-def setup():
+@click.argument("app_target", type=str)
+def setup(app_target: str):
     """
     Check and apply backend setup changes for flows, including the internal and target storage
     (to export).
+
+    APP_TARGET: path/to/app.py or installed_module.
     """
+    app_ref, _ = _parse_app_flow_specifier(app_target) # Ignore flow name for setup
```
badmonster0: I'm thinking for cases that we ignore flow name, we may wrap it in another util function, and we print out a warning if users provide it

badmonster0 (commented):
Another thing I just notice - which will be needed by migrating existing examples - is about loading of the `.env` files.

Previously they're loaded by examples' main module (under the `if __name__ == "__main__":` condition), but they're no longer executed when we invoke this standalone CLI.

So I think this standalone CLI may need to be able to load `.env` file, and we may also add a `--env_file` flag to allow override (example: [flask CLI](https://flask.palletsprojects.com/en/stable/cli/#environment-variables-from-dotenv) is doing the similar).

Let me know if you have any other thoughts. Thanks!

lemorage:
Nice catch!

python/cocoindex/cli.py
```diff
@@ -1,89 +1,233 @@
...
+    show_default=False
+)
+def cli(env_file: str | None):
     """
     CLI for Cocoindex.
     """
+    loaded_env_path = None
+    dotenv_path = env_file or os.path.join(os.getcwd(), ".env")
+
+    if load_dotenv(dotenv_path=dotenv_path, override=True):
```
badmonster0: Let's remove `override=True` here, i.e. stick to `load_dotenv()`'s default behavior to avoid surprise (actually the existing examples shouldn't have it too)

python/cocoindex/cli.py
```diff
@@ -1,89 +1,233 @@
...
 @click.option("--color/--no-color", default=True, help="Enable or disable colored output.")
 @click.option("--verbose", is_flag=True, help="Show verbose output with full details.")
-def show(flow_name: str | None, color: bool, verbose: bool):
+def show(app_flow_specifier: str, color: bool, verbose: bool):
     """
-    Show the flow spec and schema in a readable format with colored output.
+    Show the flow spec and schema.
+
+    APP_FLOW_SPECIFIER: Specifies the application and optionally the target flow.
+    Can be one of the following formats:\n
```
badmonster0: These trailing `\n` seems not necessary? It will add an _additional_ blank line.

lemorage: I keep those `\n` for formatting purposes, without these, the help message will output like this:
```
(cocoindex)$  cocoindex show --help
Usage: cocoindex show [OPTIONS] APP_FLOW_SPECIFIER

  Show the flow spec and schema.

  APP_FLOW_SPECIFIER: Specifies the application and optionally the target
  flow. Can be one of the following formats:   - path/to/your_app.py   -
  an_installed.module_name   - path/to/your_app.py:SpecificFlowName   -
  an_installed.module_name:SpecificFlowName

  :SpecificFlowName can be omitted only if the application defines a single
  flow.
```

badmonster0: Thanks for the clarification. Interesting, it looks like click's wrapping behavior.

But I guess by adding a trailing `\n`, there will be an blank line between each bullets, is this the case?

If that's not what we want, we can use `\b` to escape: https://click.palletsprojects.com/en/stable/documentation/#escaping-click-s-wrapping

lemorage: That's cool! Thanks!

python/cocoindex/cli.py
```diff
@@ -1,89 +1,233 @@
...
+         "If not provided, attempts to load '.env' from the current directory.",
+    default=None,
+    show_default=False
+)
+def cli(env_file: str | None):
     """
     CLI for Cocoindex.
     """
+    loaded_env_path = None
+    dotenv_path = env_file or os.path.join(os.getcwd(), ".env")
```
badmonster0: When `env_file` is not provided, I think we may want to use `python-dotenv`'s [`find_dotenv(usecwd=True)`](https://github.com/theskumar/python-dotenv/blob/01f899733de664cda0550207067eb36a1795062f/src/dotenv/main.py#L276-L280) to locate the path. It'll also look into the parent directory.

badmonster0 (approved):
Thanks a lot for making it happen! This is a big thing!

lemorage:
Thanks for all the help along the way! I've created another PR accounting for the docs' changes to the standalone CLI.

badmonster0:
hi @lemorage  thank you so much for all your contribution with cocoindex, we love all your contributions and made a section for you on the latest release note https://cocoindex.io/blogs/cocoindex-changelog-2025-05-31/#lemorage

PR #368, opened 2025-04-22
fix:Add support for TimeDelta type in Python and Rust

Keerthi421:
PR Description:
Issue #344 
This PR implements support for the TimeDelta type in CocoIndex as requested in issue #344.

Changes:
-->Added TimeDelta = Annotated[datetime.timedelta, TypeKind('TimeDelta')] in Python typing module
-->Added TimeDelta variant to Rust's type system
-->Implemented conversion between Python's datetime.timedelta and Rust's chrono::Duration
-->Added PostgreSQL storage support using the native interval type

Testing:
-->Tests should be added to verify:
-->Python type analysis correctly identifies datetime.timedelta objects
-->Conversion between Python and Rust time delta representations works correctly
-->PostgreSQL storage and retrieval works as expected
This addresses #344.

PR #497, opened 2025-05-16
feat: add support for TimeDelta type in Python and Rust

### LINKED ISSUE #344: [FEATURE] Support TimeDelta Type (Python and Rust)
Here are all the dataType supported in CocoIndex: https://cocoindex.io/docs/core/data_types

We want to add TimeDelta Type, following https://github.com/cocoindex-io/cocoindex/issues/217

See 
- [timedelta in Python](https://docs.python.org/3/library/datetime.html#datetime.timedelta)
- [interval in postgres](https://www.postgresql.org/docs/current/datatype.html#:~:text=four%2Dbyte%20integer-,interval,-%5B%20fields%20%5D%20%5B%20%28)
- [TimeDelta in Rust](https://docs.rs/chrono/latest/chrono/struct.TimeDelta.html)

lemorage:
This is a continuation of the unmerged #368.

Resolves #344.

### DIFF
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -35,7 +35,7 @@ This is the list of all basic types supported by CocoIndex:
 | Time | | `datetime.time` | `datetime.time` |
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
-| TimeDelta | A duration of time | `cocoindex.typing.TimeDelta` | `datetime.timedelta` |
+| TimeDelta | A duration of time | `datetime.TimeDelta` | `datetime.timedelta` |
 | Vector[*T*, *Dim*?] | *T* must be basic type. *Dim* is a positive integer and optional. |`cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `list[T]` | 
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
 
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -35,7 +35,7 @@ This is the list of all basic types supported by CocoIndex:
 | Time | | `datetime.time` | `datetime.time` |
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
-| TimeDelta | A duration of time | `datetime.TimeDelta` | `datetime.timedelta` |
+| TimeDelta | A duration of time | `datetime.timedelta` | `datetime.timedelta` |
 | Vector[*T*, *Dim*?] | *T* must be basic type. *Dim* is a positive integer and optional. |`cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `list[T]` | 
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
 
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -35,6 +35,7 @@ This is the list of all basic types supported by CocoIndex:
 | Time | | `datetime.time` | `datetime.time` |
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
+| TimeDelta | A duration of time | `cocoindex.typing.TimeDelta` | `datetime.timedelta` |
 | Vector[*T*, *Dim*?] | *T* must be basic type. *Dim* is a positive integer and optional. |`cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `list[T]` | 
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
 
--- a/python/cocoindex/typing.py
+++ b/python/cocoindex/typing.py
@@ -29,7 +29,6 @@ Range = Annotated[tuple[int, int], TypeKind('Range')]
 Json = Annotated[Any, TypeKind('Json')]
 LocalDateTime = Annotated[datetime.datetime, TypeKind('LocalDateTime')]
 OffsetDateTime = Annotated[datetime.datetime, TypeKind('OffsetDateTime')]
-TimeDelta = Annotated[datetime.timedelta, TypeKind('TimeDelta')]
 
 if TYPE_CHECKING:
     T_co = TypeVar('T_co', covariant=True)
--- a/python/cocoindex/typing.py
+++ b/python/cocoindex/typing.py
@@ -29,6 +29,7 @@ Range = Annotated[tuple[int, int], TypeKind('Range')]
 Json = Annotated[Any, TypeKind('Json')]
 LocalDateTime = Annotated[datetime.datetime, TypeKind('LocalDateTime')]
 OffsetDateTime = Annotated[datetime.datetime, TypeKind('OffsetDateTime')]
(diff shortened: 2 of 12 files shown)


src/base/duration.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    #[test]
+    fn test_iso_valid_days() {
+        check_ok(parse_duration("P1D"), Duration::days(1), "\"P1D\"");
+    }
+
+    #[test]
+    fn test_iso_valid_hours() {
+        check_ok(parse_duration("P2H"), Duration::hours(2), "\"P2H\"");
```
badmonster0: From https://en.wikipedia.org/wiki/ISO_8601#Durations and https://datatracker.ietf.org/doc/html/rfc3339#appendix-A, sees `T` must precede the hours/minutes/seconds, i.e. this should be `PT2H`.

And `P3M` below actually means (3 months). 3 minutes should be `PT3M`.

src/base/duration.rs
```diff
@@ -0,0 +1,455 @@
...
+                        'S' => total += Duration::seconds(n),
+                        c => return Err(anyhow!("Invalid unit in duration: {}", c)),
+                    }
+                    num.clear();
+                    negative = false;
+                }
+            }
+        }
+
+        if total != Duration::zero() {
```
badmonster0: I think zero duration should be valid, e.g. `PT0S` is a valid duration.

docs/docs/core/data_types.mdx
```diff
@@ -35,6 +35,7 @@ This is the list of all basic types supported by CocoIndex:
 | Time | | `datetime.time` | `datetime.time` |
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
+| TimeDelta | A duration of time | `cocoindex.typing.TimeDelta` | `datetime.timedelta` |
```
badmonster0: I think we don't have to introduce `cocoindex.typing.TimeDelta`. Always using `datetime.timedelta` is sufficient (similar to Date and Time), as there's no ambiguity like `float` and `datetime`.

badmonster0 (commented):
Thanks for working on this, especially for careful tests for the Duration parsing logic!

I didn't notice there's no existing completed Duration parser for Rust, so the effort is larger than I original expected. Thanks for the effort!

docs/docs/core/data_types.mdx
```diff
@@ -35,6 +35,7 @@ This is the list of all basic types supported by CocoIndex:
 | Time | | `datetime.time` | `datetime.time` |
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
+| TimeDelta | A duration of time | `datetime.TimeDelta` | `datetime.timedelta` |
```
badmonster0: `datetime.TimeDelta` (all lowercase)

python/cocoindex/typing.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, key: str, value: Any):
 Json = Annotated[Any, TypeKind('Json')]
 LocalDateTime = Annotated[datetime.datetime, TypeKind('LocalDateTime')]
 OffsetDateTime = Annotated[datetime.datetime, TypeKind('OffsetDateTime')]
+TimeDelta = Annotated[datetime.timedelta, TypeKind('TimeDelta')]
```
badmonster0: This can be removed

src/base/duration.rs
```diff
@@ -0,0 +1,747 @@
...
+        }
+        let num = num_str
+            .parse()
+            .map_err(|_| anyhow!("Invalid number '{}' in: {}", num_str, original_input))?;
+        if let Some(&unit) = iter.peek() {
+            if allowed_units.contains(&unit) {
+                result.push((num, unit));
+                iter.next();
+            } else {
+                return Err(anyhow!("Invalid unit '{}' in: {}", unit, original_input));
```
badmonster0: `bail!(...)` is slightly simpler

src/base/duration.rs
```diff
@@ -0,0 +1,747 @@
...
+    let mut seen_time_units = std::collections::HashSet::new();
+    for &(_, unit) in &time_components {
+        if !seen_time_units.insert(unit) {
+            return Err(anyhow!(
+                "Duplicate '{}' in time part: {}",
+                unit,
+                original_input
+            ));
+        }
+    }
```
badmonster0: All duplication problems can always be covered by `check_order` too, so seems we can avoid checking these.

Actually personally I feel it's OK not to enforce ordering. The logic can be even simpler, and it can tolerate more slight problems in LLM output (when we parse JSON we also tolerate certain problems like trailing comma). I know the downside of relaxing this (like the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) has its criticism), but personally feels for this case not enforcing order has more benefit than harm, as it errors on the side of simpler rules (the Occam's razor). If you feel strong here, we can keep this.

src/base/duration.rs
```diff
@@ -0,0 +1,747 @@
...
+
+    if is_likely_iso8601(s) {
+        // Try ISO 8601 format without fallbacks
+        return parse_iso8601_duration(s, original_input);
+    }
+
+    // Allow maybe ISO-ish formats, so still try iso8601 first
+    parse_iso8601_duration(s, original_input)
```
badmonster0: Curious in which case `is_likely_iso8601(s)` returns `false` but `parse_iso8601_duration` may succeed?

src/base/duration.rs
lemorage: Personally, I would be more on the strict side since that can avoid more problems that I don't need to consider. But I agree with your idea of making it more lenient, and we'll see if this can yield more benefits in the long run. Anyway, we can always make it stricter afterwards.

src/base/duration.rs
lemorage: umm right, I was thinking there might be some other corner cases here, but actually, `is_likely_iso8601` can cover them all.

src/base/duration.rs
```diff
@@ -0,0 +1,679 @@
...
+
+/// Parses a duration string into a `chrono::Duration`, trying ISO 8601 first, then human-readable format.
+pub fn parse_duration(s: &str) -> Result<Duration> {
+    let original_input = s;
+    let s = s.trim();
+    if s.is_empty() {
+        return Err(anyhow!("Empty duration string"));
+    }
+
+    fn is_likely_iso8601(s: &str) -> bool {
```
badmonster0: Personally I feel directly using an expression is simpler and slightly better here:

```rs
let is_likely_iso8601 = match s.as_bytes() {
  ...
};
```

src/base/duration.rs
```diff
@@ -0,0 +1,679 @@
...
+/// Parses an ISO 8601 duration string into a `chrono::Duration`.
+fn parse_iso8601_duration(s: &str, original_input: &str) -> Result<Duration> {
+    let (is_negative, s_after_sign) = if s.starts_with('-') {
+        (true, &s[1..])
+    } else {
+        (false, s)
+    };
+
+    if !s_after_sign.starts_with('P') {
+        return Err(anyhow!(
```
badmonster0: We can replace all `Err(anyhow!(...))` with `bail!(...)`

src/ops/storages/postgres.rs
```diff
@@ -217,6 +220,17 @@ fn from_pg_value(row: &PgRow, field_idx: usize, typ: &ValueType) -> Result<Value
                 BasicValueType::OffsetDateTime => row
                     .try_get::<Option<chrono::DateTime<chrono::FixedOffset>>, _>(field_idx)?
                     .map(BasicValue::OffsetDateTime),
+                // BasicValueType::TimeDelta => row
+                //     .try_get::<Option<chrono::Duration>, _>(field_idx)?
+                //     .map(BasicValue::TimeDelta),
```
badmonster0: Please cleanup these comments.

badmonster0 (approved):

badmonster0 (approved):

Issue #436, opened 2025-05-06
[FEATURE] Support union type for basic types


badmonster0:
We want to support union type for [basic types](https://cocoindex.io/docs/core/data_types#basic-types).

## Background

Union types are supported in Python (e.g. `str | int`). CocoIndex doesn't support it now, so when a library already defines a dataclass with Union type members, users cannot use the existing type as argument / return type of a CocoIndex function ([existing example from relik](https://github.com/SapienzaNLP/relik/blob/999baf657a9df095ac138fac61bff944dff3d8ea/relik/inference/data/objects.py#L99)).

### Bind union type in Python to Json type in CocoIndex

However, Json type is too generic and loses essential type information. Unexpected values can be set to the field without being noticed earlier (e.g. when calling LLM to create the dataclass value).

With this said, binding an unsupported type to Json still can be supported, because there're still types that cannot be supported even after we support union type for basic types. This will be treated as a separate feature request.

### Not limit union type to basic type

i.e. we may support union type for struct/table types too.

We may support it eventually, but it's difficult to reliably interpret the type for a specific struct value, especially our struct types are unnamed (for now).

## Things to Change

We want to add `Union` as a new variant for [`BasicValueType`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/src/base/schema.rs#L14), e.g.

```rust
Union(Vec<BasicValueType>),
```

We considered putting it as a new type at the same level as `BasicValueType`, `StructType` and `TableType`. One practical reason of not doing so is: things like `list[str | int]` will not be representable. Considering `Vector` is already part of `BasicValueType`, having `Union` as another `BasicValueType` makes sense.

Other code to update:

- Methods related `BasicValueType`, for serializing, deserializing, etc.
- Value representation doesn't need to change, but when we decode values we're taking target type as argument in [`BasicValue::from_json()`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/src/base/value.rs#L865). It needs to updated.
- [`JsonSchemaBuilder`](https://github.com/cocoindex-io/cocoindex/blob/55d112dc4e13e9e44b681f6f7d1a94eb1ccee33a/src/base/json_schema.rs)
[...]

PR #510, opened 2025-05-18
feat: support union type for basic types

chardoncs:
Here is my solution on the union type support. It includes basic conversions and representations for the union type.

If there are any problems with the change, feel free to speak out for revision or reject.

## Context

This PR is an attempt to solve Issue #436. It includes descriptions and conceptual works about union type support.

> ## Changes

> - Added union in the basic value type, including a new util struct `UnionType`.
>   - [Breaking] Added the union option to the `BasicValueType`.
>   - `UnionType` includes a `BTreeSet` to store the types, with extra features.
>     - Deduplication: E.g., `int | str | int` is `int | str`.
>     - Auto-sorting. The current sorting strategy is using the same order of the definition in `BasicValueType`. It might be > changed to explicit ranks.
>     - Auto-flattening: E.g., `int | (int | str) | str | float` is `int | str | float`
>     - If there is only one type, the parser returns the exact type.
>     - String guessing: If some of UUID, DateTime (`LocalDateTime`, `OffsetDateTime`, `Date`, and `Time`), JSON, and String appear in a union at the same time, the parser will try parsing the value against each possible type in the **reversed** order of the definition (might be changed to explicit ranks).
> - Union type conversion for Qdrant.
> - Union type conversion for PostgreSQL.
> - JSON Schema for union type, using `oneOf`.
> - Format display: `Union[type1 | type2 | ...]`.
> - [Breaking] Added union support for type encoder in Python API.~


src/base/value.rs
badmonster0, line 1: After reading the implementations, I realized I overlooked one thing when creating the original issue: value -> JSON -> value may get a different value after a roundtrip, under union type (the value representation has stronger type information and can uniquely identify the type for basic types, but JSON are weaker typed). 

This can be a problem, meaning a ser-deser roundtrip is no longer transparent (e.g. we cache intermediate computation results for reuse).

Now I'm thinking maybe we can do the following to aid:

- For union typed values, add a number tag to indicate the branch ID, e.g. `UnionVariant(tag_id, value)`, and serialize it as a 2-element array.
- This means the type representation (`UnionType`) needs to keep a `Vec`, so the `tag_id` can more easily work with it. I still like the idea of sorting possible types though - consider doing the sorting when creating the `UnionType`.
- Note that we have another `TypedValue` serialization logic: https://github.com/cocoindex-io/cocoindex/blob/95c7ece9923ddfa69fc2a47a9524d201d982cf9e/src/base/value.rs#L1049-L1057
  The difference is that these serializations are more friendly for being consumed outside cocoindex without cocoindex type info (they are used in  export / dump format now), and not for another deserialization. So for these, we don't need to carry `tag_id` along.

Welcome to any further discussions on this.

python/cocoindex/typing.py
```diff
@@ -69,7 +69,7 @@ class AnalyzedTypeInfo:
     """
     kind: str
     vector_info: VectorInfo | None  # For Vector
-    elem_type: ElementType | None   # For Vector and Table
+    elem_type: ElementType | None   # For Vector, Table, and Union
```
badmonster0: Instead of reusing `elem_type`, I feel we should better have another one like this:

```
union_variant_types: list[type] | None
```

The purpose of this `AnalyzedTypeInfo` is to unpack type info for the current type - so it's ready to be directly consumed in a recursive traversal, and store original python types for its next-level types (e.g. for the type `list[T]` we store `T` instead of `list[T]` in this dataclass). For a union type, the `typing.Union` is the structure for its current level, and its variants is its next level.

python/cocoindex/typing.py
badmonster0, line 1: Note that we also need to update `make_engine_value_decoder()` in `convert.py`.

src/utils/union.rs
```diff
@@ -0,0 +1,158 @@
...
+            let mut type_tree: BTreeSet<BasicValueType> = self.into();
+            return Ok(UnionParseResult::Single(type_tree.pop_first().unwrap()));
+        }
+
+        Ok(UnionParseResult::Union(self))
+    }
+
+    /// Move an iterable and parse it into a union type.
+    /// If there is only one single unique type, it returns a single `BasicValueType`.
+    pub fn parse_from<T>(
```
badmonster0: Looks like both `parse_from()` and `coerce_from()` are only used by tests. Then just wonder what do we lose if we directly construct the `UnionType` in tests?

I can see we may seed something similar for the following purposes:
1. When we get a type encoded from Python code, we may want to canonicalize it, e.g. make sure no duplicated code, no nested union, etc. This is the place we need this: https://github.com/cocoindex-io/cocoindex/blob/95c7ece9923ddfa69fc2a47a9524d201d982cf9e/src/ops/py_factory.rs#L171-L172
2. When building a type in some operations (like [this](https://github.com/cocoindex-io/cocoindex/blob/95c7ece9923ddfa69fc2a47a9524d201d982cf9e/src/ops/functions/split_recursively.rs#L599)) some helpers may also be needed.

For 1, in my mind a function like this may be ideal:
```rs
fn canonicalize_type(t: EnrichedValueType): EnrichedValueType 
```

For 2, what we need may be closer to this `parse_from`, while in my mind the util may directly return a `BasicValueType` for ease of use.

(since we don't have a op built in Rust needs this now, we don't need to worry about 2 for now)

badmonster0 (commented):
Thanks for the PR! Really appreciate the effort!

Please see the bunch of comments. Welcome for any further discussion!

chardoncs:
Thanks! I'll see what I can do

src/base/value.rs
chardoncs: Yes, I understand that transparency can be a problem. My current implementation includes type guessing that may cause inconsistency.

So, my understanding is that the `tag_id` is for the current type of the value, and it's serialized altogether with the value as a JSON array. Is that correct?

badmonster0: Yes, that's correct.
(sorry missed this question yesterday)

badmonster0: Thanks for updating the code! Looked at the latest in-progress implementation. This is still a little bit different from what I thought.

`UnionVariant(tag_id, value)` I suggested is a branch of `BasicValue`. Writing it out, may be something like this:

```rs
enum BasicValue {
  ...
  UnionVariant{tag_id: usize, value: Box<BasicValue>},
}
```

e.g. if a union type is `Str | Int64`, `Str` has tag ID 0, `Int64` has tag ID 1.

Both serialization and deserialization will be more straightforward:
- To serialize: directly convert it into array `[tag_id, serialize(value)]`
- To deserialize (`from_json()` method): with the tag_id, we can directly get the specific type from the vector in `UnionType`, then use it to deserialize value.

This will be simpler and more robust: we don't need to guess the `BasicType` from the `BasicValue`, which may not be always possible (e.g. zero-length vectors, or a type like `Str | Vector[Str | Int64]`).

---
Besides, I realized there're one more thing may need discuss - I didn't realize last time.

For deserialization, there're two types of use cases:
1. Deserialize what we serialized. It will guarantee we get the same thing in a roundtrip.
2. Deserialize JSON values created by some external systems. Currently the only situation is `ValueExtractor` in `json_schema.rs` (e.g. values are created by LLM for the case), which we don't expect they put the tag ID. On the other hand, we don't expect the value type preserved in a roundtrip for this case too. So we can guess by trying each possible types during deserialization.

To distinguish these two, we can add a new input to `from_json()` method, to indicate which mode, e.g. we can have another enum like:

```rs
enum DeserializationMode {
  /// Deserialize what we serialized. Guaranteed to preserve original type.
  Internal,
  /// Deserialize JSON values produced by external systems. 
  External,
}
```

---
Thanks for your patience! Let me know if you have any other thoughts. Thanks!

chardoncs: Thanks for addressing the misunderstanding. I was actually concerned by serializing the element type.

It is a simple way to eliminate ambiguity.

chardoncs: By the way, I was initially thinking about serialize and deserialize `BasicValue` without a `UnionVariant` branch.

The serialization creates the `tag_id` when `BasicValueType` is `Union`:

> E.g. `BasicValue::Int64(10)` with type `Str | Int` -> `{ tag_id: 1, value: 10 }`

And deser uses the `BasicValueType` and `tag_id` to construct `BasicValue`:

> E.g. `{ tag_id: 1, value: 10 }` with type `Str | Int` -> `BasicValue::Int64(10)`

badmonster0: Nice thought! This may have the following difficulties though:

- During serialization we don't have the `BasicValueType`. The serialization is designed in a way that can work standlone with the `Value` (otherwise we need to do it by a serialization method not standard for `serde`, and a lot things need to be changed, e.g. other serializable structs using `Value` as fields).
- Even if we pass the type down, we still need to guess the type based on the value. It's doable for most cases, but there're still some difficult cases. e.g. when there's a vector,  we need to look into the element, which can be union type too. We'll introduce more times in the future (e.g. `Enum`, #523), which may add more complexity (e.g. how about `Vector[Str] | Vector[Enum1]`)?

I prefer avoiding these complexities, to avoid hitting a case that we cannot handle some day.

src/utils/union.rs
chardoncs: I see. I have removed all those methods so `UnionType` should be created only through pyo3.

python/cocoindex/typing.py
chardoncs: Actually, I'm currently confused by this decoding. It seems like there is no way to encode a union engine value from a Python value. The Python union type cannot be detected from `encode_engine_value()`. (Maybe a custom class is needed)

Also, for decoding. Currently, my implementation for the `UnionVariant` branch in the Rust function `basic_value_to_py_object<'py>()` is like this:

```rust
let result = match v {
// snip
        value::BasicValue::UnionVariant { value, .. } => {
            basic_value_to_py_object(py, &value)?
        }
};
```

In this case, I suspect the union type cannot be detected from the make decoder function in any way, neither through `encode_engine_value()` nor `TransientFlow::evaluate_async()`.

But if we return a struct that is converted into the Python value (such as something like `{ type: "Str", value: "foo" }`), then the make decoder function can detect it.

badmonster0: In Python, I think there's no common way to get the specific branch for a value annotated with Union type, e.g. for `list[int] | list[float]`, type information only exists in each specific list element. There's no type information to directly distinguish `list[int]` and `list[float]`.

So, for encoding (Python->Rust), probably we have to try different branches and see which will succeed, on Rust side. Your current implementation looks good!

For decoding (Rust->Python), since both Rust and Python side have consistent type information, I think we can follow similar approach as how we serialize to JSON. On Rust side, we may convert it into a tuple of `(tag_id, value)`, then on Python side using `tag_id` we can find out the type of the specific branch. What do you think about this? Thanks!

chardoncs: Yes, that should do the trick.

chardoncs: On a second thought, there is a problem with the B-Tree set. It seems like we have to use `Vec` to preserve the order if `tag_id` is returned.

badmonster0: Yes, please use `Vec` - this is compatible with the `tag_id` approach.

chardoncs:
It seems like there is a new DB support (KuzuDB), and it natively supports union.

Update: I just checked the unions in KuzuDB. It is C/C++-like unions, which means every type is named (e.g., `UNION(a STRING, b INT64)`). So I think the names should be automatically generated like `val1`, `val2` (or `val0`, `val1`, ...), or something like that.

src/base/value.rs
```diff
@@ -954,6 +967,26 @@ impl BasicValue {
...
+
+                let value = obj.get(1)
+                    .ok_or_else(|| anyhow::anyhow!("`value` is not available in the value"))?;
+
+                let cur_type = typ.types().get(tag_id)
+                    .ok_or_else(|| anyhow::anyhow!("No type in `tag_id` \"{tag_id}\" found"))?;
+
+                BasicValue::UnionVariant {
+                    tag_id,
+                    value: Box::new(BasicValue::from_json(value.clone(), cur_type)?),
```
badmonster0: This clone can be avoided  (it can be a large string), as we have ownership of `v` and can move from it.

src/ops/storages/neo4j.rs
```diff
@@ -226,6 +226,19 @@ fn basic_value_to_bolt(value: &BasicValue, schema: &BasicValueType) -> Result<Bo
             _ => anyhow::bail!("Non-vector type got vector value: {}", schema),
         },
         BasicValue::Json(v) => json_value_to_bolt_value(v)?,
+        BasicValue::UnionVariant { tag_id: tag_id_ref, value } => {
```
badmonster0: nit: introducing a variable name for reference is usually unnecessary. This is simpler:

```rs
BasicValue::UnionVariant { tag_id, value } => {
  ...
  let type = s.types().get(*tag_id)
```

src/ops/storages/postgres.rs
```diff
@@ -159,6 +159,9 @@ fn bind_value_field<'arg>(
                     builder.push_bind(sqlx::types::Json(v));
                 }
             },
+            BasicValue::UnionVariant { tag_id, value } => {
+                builder.push_bind(sqlx::types::Json(serde_json::json!([tag_id, value])));
```
badmonster0: For the target storage, the value is consumed by an external system and doesn't need to be read back, so we only to export the value without `tag_id`. 

We can serialize a `TypedValue` to JSON (similar to the last branch a few lines below) .

I know it'll introduce difficulty for the implementation of `from_pg_value()` below - but actually that one will be deprecated soon, so we don't need to support union type for it (background: https://github.com/cocoindex-io/cocoindex/issues/504; currently all examples are not using it) - we can raise an error when encountering union type there.

src/ops/storages/qdrant.rs
```diff
@@ -139,6 +138,9 @@ fn values_to_payload(
                         vectors = vectors.add_vector(field_name, vector);
                         continue;
                     }
+                    BasicValue::UnionVariant { tag_id, value } => {
+                        json!([tag_id, value])
```
badmonster0: Similar here, we only need to output the `value` here, without `tag_id`.

src/py/convert.rs
```diff
@@ -156,6 +160,22 @@ fn basic_value_from_py_object<'py>(
...
+                if let Ok(value) = basic_value_from_py_object(typ, v) {
+                    valid_value = Some(value::BasicValue::UnionVariant {
+                        tag_id: i,
+                        value: Box::new(value),
+                    });
+                    break;
+                }
+            }
+
+            valid_value.ok_or_else(|| PyErr::new::<PyTypeError, _>("invalid union"))?
```
badmonster0: we may put the type name in the error message, to make it easier for users to debug

src/utils/union.rs
```diff
@@ -0,0 +1,19 @@
+use crate::{base::schema::BasicValueType, prelude::*};
+
+/// Union engine type helper
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct UnionType {
+    types: Vec<BasicValueType>,
```
badmonster0: Seems a separate mod is unnecessary. Probably just put this in `schema.rs`?

src/utils/union.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+impl UnionType {
+    pub fn types(&self) -> &[BasicValueType] {
+        self.types.as_slice()
+    }
+}
+
+impl Into<Vec<BasicValueType>> for UnionType {
+    fn into(self) -> Vec<BasicValueType> {
```
badmonster0: Is this used anywhere?

src/utils/union.rs
```diff
@@ -0,0 +1,19 @@
+use crate::{base::schema::BasicValueType, prelude::*};
+
+/// Union engine type helper
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct UnionType {
+    types: Vec<BasicValueType>,
+}
+
+impl UnionType {
+    pub fn types(&self) -> &[BasicValueType] {
```
badmonster0: Personally I feel this doesn't provide much value. We can just make the `types` field public instead.

badmonster0 (commented):
Thanks for your patience! Very close!

src/utils/union.rs
chardoncs: Sure, initially I thought there would be many methods for processing union types. Since it has been slimmed down, it is fine to put it in `schema.rs`.

I would like to rename it `UnionTypeSchema` after `VectorTypeSchema` in `schema.rs`.

src/ops/storages/qdrant.rs
chardoncs: Is the `value` inside the `UnionVariant` serialized into JSON for this, while it is `TypedValue` for the last one (in `postgres.rs`)?

src/base/value.rs
```diff
@@ -954,6 +967,35 @@ impl BasicValue {
...
+                // Take first element
+                let tag_id = obj_iter
+                    .next()
+                    .and_then(|value| value.as_u64().map(|num_u64| num_u64 as usize))
+                    .ok_or_else(|| anyhow::anyhow!("`tag_id` is not available in the value"))?;
+
+                // Take second element
+                let value = obj_iter
+                    .next()
+                    .ok_or_else(|| anyhow::anyhow!("`value` is not available in the value"))?;
```
badmonster0: Since we already checked `len()` is 2 above, this is an invariance now. We can do this to keep code simpler:

```rs
let value = obj_iter.next().ok_or_else(invariance_violation)?;
```

python/cocoindex/typing.py
badmonster0, line 1: `TypedValue` in this file also needs to be updated. for `UnionType`, we should only serialize the value part, without the tag.

src/base/value.rs
badmonster0: Note that `TypedValue` in this file also needs to be updated. for `UnionType`, we should only serialize the value part, without the tag.

python/cocoindex/typing.py
```diff
@@ -182,7 +184,23 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
     elif base_type is collections.abc.Mapping or base_type is dict:
         args = typing.get_args(t)
         elem_type = (args[0], args[1])
-        kind = "KTable"
+        kind = 'KTable'
+    elif base_type is types.UnionType:
+        possible_types = typing.get_args(t)
```
badmonster0, line 245: Line 128-140 above, there's already handling for `UnionType`, and will error out when there are more than one non-none branches, so I think it won't reach here. I think one way is to directly change the code in L128-140 above for our UnionType support.

src/ops/storages/qdrant.rs
badmonster0: We should use `TypedValue` for this too. Since a value in qdrant is just a JSON without specific types, actually we can just use `TypedValue` for all types (except vectors, which is special in qdrant). The existing code to handle this can be simplified. I just created https://github.com/cocoindex-io/cocoindex/pull/587 to use `TypedValue` for most cases. Then this file no longer needs additional updates for union type support.

chardoncs: Yeah, I just merged the changes from the upstream.

src/base/value.rs
chardoncs: I see. What about `.unwrap()` or `.expect()`?

python/cocoindex/typing.py
chardoncs: Oh, initially I removed the check in L128-140. It seems like the change was overwritten when I was merging.

src/base/value.rs
badmonster0: `.unwrap()` looks good as the validation on length is very close. Thanks!

src/ops/storages/postgres.rs
chardoncs: Sorry for my late response.

I have done some changes days ago. I just checked my changes, and it seems will do the trick.

chardoncs:
I just spared some time to check the changes. Looks like everything has been addressed for now. But it needs thorough testing.

badmonster0:
> I just spared some time to check the changes. Looks like everything has been addressed for now. But it needs thorough testing.

That's great, thanks!

I just added a testutil `validate_full_roundtrip()`:

https://github.com/cocoindex-io/cocoindex/blob/48a03310d47c8dab9e3de03fe8009d8dfc8d286e/python/cocoindex/tests/test_convert.py#L67-L69

It can be used to in test to cover multiple encode/decode/serialize/deserialize functions on both Rust and Python side. Would you try to add a few tests for union types with this and see if it passes?

chardoncs:
Sorry for the late reply. I was sick these days and couldn't think or code.

I will check the change and add test cases, and try not to delay this PR any further.

badmonster0:
> Sorry for the late reply. I was sick these days and couldn't think or code.
> 
> I will check the change and add test cases, and try not to delay this PR any further.

@chardoncs  I'm sorry to hear you're sick. No worries, take your time and have enough rest! Hope you'll be better soon.

chardoncs:
I just added some test cases and fixed some bugs.

There are some problems with `datetime` conversions.

python/cocoindex/tests/test_convert.py
```diff
@@ -477,6 +477,48 @@ def test_field_position_cases(
...
+def test_roundtrip_union_with_inactive_uuid() -> None:
+    t = str | uuid.UUID | int
+    value = "5a9f8f6a-318f-4f1f-929d-566d7444a62d" # it's a string
+    validate_full_roundtrip(value, t)
+
+
+#def test_roundtrip_union_datetime() -> None:
+#    t = str | uuid.UUID | float | int | datetime.datetime
+#    value = datetime.datetime.now()
+#    validate_full_roundtrip(value, t)
```
chardoncs: There are some problems with this test case. It seems like `datetime` is converted as `date`.

badmonster0: There was a for `LocalDateTime` and fixed in https://github.com/cocoindex-io/cocoindex/pull/624

Also note that `datetime` is equivalent to `OffsetDateTime` and a timezone is needed for now. `LocalDateTime` doesn't need timezone. Added roundtrip test for these two types in https://github.com/cocoindex-io/cocoindex/pull/624

chardoncs:
I have updated the test cases, and it works now.

Now I'm trying to figure out if this implementation works with databases (i.e., Qdrant, Postgres, Neo4j, and Kuzu). My concern right now is especially on Kuzu, because it has a native C-like union type, which functions very different from the others.

badmonster0:
> My concern right now is especially on Kuzu, because it has a native C-like union type, which functions very different from the others.

That's a fair concern. I had similar concern, e.g. users need to access using tags like `val0`, `val1` etc. which is a little bit arbitrary. I think probably we want to store it as JSON in Kuzu, similar to Postgres. But JSON requires an extension on Kuzu (https://docs.kuzudb.com/extensions/). I think probably you can skip supporting Kuzu for this PR (similar to the Json type, which is also unsupported for Kuzu yet). Later I'll add JSON extension support for Kuzu which will make us easily support union type for it too.

chardoncs:
I just checked current implementation.

I have removed union implementation for Kuzu. And it is JSON for Postgres and Qdrant, but I have introduced `tag_id` back for JSON conversion, because otherwise the full roundtrip tests would fail. However, there is no `tag_id` for Neo4j.

badmonster0:
> I just checked current implementation.
> 
> I have removed union implementation for Kuzu. And it is JSON for Postgres and Qdrant, but I have introduced `tag_id` back for JSON conversion, because otherwise the full roundtrip tests would fail. However, there is no `tag_id` for Neo4j.

Yes, we should have `tag_id` for the the normal serialization.

For Postgres, we're using `TypedValue` to serialize it. As I explained in another comment thread (you can search for "TypedValue", the thread has more context about the specific consideration/context for `TypedValue`), serialization for `TypedValue` needs to be updated to drop the tag_id for union types. Hence Postgres won't have `tag_id` in its JSON.

src/base/value.rs
```diff
@@ -1128,7 +1128,17 @@ impl Serialize for TypedValue<'_> {
...
+                match t {
+                    BasicValueType::Union(_) => match v {
+                        Value::Basic(BasicValue::UnionVariant { value, .. }) => {
+                            value.serialize(serializer)
+                        }
+                        _ => Err(serde::ser::Error::custom("Unmatched union type and value for `TypedValue`")),
+                    },
+                    _ => v.serialize(serializer),
+                }
+            }
```
chardoncs: I have updated the `TypedValue` serialization.

In this case, if `TypedValue` is serialized, the `tag_id` will get dropped and the inner `value` will be serialized. And the serialization for `BasicValue` and `Value` is not affected.

src/base/value.rs
```diff
@@ -956,6 +969,33 @@ impl BasicValue {
                     .collect::<Result<Vec<_>>>()?;
                 BasicValue::Vector(Arc::from(vec))
             }
+            (v, BasicValueType::Union(typ)) => {
+                let obj: Vec<serde_json::Value> = serde_json::from_value(v)
```
badmonster0: Something like this is more concise and efficient:

```rs
let arr = match v {
  serde_json::Value::Array(arr) => arr,
  _ => bail!("Invalid JSON value for union, expect array"),
};
```

badmonster0 (commented):
Please also run `./check.sh`, which will run autoformat and will resolve github workflow check errors.

badmonster0:
Thanks a lot for pushing through this feature, and all revisions along the way!

badmonster0:
Hi @chardoncs Charles, thank you so much for your contribution. We made a section for you in the latest release note

https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#chardoncs

we love your contribution!

PR #518, opened 2025-05-20
update text-embedding example to use latest query handler

badmonster0:


examples/text_embedding/README.md
```diff
@@ -1,7 +1,17 @@
...
+
+In this example, we will build a text embedding index and a semantic search flow based on local markdown files.
+
+
+<img width="461" alt="Screenshot 2025-05-19 at 5 48 28 PM" src="https://github.com/user-attachments/assets/b6825302-a0c7-4b86-9a2d-52da8286b4bd" />
+
+- We will ingest from a list of local files.
+- For each file, perform chunking (Recursive Split) and then embeddings. 
+- We will save the embeddings and the metadata in Postgres with PGVector.
+- And then add a simpler query handler for semantic search.
```
badmonster0: There's no more "query handler". Maybe just something like "We will match against user-provided text by a SQL query, reusing the embedding operation in the indexing flow."

badmonster0: https://github.com/cocoindex-io/cocoindex/blob/jelly-ex2/examples/text_embedding/README.md ptal

examples/text_embedding/README.md
```diff
@@ -1,6 +1,24 @@
...
+
+### Indexing Flow:
+<img width="461" alt="Screenshot 2025-05-19 at 5 48 28 PM" src="https://github.com/user-attachments/assets/b6825302-a0c7-4b86-9a2d-52da8286b4bd" />
+
+1. We will ingest from a list of local files.
+2. For each file, perform chunking (Recursive Split) and then embeddings. 
+3. We will save the embeddings and the metadata in Postgres with PGVector.
+   
+### Query:
+1. We will match against user-provided text by a SQL query, reusing the embedding operation in the indexing flow.
```
badmonster0: There's no `2.`, probably no longer needs `1.`?

PR #534, opened 2025-05-21
upgrade query for image search

badmonster0:


examples/image_search/README.md
```diff
@@ -35,13 +33,6 @@ ollama pull gemma3
 ollama serve
 ```
 
-## Create virtual environment and install dependencies
-```
-python -m venv .venv
-source .venv/bin/activate
-pip install -r requirements.txt
```
badmonster0, line 42: We need a `pip install -e .` to install dependencies

PR #535, opened 2025-05-21
Update jupyter notebook code in text embedding example

lemorage:
Relates to #515, #518.

The query handler logic has been changed since those PRs, but it seems the code part in the Jupyter Notebook has never been changed (#530 updates the CLI part, but not query logic).

P.S. It appears that my editor rearranges the metadata a lot when I modify the notebook code.


badmonster0 (approved):
Thanks a lot for updating this!

PR #554, opened 2025-05-25
 chore(mypy): make python codebase strictly typed

### LINKED ISSUE #539: [FEATURE] Enable strict mode for `mypy`
Currently we don't set `strict = True` for `mypy` because we're still missing type annotations. We want to enable it (in `pyproject.toml`) after missing type annotations are added.

Note: `tests/` can be excluded from strict mode.

---

lemorage:
Closes #539.

### DIFF
--- a/python/cocoindex/convert.py
+++ b/python/cocoindex/convert.py
@@ -8,7 +8,7 @@ import inspect
 import uuid
 
 from enum import Enum
-from typing import Any, Callable, get_origin, Mapping, Type
+from typing import Any, Callable, get_origin, Mapping
 from .typing import (
     analyze_type_info,
     encode_enriched_type,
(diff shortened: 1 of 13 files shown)


python/cocoindex/convert.py
```diff
@@ -41,7 +41,7 @@ def encode_engine_value(value: Any) -> Any:
 def make_engine_value_decoder(
     field_path: list[str],
     src_type: dict[str, Any],
-    dst_annotation,
+    dst_annotation: Type[Any] | None,
```
badmonster0: `typing.Type` seems deprecated, in favor of `type`:

https://docs.python.org/3/library/typing.html#typing.Type

python/cocoindex/convert.py
```diff
@@ -60,7 +60,7 @@ def make_engine_value_decoder(
     if (
         dst_annotation is None
         or dst_annotation is inspect.Parameter.empty
-        or dst_annotation is Any
+        or get_origin(dst_annotation) is Any
```
badmonster0: This update will change the behavior. If the annotation is `Any`, `get_origin(dst_annotation)` will be `None`

python/cocoindex/convert.py
lemorage: Thanks!

badmonster0 (approved):

Issue #400, opened 2025-04-27
Allow `LlmGenerationClient::geneate()` directly return a JSON value


badmonster0:
We may change the trait to allow [`LlmGenerationClient::geneate()`](https://github.com/cocoindex-io/cocoindex/blob/1218420a02b01886e614b340bd450eb7dce908d6/src/llm/mod.rs#L46-L49) directly return a JSON. The return type can be an enum, with either a JSON or a String, depending on if JSON schema is provided in request.

_Originally posted by @badmonster0 in https://github.com/cocoindex-io/cocoindex/pull/395#discussion_r2061032086_

par4m:
I can work on this, please assign!

badmonster0:
hi @par4m thanks a lot for taking a look at this ticket, we freed up this ticket for now since it has been inactive for a month, if you are actively working on it, please let us know and we will reassign to you! 
really appreciate your contribution!

aryasoni98:
I’m happy to work on this! Could you please assign the issue to me?

georgeh0:
> I’m happy to work on this! Could you please assign the issue to me?

@aryasoni98 thanks for requesting. I already assigned https://github.com/cocoindex-io/cocoindex/issues/994 to you as requested. We encourage contributors taking one issue at a time especially at the beginning. We'll assign more issues after the existing one is done. Thanks!

prvcds:
Hey! can i work on this issue?

Dhruvm23:
Hi @georgeh0 ,Can i take this issue?

Lzzzzzt:
Hello @georgeh0, I'd like to working on this issue, could you assign this to me?

thisisharsh7:
I want to work on this issue

badmonster0:
@thisisharsh7 assigned thanks!

@Lzzzzzt looks like your prev effort is inactive. thanks a lot for your contribution!

PR #570, opened 2025-05-28
Feat: LLM Generation Client Returns Json

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/400

Note: Still testing OpenAI and Anthropic clients, Gemini and Ollama work as expected. 

##  Return Type: `LlmGenerationResponse`
All client files (`anthropic.rs, gemini.rs, ollama.rs, openai.rs`) implement the `LlmGenerationClient` trait and their `generate() `methods return `LlmGenerationResponse`, which supports both `Json(serde_json::Value)` and `Text(String)`.
This is consistent with the enum definition in `mod.rs`.

### Anthropic:
- Checks if `output_format` is `JsonSchema` and prepends `STRICT_JSON_PROMPT` to the system prompt.
- Uses `payload["system"] = ...` for the system prompt.
- `json5` fallback removed as no longer required.
### Gemini:
- Prepends
- Uses `payload["systemInstruction"] = ...` for the system prompt.
### Ollama:
- Prepends
- Uses `system: Some(system_prompt.as_str())` in the request.
### OpenAI:
- Does not prepend `STRICT_JSON_PROMPT`. Instead, uses OpenAI’s native strict JSON mode via `ResponseFormat::JsonSchema { ... strict: Some(true), ... }`.
- This matches the intended design: strict prompt only for Anthropic, Gemini, and Ollama.

## Prompt Utility (`prompt_utils.rs`)
```rust
pub const STRICT_JSON_PROMPT: &str = "IMPORTANT: Output ONLY valid JSON that matches the schema. Do NOT say anything else. Do NOT explain. Do NOT preface. Do NOT add comments. If you cannot answer, output an empty JSON object: {}.";
```
- This is imported and used where needed (Anthropic, Gemini, Ollama).

### Follow Up: Requires Unit Tests to test all LLM Clients at once at once after its merged.
- Request Response Format testing using `httpmock` crate
- End to End testing with Schema Validation using `jsonschema`


src/llm/ollama.rs
```diff
@@ -71,9 +76,15 @@ impl LlmGenerationClient for Client {
...
         let body = res.text().await?;
         let json: OllamaResponse = serde_json::from_str(&body)?;
-        Ok(super::LlmGenerateResponse {
-            text: json.response,
-        })
+        // Check if output_format is JsonSchema, try to parse as JSON
+        if let Some(super::OutputFormat::JsonSchema { .. }) = request.output_format {
+            match serde_json::from_str::<serde_json::Value>(&json.response) {
+                Ok(val) => Ok(super::LlmGenerateResponse::Json(val)),
+                Err(_) => Ok(super::LlmGenerateResponse::Text(json.response)),
```
badmonster0, line 83: I think we want to just pop up the error in this case. When the caller wants JSON, returning a text will leave more burden on the caller side: more cases need to be handled, and when they perform exactly the same way to parse (using `serde_json::from_str()`) they will fail again.

src/llm/openai.rs
```diff
@@ -64,8 +64,10 @@ impl LlmGenerationClient for Client {
             },
         ));
 
+        // Save output_format before it is moved.
+        let output_format = request.output_format.clone();
```
badmonster0, line 68: I think we only need to save a boolean here to indicate if JSON is expected. No need to clone the entire thing.

par4m:
Closing this as out of date and would require rebasing. Will reopen with all the LLM APIs converted.

PR #572, opened 2025-05-28
chore: use theme aligning with user system preferences

lemorage:
A small change on top of our existing docs site.

According to https://docusaurus.io/docs/api/themes/configuration, the current behavior defaults it to "light" color mode and allows users to toggle theme mode as needed. This PR changes it to check user's system preference first.


badmonster0:
Good catch! Thanks!

Issue #576, opened 2025-05-31
Deprecate `ResourceSetupStatus::as_any()`


badmonster0:
`ResourceSetupStatus` has an `as_any()` method, mainly as a workaround for Rust not supporting dyn upcasting.

https://github.com/cocoindex-io/cocoindex/blob/b2d6de588b89067f90027637d29a625f3131c442/src/setup/states.rs#L226-L229

The support in Rust has been stabilized in Rust v1.86, so we can remove it and use dyn upcasting to convert it to `Any` when needed.

samuraikillers:
Hi @badmonster0 , will work on the issue

badmonster0:
Unassignning this issue as no further progress on https://github.com/cocoindex-io/cocoindex/pull/579 after a week.

vumichien:
@badmonster0 Can I work on this issue?

badmonster0:
@vumichien it's yours, thanks a lot for taking a look! welcome to the community! we are very active on https://discord.com/invite/zpA9S2DR7s for any questions and support. We also closely watch all the issues and questions on the github issues. if there's anything we can help please let us know anytime.

PR #579, opened 2025-05-31
removing as_any func

samuraikillers:
This PR resolve issue #576

### DIFF
--- a/src/setup/states.rs
+++ b/src/setup/states.rs
@@ -18,6 +18,7 @@ use owo_colors::{AnsiColors, OwoColorize};
 use std::fmt::Debug;
 use std::fmt::{Display, Write};
 use std::hash::Hash;
+use std::any::Any;
 
 use super::db_metadata;
 use crate::execution::db_tracking_setup::{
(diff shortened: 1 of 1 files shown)


badmonster0:
Thanks for taking this! This PR fails to build - there're existing usages of `as_any()` not converted yet.

samuraikillers:
Hi @badmonster0 ,
So basically the change I made for the `ResourceSetupStatus` function changing static to Any, the same has to be done for others right?

badmonster0:
> Hi @badmonster0 , So basically the change I made for the `ResourceSetupStatus` function changing static to Any, the same has to be done for others right?

@samuraikillers 
You removed `as_any()` but code using `as_any()` is still there, so the compile doesn't pass. Please see the check failures for more details: https://github.com/cocoindex-io/cocoindex/actions/runs/15361469976/job/43236701928?pr=579

You'll need to update the code using `as_any()`.

samuraikillers:
@badmonster0 that's exactly my question. The current push I did I have removed `as_any` inside the functions and added it in function type.

So I assume the same has to be done for all the code where we are observing this error?

badmonster0:
> @badmonster0 that's exactly my question. The current push I did I have removed `as_any` inside the functions and added it in function type.
> 
> So I assume the same has to be done for all the code where we are observing this error?

@samuraikillers Of course, when a trait method is removed, all implementations for this trait method need to be cleaned up, and call sites for this method need to be migrated (for this specific case, using the new [trait upcasting](https://blog.rust-lang.org/2025/04/03/Rust-1.86.0/#trait-upcasting) feature added in Rust 1.86)

badmonster0:
Closing this as this PR doesn't have much progress and no response for a week, and free up the original issue. Feel free open another PR if you addressed it.

vumichien:
@badmonster0 Can I work on this issue?

PR #586, opened 2025-06-02
feat: add NumPy array support for vector representations

### LINKED ISSUE #525: [FEATURE] Support Vector of numbers binding to `numpy.NDArray`
Currently we're using `list` for vectors in Python. It's not efficient and needs extra conversions, e.g. when building vector search queries with pgvector like the `::vector` here: https://github.com/cocoindex-io/cocoindex/blob/a37eddcfc3309be3e82d2fd598e0c40dfa9e1467/examples/text_embedding/main.py#L55

We want to support vector of numbers represented by `numpy.typing.NDArray`, e.g. `numpy.typing.NDArray[npmpy.float32]`, `numpy.typing.NDArray[npmpy.float64]`, `numpy.typing.NDArray[npmpy.int64]` for corresponding types.

#437 is a similar issue - which also added a new Python representation for an existing type.

---

lemorage:
Three main contents have been done in this PR:
1. We added the NDArray support as vector representations.
2. We added a few test suites for verifying this.
3. We updated the code examples and documentation accordingly.

Closed #525.

### DIFF
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -36,8 +36,8 @@ This is the list of all basic types supported by CocoIndex:
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
 | TimeDelta | A duration of time | `datetime.timedelta` | `datetime.timedelta` |
-| Vector[*T*, *Dim*?] | *T* must be basic type. *Dim* is a positive integer and optional. |`cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `list[T]` | 
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
+| Vector[*T*, *Dim*?] | A vector of numbers with type *T* (`np.float32`, `np.float64`, `np.int32`, `np.int64`, `np.uint8`, `np.uint16`, `np.uint32`, or `np.uint64`). *Dim* is a positive integer and optional. | `cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `numpy.typing.NDArray[T]` or `list[T]` |
 
 Values of all data types can be represented by values in Python's native types (as described under the Native Python Type column).
 However, the underlying execution engine and some storage system (like Postgres) has finer distinctions for some types, specifically:
--- a/docs/docs/core/data_types.mdx
+++ b/docs/docs/core/data_types.mdx
@@ -37,15 +37,16 @@ This is the list of all basic types supported by CocoIndex:
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
 | TimeDelta | A duration of time | `datetime.timedelta` | `datetime.timedelta` |
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
-| Vector[*T*, *Dim*?] | A vector of numbers with type *T* (`np.float32`, `np.float64`, `np.int32`, `np.int64`, `np.uint8`, `np.uint16`, `np.uint32`, or `np.uint64`). *Dim* is a positive integer and optional. | `cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `numpy.typing.NDArray[T]` or `list[T]` |
+| Vector[*T*, *Dim*?] | *T* can be a basic type or a numeric type. *Dim* is a positive integer and optional. | `cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `numpy.typing.NDArray[T]` or `list[T]` |
 
 Values of all data types can be represented by values in Python's native types (as described under the Native Python Type column).
 However, the underlying execution engine and some storage system (like Postgres) has finer distinctions for some types, specifically:
 
 *   *Float32* and *Float64* for `float`, with different precision.
 *   *LocalDateTime* and *OffsetDateTime* for `datetime.datetime`, with different timezone awareness.
-*   *Vector* has optional dimension information.
 *   *Range* and *Json* provide a clear tag for the type, to clearly distinguish the type in CocoIndex.
+*   *Vector* holds elements of type *T*. If *T* is numeric (e.g., `np.float32` or `np.float64`), it's represented as `NDArray[T]`; otherwise, as `list[T]`.
+*   *Vector* also has optional dimension information.
 
 The native Python type is always more permissive and can represent a superset of possible values.
 *   Only when you annotate the return type of a custom function, you should use the specific type,
--- a/examples/text_embedding/main.py
+++ b/examples/text_embedding/main.py
@@ -1,5 +1,6 @@
 from dotenv import load_dotenv
 from psycopg_pool import ConnectionPool
+from pgvector.psycopg import register_vector
 import cocoindex
 import os
 from numpy.typing import NDArray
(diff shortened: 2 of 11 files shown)


python/cocoindex/typing.py
```diff
@@ -50,33 +51,38 @@ def __init__(self, key: str, value: Any):
...
+                # Element type and dimension provided, e.g., Vector[np.float32, Literal[3]]
+                numpy_dtype, dim_literal = params
+                if numpy_dtype not in DtypeRegistry.supported_dtypes():
+                    raise TypeError(
+                        f"Vector dtype must be a NumPy dtype, got {numpy_dtype}"
+                    )
+                if typing.get_origin(dim_literal) is Literal:
+                    dim_val = typing.get_args(dim_literal)[
+                        0
+                    ]  # Extract the literal value
```
badmonster0: nit: these line breaks look a little bit weird. Maybe put the comment before the statement.

examples/text_embedding/main.py
```diff
@@ -68,7 +70,7 @@ def search(pool: ConnectionPool, query: str, top_k: int = 5):
         text_embedding_flow, "doc_embeddings"
     )
     # Evaluate the transform flow defined above with the input query, to get the embedding.
-    query_vector = text_to_embedding.eval(query)
+    query_vector = text_to_embedding.eval(query).tolist()
```
badmonster0: I think pgvector API natively support numpy array. I think we don't  need the `tolist()` here. And we can remove the `::vector` in the query below (the purpose of it was to convert a list to a vector).

python/cocoindex/typing.py
```diff
@@ -89,6 +95,51 @@ def _is_struct_type(t: ElementType | None) -> bool:
...
+    @staticmethod
+    def get_by_kind(kind: str) -> DtypeInfo:
+        for info in DtypeRegistry._mappings.values():
+            if info.kind == kind:
+                return info
+        raise ValueError(f"Unsupported type kind: {kind}")
+
+    @staticmethod
+    def supported_dtypes() -> set[type]:
+        return set(DtypeRegistry._mappings.keys())
```
badmonster0: Directly returning `DtypeRegistry._mappings.keys()` is sufficient. We don't have to build a set on top of it. (for return type annotation, probably `typing.Set[type]` will just work)

python/cocoindex/typing.py
```diff
@@ -179,33 +230,51 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
...
-        elif t is datetime.time:
-            kind = "Time"
-        elif t is datetime.datetime:
-            kind = "OffsetDateTime"
-        elif t is datetime.timedelta:
-            kind = "TimeDelta"
-        else:
-            raise ValueError(f"type unsupported yet: {t}")
+        try:
+            dtype_info = DtypeRegistry.get_by_dtype(t)
```
badmonster0: Usually we want to leave exceptions for real exceptional cases. This not found is a normal case. So probably better the change `get_by_dtype()` to return `None` if not found.

docs/docs/core/data_types.mdx
```diff
@@ -36,8 +36,8 @@ This is the list of all basic types supported by CocoIndex:
 | LocalDatetime | Date and time without timezone | `cocoindex.LocalDateTime` | `datetime.datetime` |
 | OffsetDatetime | Date and time with a timezone offset | `cocoindex.OffsetDateTime` | `datetime.datetime` |
 | TimeDelta | A duration of time | `datetime.timedelta` | `datetime.timedelta` |
-| Vector[*T*, *Dim*?] | *T* must be basic type. *Dim* is a positive integer and optional. |`cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `list[T]` | 
 | Json | | `cocoindex.Json` | Any data convertible to JSON by `json` package | 
+| Vector[*T*, *Dim*?] | A vector of numbers with type *T* (`np.float32`, `np.float64`, `np.int32`, `np.int64`, `np.uint8`, `np.uint16`, `np.uint32`, or `np.uint64`). *Dim* is a positive integer and optional. | `cocoindex.Vector[T]` or `cocoindex.Vector[T, Dim]` | `numpy.typing.NDArray[T]` or `list[T]` |
```
badmonster0: Originally `Vector` isn't limited to number types, e.g. it's perfectly fine to have a `Vector[Str]`. 

I'm thinking we still want to allow both, but `NDArray` only for number types.

For `cocoindex.Vector[T]`, if `T` is a number type provided by `np`, we use `NDArray[T]` as the underlying type, otherwise use `list[T]`.

What do you think?

lemorage: Fairly good to me. This makes quite a distinction between when to use `NDArray` and when to use `list`, rather than just leaving `list` as a fallback option, as I did here.

python/cocoindex/convert.py
```diff
@@ -27,6 +29,8 @@ def encode_engine_value(value: Any) -> Any:
         ]
     if is_namedtuple_type(type(value)):
         return [encode_engine_value(getattr(value, name)) for name in value._fields]
+    if isinstance(value, np.ndarray):
+        return value.tolist()
```
badmonster0: `numpy` has its native binding in PyO3 for Rust:
https://github.com/PyO3/rust-numpy

Ideally we won't need to convert between `NDArray` and `list`.

If you prefer, I'm also OK with merging the current PR first, and treating using `rust-numpy` as a separate issue. Let me know.

python/cocoindex/typing.py
```diff
@@ -50,33 +52,41 @@ def __init__(self, key: str, value: Any):
...
-                return Annotated[list[elem_type], VectorInfo(dim=dim)]
+                # Element type and dimension provided, e.g., Vector[np.float32, Literal[3]]
+                dtype, dim_literal = params
+                # Extract the literal value
+                dim_val = (
+                    typing.get_args(dim_literal)[0]
+                    if typing.get_origin(dim_literal) is Literal
+                    else None
+                )
+                if dtype in DtypeRegistry.supported_dtypes():
```
badmonster0: nit: The positive branch is using `DtypeRegistry.get_by_dtype(dtype) is not None` but here is using `dtype in DtypeRegistry.supported_dtypes()`, for the same purpose. Let's change them to the same way for clarity.

python/cocoindex/typing.py
```diff
@@ -179,12 +231,29 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
...
+        args = typing.get_args(t)
+        _, dtype_spec = args
+
+        dtype_args = typing.get_args(dtype_spec)
+        if not dtype_args:
+            raise ValueError("Invalid dtype specification for NDArray")
+
+        numpy_dtype = dtype_args[0]
+        dtype_info = DtypeRegistry.get_by_dtype(numpy_dtype)
+        elem_type = None if dtype_info is None else dtype_info.annotated_type
```
badmonster0: We may want to raise an error if `dtype_info is None`

python/cocoindex/convert.py
lemorage: Oops, I did not realize that. Let's change this as of jetzt. Thanks for letting me know.

lemorage: feel that might be better to post another PR to address this. We can talk about a few tweaks to be made in #595.

badmonster0 (approved):

python/cocoindex/convert.py
```diff
@@ -122,6 +126,38 @@ def decode(value: Any) -> Any | None:
...
+            )
+            if expected_dim is not None and len(value) != expected_dim:
+                raise ValueError(
+                    f"Vector dimension mismatch for `{''.join(field_path)}`: "
+                    f"expected {expected_dim}, got {len(value)}"
+                )
+
+            # Use NDArray for supported numeric dtypes, else return list
+            if dtype_info is not None:
+                return np.array(value, dtype=dtype_info.numpy_dtype)
```
badmonster0, line 156: As a second thought, I think it could cause hard-to-debug issues if the target numpy type has narrower range than the source (e.g. if the target is `NDArray[npmpy.int16]` for `Vector[Int64]`). So probably we want to only support a few numpy types (`int64`, `float32` and `float64`) to be safer.

What do you think? @lemorage

lemorage: Yes, on the numpy side, I was trying to get as much support as we could. But tbh, I think unit or int16 those dtypes are rarely used in real use cases. So, rather than support them by casting, directly aborting might be more distinct. Let's change that in the next commit.

badmonster0:
Hi Miao, we love your contributions and made a section for you in latest release note
https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#lemorage

Thank you!

PR #595, opened 2025-06-07
feat: handle NumPy array vector in Python conversion

lemorage:
Relates to #586. Should have been part of it.

This PR uses PyO3 NumPy bindings to handle NumPy array vector conversions.

### DIFF
--- a/python/cocoindex/typing.py
+++ b/python/cocoindex/typing.py
@@ -111,6 +111,11 @@ class DtypeInfo:
 
 
 class DtypeRegistry:
+    """
+    Registry for NumPy dtypes used in CocoIndex.
+    Provides mappings from NumPy dtypes to CocoIndex's type representation.
+    """
+
     _mappings: dict[type, DtypeInfo] = {
         np.float32: DtypeInfo(np.float32, "Float32", float),
         np.float64: DtypeInfo(np.float64, "Float64", float),
@@ -124,6 +129,7 @@ class DtypeRegistry:
 
     @classmethod
     def get_by_dtype(cls, dtype: Any) -> DtypeInfo | None:
+        """Get DtypeInfo by NumPy dtype."""
         if dtype is Any:
             raise TypeError(
                 "NDArray for Vector must use a concrete numpy dtype, got `Any`."
@@ -132,13 +138,21 @@ class DtypeRegistry:
 
     @staticmethod
     def get_by_kind(kind: str) -> DtypeInfo | None:
+        """Get DtypeInfo by kind."""
         return next(
             (info for info in DtypeRegistry._mappings.values() if info.kind == kind),
             None,
         )
 
     @staticmethod
+    def rust_compatible_kind(kind: str) -> str:
+        """Map to a Rust-compatible kind for schema encoding."""
+        # incompatible_integer_kinds = {"Int32", "UInt8", "UInt16", "UInt32", "UInt64"}
+        return "Int64" if "Int" in kind else kind
+
+    @staticmethod
     def supported_dtypes() -> KeysView[type]:
+        """Get a list of supported NumPy dtypes."""
         return DtypeRegistry._mappings.keys()
 
 
@@ -340,8 +354,10 @@ def _encode_type(type_info: AnalyzedTypeInfo) -> dict[str, Any]:
             raise ValueError("Vector type must have a vector info")
         if type_info.elem_type is None:
             raise ValueError("Vector type must have an element type")
-        encoded_type["element_type"] = _encode_type(
-            analyze_type_info(type_info.elem_type)
+        elem_type_info = analyze_type_info(type_info.elem_type)
+        encoded_type["element_type"] = _encode_type(elem_type_info)
+        encoded_type["element_type"]["kind"] = DtypeRegistry.rust_compatible_kind(
+            elem_type_info.kind
         )
         encoded_type["dimension"] = type_info.vector_info.dim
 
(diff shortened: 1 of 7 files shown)


badmonster0:
In https://github.com/cocoindex-io/cocoindex/pull/594 we added a new testutil `validate_full_roundtrip()` that can execute a full roundtrip of conversion (involving both Python and Rust conversion logic). Can we add new tests using this for vector types in `test_convert.py`? Thanks!

src/py/convert.rs
```diff
@@ -150,16 +148,108 @@ fn basic_value_from_py_object<'py>(
...
+) -> PyResult<Bound<'py, PyAny>> {
+    match v.first() {
+        Some(value::BasicValue::Float32(_)) => {
+            let data = v
+                .iter()
+                .filter_map(|x| {
+                    if let value::BasicValue::Float32(f) = x {
+                        Some(*f)
+                    } else {
+                        None
```
badmonster0: Probably it's better to return an error if type mismatch. Silently ignoring elements with unmatched types may hide some unnoticable issues.

src/py/convert.rs
```diff
@@ -150,16 +148,108 @@ fn basic_value_from_py_object<'py>(
...
+    match elem_type {
+        &schema::BasicValueType::Float32 => try_convert!(f32, value::BasicValue::Float32),
+        &schema::BasicValueType::Float64 => try_convert!(f64, value::BasicValue::Float64),
+        &schema::BasicValueType::Int64 => {
+            try_convert!(i32, |v| value::BasicValue::Int64(v as i64));
+            try_convert!(i64, value::BasicValue::Int64);
+            try_convert!(u8, |v| value::BasicValue::Int64(v as i64));
+            try_convert!(u16, |v| value::BasicValue::Int64(v as i64));
+            try_convert!(u32, |v| value::BasicValue::Int64(v as i64));
+            try_convert!(u64, |v| value::BasicValue::Int64(v as i64));
```
badmonster0: A `u64` is not always convertible to an `i64`, so there can be problems. I think probably we can stop supporting `NDArray[uint64]`. What do you think?

python/cocoindex/typing.py
```diff
@@ -132,13 +138,21 @@ def get_by_dtype(cls, dtype: Any) -> DtypeInfo | None:
...
         return next(
             (info for info in DtypeRegistry._mappings.values() if info.kind == kind),
             None,
         )
 
+    @staticmethod
+    def rust_compatible_kind(kind: str) -> str:
+        """Map to a Rust-compatible kind for schema encoding."""
+        # incompatible_integer_kinds = {"Int32", "UInt8", "UInt16", "UInt32", "UInt64"}
+        return "Int64" if "Int" in kind else kind
```
badmonster0: We need this conversion because things like `Int32` and `Uint32` appear in `DtypeRegistry.kind`, but they are not really type kinds supported by CocoIndex. Just wonder will it work if we directly use cocoindex-supported kinds (e.g. `Int64` for all np integer types)? Then seems we don't need this conversion, and things will be simpler.

src/py/convert.rs
lemorage: Sounds good!

python/cocoindex/typing.py
lemorage: I assume changing like this?
```
    _mappings: dict[type, DtypeInfo] = {
        np.float32: DtypeInfo(np.float32, "Float32", float),
        np.float64: DtypeInfo(np.float64, "Float64", float),
        np.int32: DtypeInfo(np.int32, "Int64", int),
        np.int64: DtypeInfo(np.int64, "Int64", int),
        np.uint8: DtypeInfo(np.uint8, "Int64", int),
        np.uint16: DtypeInfo(np.uint16, "Int64", int),
        np.uint32: DtypeInfo(np.uint32, "Int64", int),
    }
```

This is quite cleaner, of course. But it brings an issue that `get_by_kind` will not work as expected, so in `convert.py`, it is hard to handle. What do u suggest?
https://github.com/cocoindex-io/cocoindex/blob/c84fc473ecf0deffadfa52ded59e9a3a5e6c17c8/python/cocoindex/convert.py#L129-L132

badmonster0: Yes, this is what's in my mind. Thanks for the explanation.

Actually how about having a new field in `AnalyzedTypeInfo`, e.g. call it `np_number_type` (value can be things like `np.float32`, `np.int32`). Currently if a type has multiple representations in Python, we already use extra fields in `AnalyzedTypeInfo` to describe it (e.g. `struct_type` for specific dataclass or named tuple type information), and np type belongs to the same type of thing. Then we leave `kind` to always represent a type kind existing in cocoindex core.

What do you think?

lemorage: Good idea!

python/cocoindex/typing.py
```diff
@@ -154,6 +153,7 @@ class AnalyzedTypeInfo:
 
     key_type: type | None  # For element of KTable
     struct_type: type | None  # For Struct, a dataclass or namedtuple
+    np_number_type: type | None  # Specific NumPy dtype
```
badmonster0: Let's make the comment more clearly about this is for element type, such as "Specify NumPy dtype for element, if it's a vector represented by `numpy.NDArray`", as other fields are usually for type at the current level.

(BTW, actually I was thinking `np_number_type` should be a property for the element type, instead of the vector type. It should resolve information one level at a time, and it'll also allows us consistently support individual `np.float32` types outside a vector - if we want to support it. But this can be improved later)

badmonster0 (approved):

badmonster0:
Really neat. Thanks a lot! Merging.

Issue #562, opened 2025-05-27
[FEATURE] Allow database not provided for some cocoindex operations (e.g. transform flow)


badmonster0:
Currently, `cocoindex.init()` is needed for most operations. `cocoindex.init()` requires a Postgres database setting (for cocoindex internal storage).

For some cases, a database as internal storage is not needed:

- Users don't want to build/update index. They only need to query, but needs to do this using CocoIndex's [transform flow](https://cocoindex.io/docs/query#transform-flow).
- In tests. For example, sometimes we only want to involve transform flows in tests (it involves both Rust and Python, so can already cover a broad range of cocoindex logic even without bringing up database!).

We should make the database setting optional, and complain on operations that requires cocoindex internal db.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

vumichien:
I'm working on it

PR #608, opened 2025-06-10
feat: allow optional database configuration

vumichien:
This PR solved #562

### DIFF
--- a/src/lib_context.rs
+++ b/src/lib_context.rs
@@ -79,7 +79,7 @@ impl DbPools {
 
 pub struct LibContext {
     pub db_pools: DbPools,
-    pub builtin_db_pool: PgPool,
+    pub builtin_db_pool: Option<PgPool>,
     pub flows: Mutex<BTreeMap<String, Arc<FlowContext>>>,
     pub all_setup_states: RwLock<setup::AllSetupState<setup::ExistingMode>>,
 }
@@ -97,6 +97,12 @@ impl LibContext {
             })?
             .clone();
         Ok(flow_ctx)
+    }
+
+    pub fn require_builtin_db_pool(&self) -> Result<&PgPool> {
+        self.builtin_db_pool
+            .as_ref()
+            .ok_or_else(|| anyhow!("Database is required for this operation. Please set COCOINDEX_DATABASE_URL environment variable and call cocoindex.init() with database settings."))
     }
 }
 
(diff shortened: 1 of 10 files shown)


src/lib_context.rs
```diff
@@ -117,11 +123,18 @@ pub fn create_lib_context(settings: settings::Settings) -> Result<LibContext> {
...
+    let (pool, all_setup_states) = if let Some(database_spec) = &settings.database {
+        let (pool, all_setup_states) = get_runtime().block_on(async {
+            let pool = db_pools.get_pool(database_spec).await?;
+            let existing_ss = setup::get_existing_setup_state(&pool).await?;
+            anyhow::Ok((Some(pool), existing_ss))
+        })?;
+        (pool, all_setup_states)
+    } else {
+        // No database configured - create empty setup states
+        (None, setup::AllSetupState::default())
```
badmonster0: I feel it might be cleaner if we use `None` to represent `AllSetupState` not available.

Moreover, I may also consider having a wrapper struct for all fields based on db, e.g. something like:

```rs
pub struct PersistenceContext {
  pub builtin_db_pool: PgPool,
  pub all_setup_states: RwLock<setup::AllSetupState<setup::ExistingMode>>,
}

pub struct LibContext {
  pub persistence_ctx: Option<PersistenceContext>,
  ...
}
```

This will guarantee these related fields co-exist by the type system.

vumichien: Good suggestions! Both improvements make sense. I implemented both in the following commit

vumichien:
@badmonster0  I'm unsure why the CI failed to create a .venv. ~~Could you help me in rerunning the CI if necessary?~~. It works now after pushing some commits that altered the environment in the GitHub workflow. This is not a permanent solution; we should address the root cause in a different pull request.

lemorage:
@vumichien personally, I think this might be due to the python version changes (?). I found the latest successful action in #611, it used Python 3.13.3 (/opt/hostedtoolcache/Python/3.13.3/x64). However, the failed ones are using Python 3.13.4 shown in the `Setup venv` log (/opt/hostedtoolcache/Python/3.13.4/x64).

badmonster0:
@vumichien @lemorage Thanks for debugging! This is because the symbolic `.venv/bin/python` in the cache is broken when a different Python version is live. https://github.com/cocoindex-io/cocoindex/pull/615 should solve the problem.

PR #620, opened 2025-06-13
feat: support scalar NumPy value encodings

### LINKED ISSUE #616: [FEATURE] Support Int64/Float32/Float64 to be bound with corresponding numpy number types
In #586 and #595 we added support to bind `Vector[Int64]`, `Vector[Float32]` and `Vector[Float64]` (as [CocoIndex data types](https://cocoindex.io/docs/core/data_types)) to corresponding numpy array types like `NDArray[numpy.int64]`, `NDArray[numpy.float32]` and `NDArray[numpy.float64]`.

For consistency and composability, we want to support numpy type binding on scalar number level too. e.g.
Users should be able to use `numpy.int64` for a field with `Int64` type, etc.

Also see a related comment https://github.com/cocoindex-io/cocoindex/pull/595#discussion_r2139060363 for implementation hint:

> BTW, actually I was thinking `np_number_type` should be a property for the element type, instead of the vector type. It should resolve information one level at a time, and it'll also allows us consistently support individual `np.float32` types outside a vector - if we want to support it.

---

lemorage:
As said in #616, this PR adds support for numpy type binding at the scalar number level.

Will close #616.


python/cocoindex/tests/test_convert.py
badmonster0, line 1: Can we add full roundtrip tests for numpy number types?

python/cocoindex/typing.py
```diff
@@ -202,11 +210,12 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
             vector_info = attr
         elif isinstance(attr, TypeKind):
             kind = attr.kind
+            if dtype_info := DtypeRegistry.get_by_kind(attr.kind):
```
badmonster0: Under this logic, even if when the base type is Python's native number types like `int` and `float`, if there're a kind, we're setting `np_number_type`. It doesn't look right to me.

Using `Float64` as an example, the following Python type annotations are possible:
1. `float` and `Annotated[float, TypeKind("Float64")]`, they're equivalent
2. `numpy.float64` and `Annotated[numpy.float64, TypeKind("Float64")]`, they're equivalent

For case 1, we shouldn't set `np_number_type` as a np type is not evolved; but for `numpy.float64` in case 2, we should set `np_number_type`. So overall I think the `np_number_type` should be set based on the base type, instead of the kind.

python/cocoindex/convert.py
```diff
@@ -127,7 +143,8 @@ def decode(value: Any) -> Any | None:
         return lambda value: uuid.UUID(bytes=value)
 
     if src_type_kind == "Vector":
-        dtype_info = DtypeRegistry.get_by_dtype(dst_type_info.np_number_type)
+        elem_type_info = analyze_type_info(dst_type_info.elem_type)
+        dtype_info = DtypeRegistry.get_by_dtype(elem_type_info.np_number_type)
```
badmonster0: This will actually create value in the unexpected type for `list[numpy.float64]`. So I think we want to use fields in `AnalyzedTypeInfo` to indicate whether the underlying type is a `NDArray`.

I think we can reuse `numpy_dtype` for this purpose. i.e. 
- At vector level, if the vector is a supported NDArray, e.g. `NDArray[numpy.float64]`, we set `numpy_dtype` to `NDArray[numpy.float64]`.
- At scalar level, we still set it to the scalar numpy type like `numpy.float64`.

When `numpy_dtype` is not set for a Vector, we want to make a decoder for the element type, iterate and call it for each element, so that we can decode any element type we support (besides `list[numpy.float64]`, we also have other not well supported today, like `list[Uuid]`)

lemorage:
Thanks for the detailed explanations, I misunderstood a bit from the suggested proposal. I am looking into those issues.

lemorage:
@badmonster0 Do u have any idea why the formatting check is always failing, even though it works well locally?

```cmd
(cocoindex) ☁  cocoindex [handle-numpy-vector]  ruff format python/cocoindex/tests/test_typing.py
1 file left unchanged
(cocoindex) ☁  cocoindex [handle-numpy-vector]  ruff --version
ruff 0.6.8
```

lemorage:
Turns out it is due to version diffs.

python/cocoindex/typing.py
```diff
@@ -233,18 +252,12 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
...
+        np_number_type = t
+        numpy_dtype = extract_ndarray_scalar_dtype(np_number_type)
+        dtype_info = DtypeRegistry.get_by_dtype(numpy_dtype)
         if dtype_info is None:
             raise ValueError(
-                f"Unsupported numpy dtype for NDArray: {np_number_type}. "
+                f"Unsupported numpy dtype for NDArray: {numpy_dtype}. "
                 f"Supported dtypes: {DtypeRegistry.supported_dtypes()}"
             )
         elem_type = dtype_info.annotated_type
```
badmonster0: Now seems we can directly use `numpy_dtype` to set `elem_type`. Then we don't even need `annotated_type` field in `DtypeInfo`. (and seems its `python_type` field can be removed too?)

lemorage: Ah right, I'll see!

python/cocoindex/typing.py
```diff
@@ -100,41 +113,40 @@ def _is_struct_type(t: ElementType | None) -> bool:
...
-        """Get DtypeInfo by NumPy dtype."""
+    def get_by_dtype(cls, dtype: Any) -> tuple[type, str] | None:
+        """Get the NumPy dtype and its CocoIndex kind by dtype."""
         if dtype is Any:
             raise TypeError(
                 "NDArray for Vector must use a concrete numpy dtype, got `Any`."
             )
-        return cls._mappings.get(dtype)
+        kind = cls._DTYPE_TO_KIND.get(dtype)
+        return None if kind is None else (dtype, kind)
```
badmonster0: The returned `dtype` is always the same as input. So it's sufficient to only return `kind`.

The same for `validate_and_get_dtype_info()` below. We can simplify the return type.

(and consider merge these two methods, as `get_by_dtype()` is as simple as a dict get now)

lemorage: Yeah, we can drop the getter method if direct access is acceptable.

Issue #621, opened 2025-06-13
[FEATURE] Skip processing the entire row if content fingerprint has no change


badmonster0:
Original requests from Discord:
> I tried to run code embedding flow via Github action whenever push event happens on the main branch. The github workflow will checkout the repository and index the code files.
> 
> What I found is that ALL source rows are always processed/updated even there's no code change. Do you have any insights on the reason of such behaviour?

Currently we use an "ordinal" (based on file modification time for file-based sources) to determine if there's content change. Sometimes modification time change without content change (e.g. `git` always use the current time after checkout), and we want to skip processing in this case too.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

vumichien:
I'm working on it

Issue #618, opened 2025-06-11
Add test on Python side to execute transform flow end to end


badmonster0:
[Transform flow](https://cocoindex.io/docs/query#transform-flow) is a sub flow without sources / targets, so it's relatively easier to run in tests (it's stateless and don't require database connection). While executing a transform flow still covers a broad range of cocoindex core logic, so it's great if we can add test to execute a transform flow.

In the test, we may define 2 transform flows (a very simple one, a more complex one with a "for each" on child rows), and use custom functions in Python for these transform flows. Evaluate these flows with given inputs, and asserts on output.

This is one step towards a full fledged end to end test involving database (which needs more work to be done and will be tracked separately)

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

PR #629, opened 2025-06-17
feat: add content hash support for change detection in source processing

vumichien:
This PR solved issue #621


src/execution/indexing_status.rs
```diff
@@ -37,7 +37,8 @@ pub async fn get_source_row_indexing_status(
         &src_eval_ctx.key,
         &interface::SourceExecutorGetOptions {
             include_value: false,
-            include_ordinal: true,
+            include_ordinal: false,
```
badmonster0: `include_ordinal` needs to be `true`.

src/execution/row_indexer.rs
```diff
@@ -91,8 +102,18 @@ impl SourceVersion {
         update_stats: Option<&stats::UpdateStats>,
     ) -> bool {
         let should_skip = match (self.ordinal.0, target.ordinal.0) {
-            (Some(orginal), Some(target_ordinal)) => {
-                orginal > target_ordinal || (orginal == target_ordinal && self.kind >= target.kind)
+            (Some(existing_ordinal), Some(target_ordinal)) => {
```
badmonster0, line 118: "ordinal" is named ordinal because we want it to be monotonic. In rare situations, we can see an older ordinal (things can happen in parallel, and an earlier read can be processed later, especially when we do it on different processes), and we always want to skip the older ordinal in this case. So we want to check ordinal order first, and always return `true` if the target is older. Hash only needs to be compared otherwise.

On the other hand, one more complexity is, even if the hash equals, we cannot really fully skip. For example, consider seeing the following in order:

1. ordinal = 1, hash= a
2. ordinal = 3, hash = a
3. ordinal = 2, hash = b 

If we skip for "2", for "3" we'll process the input with a lower ordinal, and the content "b" will override "a", which is not expected because the maximum ordinal is 3 with content hash "a".

This means even if hash is the same, we still want to update ordinal fields.

---

So I think we can revert the change here. Note that "ordinal" indicates a monotonic invariance that we don't want to violate - that's why we're very carefully checking ordinal order on multiple phases. Checking for "hash" is essentially an optimization. We only need to check it once, e.g. at the beginning of `update_source_row()`  (or before calling it).

- If the content hash + logic doesn't change, we only need to update the tracking table. We can perform similar logic in precommit + commit within a single transaction. `max_process_ordinal`, `processed_source_ordinal` and `process_ordinal` need to be updated.
- Otherwise, run the regular logic.

src/ops/interface.rs
```diff
@@ -99,34 +102,43 @@ pub struct SourceChangeMessage {
 #[derive(Debug, Default)]
 pub struct SourceExecutorListOptions {
     pub include_ordinal: bool,
+    /// Include content hash for change detection.
+    /// When enabled, sources should compute and return content hashes.
+    pub include_content_hash: bool,
```
badmonster0: I think probably we don't have to expose this through the source. We can compute a fingerprint in the core logic based on the content returned from the source.

We have two types of sources.

*  Those doesn't provide light way to access a hash, e.g. `LocalFile`.
 Reading the content during "list" seems too heavy. It also significantly increase read bytes.

    1. For an empty flow without data, both "list" and "get" will need to read the content, so doubles the read bytes, versus only read content in "get" before. (1X->2X)
    2. For an existing flow with almost no updates and source ordinal preserved, we don't need to read any content before, but we'll need to read all content during "list" (0->1X)
    3. Only under the scenario that source ordinal not preserved, this won't increase bytes read.

*  Those provide light way to access a hash, like S3, there're builtin hash (ETag) that we don't need to read the full content. Under case "3" above, exposing hash from source may be beneficial, but the gain is also limited, because: (1) getting ETag also needs separate API calls, so for 1+2 it's still extra overhead; (2) the chance of case 3 for sources like S3 is low, as its ordinal is usually more stable - unlike local files.

For "get" API, currently a hash get always comes together with a content get, so also not a saving.

So, in my mind, overall the cost of having this is higher than the benefit. It also adds more complexity to the source implementation. So I prefer not to add this into the interface. What do you think?

badmonster0:
Thanks for taking this! It touches a core aspect of cocoindex. Left a few thoughts. Welcome of more discussions!

src/execution/indexing_status.rs
vumichien: resolved

src/ops/interface.rs
vumichien: It makes sense. Revise as a suggestion

src/execution/row_indexer.rs
vumichien: Thank you for the clear explanation. Revised it according to your suggestions.

src/execution/row_indexer.rs
```diff
@@ -538,18 +616,68 @@ pub async fn update_source_row(
             let existing_version = SourceVersion::from_stored(
                 info.processed_source_ordinal,
                 &info.process_logic_fingerprint,
+                &info.processed_source_content_hash,
```
badmonster0: Since now we switched to read the content hash from db here, I think we can add it as a field of `StoredMemoizationInfo` - this also avoids a db schema upgrade (which is a hurdle for users).

src/execution/row_indexer.rs
```diff
@@ -472,6 +489,67 @@ async fn commit_source_tracking_info(
...
+
+    // Get existing target keys to preserve them
+    let existing_target_keys = tracking_info
+        .as_ref()
+        .and_then(|info| info.target_keys.as_ref())
+        .map(|keys| keys.0.clone())
+        .unwrap_or_default();
+
+    // Update tracking info with new ordinal but keep existing target keys
+    db_tracking::commit_source_tracking_info(
```
badmonster0: - `max_process_ordinal` field needs to be updated - it's to make sure the next time during precommit the generated process ordinal is monotonic.
- To be safer, we don't want to set the staging keys to empty. We don't want to touch it.  The reason of having "staging" is mainly becuase there can be an earlier process terminated between "precommit" and "commit", in this case we rely on next time of processing to issue deletions for the target with the staging target keys. Cleaning them up may have the risk of leaking content in the target permanently.

So I think probably we want to have a specific method within `db_tracking.rs`, to only update the fields we need to update under this situation.

src/execution/row_indexer.rs
```diff
@@ -472,6 +489,67 @@ async fn commit_source_tracking_info(
...
+
+    // Read precommit info to get existing target keys
+    let tracking_info = db_tracking::read_source_tracking_info_for_precommit(
+        source_id,
+        source_key_json,
+        db_setup,
+        &mut *txn,
+    )
+    .await?;
+
```
badmonster0, line 544: We still need to check the source ordinal, to make sure there's no new input coming in simultaneously, similar to this

https://github.com/cocoindex-io/cocoindex/blob/ffed325c4197de292a508b0e0cdbd98f0c2fd7b9/src/execution/row_indexer.rs#L173-L182

src/ops/interface.rs
badmonster0, line 1: Can we revert all changes for files under `ops`, as they're no longer needed.

src/execution/row_indexer.rs
```diff
@@ -538,18 +616,68 @@ pub async fn update_source_row(
...
+                .into_fingerprint(),
+        ),
+        interface::SourceValue::NonExistence => None,
+    };
+
+    // Create source version with computed content hash
+    let source_version_with_hash = SourceVersion {
+        ordinal: source_version.ordinal,
+        kind: source_version.kind,
+        content_hash,
```
badmonster0: Since content hash is only present in this specific phase as an optimization, and `SourceVersion` is needed more broadly for a more critical purpose, probably we don't need to add it to `SourceVersion`.

src/execution/row_indexer.rs
vumichien: You're right. The content hash optimization needs to check the source ordinal first to maintain monotonic invariance

src/execution/row_indexer.rs
vumichien: Got it

src/execution/db_tracking.rs
```diff
@@ -84,6 +83,8 @@ pub struct SourceTrackingInfoForProcessing {
 
     pub processed_source_ordinal: Option<i64>,
     pub process_logic_fingerprint: Option<Vec<u8>>,
+    #[allow(dead_code)]
+    pub processed_source_content_hash: Option<Vec<u8>>,
```
badmonster0: This is not used, so can be cleaned up?

src/execution/db_tracking_setup.rs
```diff
@@ -42,6 +42,14 @@ async fn upgrade_tracking_table(
         );
         sqlx::query(&query).execute(pool).await?;
     }
+    
+    if existing_version_id < 2 && target_version_id >= 2 {
```
badmonster0: Since we start to use `memoization_info` for fingerprint, this is no longer needed.

src/execution/row_indexer.rs
```diff
@@ -535,36 +559,76 @@ pub async fn update_source_row(
...
+            let stored_info = evaluation_memory.into_stored()?;
+
+            // Content hash optimization: if content hasn't changed, use fast path
+            // Verify ordinal allows processing
+            if let (Some(existing_version), Some(existing_info), Some(existing_content_hash)) = (
+                &existing_version,
+                &memoization_info,
+                &stored_info.content_hash,
+            ) {
+                if !existing_version.should_skip(source_version, None)
```
badmonster0: `should_skip` check here is unnecessary, as it's already checked at L568 above.

On the other hand, we need to add the following checks:
- the existing version is based on the same logic here (`kind == CurrentLogic`)
- `max_process_ordinal == process_ordinal`: this is because `memoization_info` and `max_process_ordinal` are written together (during precommit), and only when `max_process_ordinal == process_ordinal` we can guarantee commit for that already succeeded (we need to read more fields from db).

src/execution/row_indexer.rs
```diff
@@ -535,36 +559,76 @@ pub async fn update_source_row(
...
+                    && existing_info.content_hash.as_ref() == Some(existing_content_hash)
+                {
+                    // Ordinal allows processing AND content unchanged - use fast path
+                    db_tracking::update_source_ordinal_and_fingerprints_only(
```
badmonster0: We still need to perform this check within the same transaction:

1. Same check as https://github.com/cocoindex-io/cocoindex/blob/93196184dab1b7b0546175338ee945e98fb5f2b0/src/execution/row_indexer.rs#L174-L180

  Skip all following processings if we see a newer version.

2. Check `process_ordinal` again, to make sure it hasn't changed. If changed, it means there's another version processed in between, and  it's not newer than current, because we already passed check "1" above). We should fallback to the ordinary processing in this case.

Since two conditions need to be checked and there're different outcomes on each condition (skip vs fallback to ordinary), it's not doable by a single `UPDATE` statement - we still need a transaction.

---

Besides, as a second thought, actually I don't need to update `max_process_ordinal` and `process_ordinal` (if both check 1 and 2 above pass, it already prevents the possibility of new version in between). We can simplify the update statement a little bit.

vumichien: That sounds like a great idea. If both checks—1 (new version detection) and 2 (process_ordinal unchanged)—have passed, we can be confident that there isn't a new version in between. In this case, we only need to update the minimal tracking fields to accurately reflect the current source state without triggering any downstream processes. This approach is much simpler than my update_source_ordinal_and_fingerprints_only function, which unnecessarily updated more fields.

src/execution/row_indexer.rs
vumichien: Remove redundant should_skip check

src/execution/db_tracking.rs
vumichien: cleanup

src/execution/db_tracking_setup.rs
vumichien: cleanup

src/execution/row_indexer.rs
```diff
@@ -525,46 +548,137 @@ pub async fn update_source_row(
...
             let output =
                 evaluate_source_entry(src_eval_ctx, source_value, &evaluation_memory).await?;
-            (Some(output), evaluation_memory.into_stored()?)
+            let stored_info = evaluation_memory.into_stored()?;
+
+            if let (Some(existing_info), Some(existing_content_hash)) =
+                (&memoization_info, &stored_info.content_hash)
+            {
+                if can_use_content_hash_optimization
+                    && existing_info.content_hash.as_ref() == Some(existing_content_hash)
```
badmonster0: This is checked too late. We want it checked before `evaluate_source_entry()` is called, since the main computation happens within `evaluate_source_entry()` and the main purpose of this optimization is to skip this computation.

src/execution/row_indexer.rs
```diff
@@ -525,46 +548,137 @@ pub async fn update_source_row(
...
+                            return Ok(SkippedOr::Skipped(current_source_version));
+                        }
+
+                        // Check 2: Verify process_ordinal hasn't changed (no concurrent processing)
+                        let original_process_ordinal = original_tracking_info
+                            .as_ref()
+                            .and_then(|info| info.process_ordinal);
+                        if current_info.process_ordinal != original_process_ordinal {
+                        } else {
+                            let query_str = format!(
```
badmonster0: Please follow existing convention to execute SQL expressions within `db_tracking.rs`.

vumichien: Create a proper function in db_tracking.rs

src/execution/row_indexer.rs
vumichien: Moved content hash check earlier

src/execution/row_indexer.rs
```diff
@@ -500,7 +523,7 @@ pub async fn evaluate_source_entry_with_memory(
             src_eval_ctx.key,
             &SourceExecutorGetOptions {
                 include_value: true,
-                include_ordinal: false,
+                include_ordinal: true,
```
badmonster0: We're not using the ordinal in this case. We don't need `include_ordinal` to be `true`.

src/execution/row_indexer.rs
```diff
@@ -525,46 +548,148 @@ pub async fn update_source_row(
...
+                    .max_process_ordinal
+                    .zip(info.process_ordinal)
+                    .map(|(max_ord, proc_ord)| max_ord == proc_ord)
+                    .unwrap_or(false);
+
+            let memoization_info = info
+                .memoization_info
+                .as_ref()
+                .and_then(|info| info.0.as_ref())
+                .cloned();
```
badmonster0: `memonization_info` can be large. Let's avoid cloning it.

We only need to take out the specific fields from  `info` that needs to be read later, instead of the entire `info`, so this clone can be avoided.

src/execution/row_indexer.rs
```diff
@@ -525,46 +548,148 @@ pub async fn update_source_row(
...
-                memoization_info,
+                memoization_info.clone(),
                 EvaluationMemoryOptions {
                     enable_cache: true,
                     evaluation_only: false,
                 },
             );
+
+            // Compute and set content hash from source data
+            let content_hash = Fingerprinter::default()
```
badmonster0: Computing content hash isn't light. We should compute it only once before checking for the optimization.

src/execution/row_indexer.rs
```diff
@@ -525,46 +548,148 @@ pub async fn update_source_row(
...
+    if let interface::SourceValue::Existence(ref source_value) = source_value {
+        if let Some(existing_info) = &memoization_info {
+            if can_use_content_hash_optimization {
+                // Compute content hash directly from source data (no expensive evaluation needed)
+                let current_content_hash = Fingerprinter::default()
+                    .with(source_value)?
+                    .into_fingerprint();
+
+                if existing_info.content_hash.as_ref() == Some(&current_content_hash) {
+                    // Content hash matches - try optimization
```
badmonster0: It may be cleaner if we extract a major part of this into a separate function. The function may return `Result<Option<SkippedOr<()>>>` (`None` if check 2 doesn't pass, so fall through)

src/execution/db_tracking.rs
```diff
@@ -305,3 +306,34 @@ pub async fn read_source_last_processed_info(
...
+         process_time_micros = $6 \
+         WHERE source_id = $1 AND source_key = $2",
+        db_setup.table_name
+    );
+    sqlx::query(&query_str)
+        .bind(source_id) // $1
+        .bind(source_key_json) // $2
+        .bind(processed_source_ordinal) // $3
+        .bind(process_logic_fingerprint) // $4
+        .bind(process_ordinal) // $5
```
badmonster0: We don't need to set either `process_logic_fingerprint` and `process_ordinal`, since we already verified their value before calling this.

src/execution/row_indexer.rs
vumichien: Changed include_ordinal to false

src/execution/db_tracking.rs
vumichien: Removed them

src/execution/row_indexer.rs
vumichien: Good idea

src/execution/row_indexer.rs
```diff
@@ -525,46 +613,128 @@ pub async fn update_source_row(
...
+        }
+    }
+
     let (output, stored_mem_info) = match source_value {
         interface::SourceValue::Existence(source_value) => {
-            let evaluation_memory = EvaluationMemory::new(
+            let mut evaluation_memory = EvaluationMemory::new(
                 process_time,
-                memoization_info,
+                memoization_info.clone(),
```
badmonster0: Can we also avoid this clone? `MemonizationInfo` can be large.

src/execution/row_indexer.rs
```diff
@@ -525,46 +613,128 @@ pub async fn update_source_row(
...
+        interface::SourceValue::Existence(source_value) => {
+            Some(Fingerprinter::default()
+                .with(source_value)?
+                .into_fingerprint())
+        }
+        interface::SourceValue::NonExistence => None,
+    };
+
+    if let interface::SourceValue::Existence(ref _source_value) = source_value {
+        if let Some(ref existing_hash) = existing_content_hash {
```
badmonster0: Here can do something like this to merge two layers of `if`:

```rs
if can_use_content_hash_optimization {
  if let (Some(existing_hash), Some(current_hash)) = (&existing_content_hash, &current_content_hash) {
    ...
  }
}
```

src/execution/row_indexer.rs
```diff
@@ -525,46 +613,128 @@ pub async fn update_source_row(
...
+    let current_content_hash = match &source_value {
+        interface::SourceValue::Existence(source_value) => {
+            Some(Fingerprinter::default()
+                .with(source_value)?
+                .into_fingerprint())
+        }
+        interface::SourceValue::NonExistence => None,
+    };
+
+    if let interface::SourceValue::Existence(ref _source_value) = source_value {
```
badmonster0: This layer of `if` seems unnecessary, and `_source_value` is not used any way.

src/execution/row_indexer.rs
```diff
@@ -525,46 +613,128 @@ pub async fn update_source_row(
...
                 EvaluationMemoryOptions {
                     enable_cache: true,
                     evaluation_only: false,
                 },
             );
+
+            // Use the pre-computed content hash
+            if let Some(hash) = current_content_hash {
+                evaluation_memory.set_content_hash(hash);
+            }
```
badmonster0, line 709: Seems `content_hash` doesn't have to be a field of `EvaluationMemory`. We can directly put something like this. 

```rs
let mut stored_info = evaluation_memory.into_stored()?;
stored_info.content_hash = current_content_hash;
```

src/execution/row_indexer.rs
```diff
@@ -525,46 +613,128 @@ pub async fn update_source_row(
...
-            (Some(output), evaluation_memory.into_stored()?)
+            let stored_info = evaluation_memory.into_stored()?;
+
+            (Some(output), stored_info)
+        }
+        interface::SourceValue::NonExistence => {
+            let stored_info = StoredMemoizationInfo {
+                cache: Default::default(),
+                uuids: Default::default(),
+                content_hash: current_content_hash,
```
badmonster0: It should always be `None` for this branch. We can restore to use `Default::default()` for this entire branch.

src/execution/row_indexer.rs
badmonster0: The `cloned()` is still here.

Actually there's another way (which may be simpler than what I suggested in last comment): we change `match existing_tracking_info` to `match &existing_tracking_info` (i.e. don't move `existing_tracking_info` here), and don't even need to touch `memoization_info` here. `memoization_info` is only used in Line 712 below, and we can move it out of `existing_tracking_info` there.

vumichien: Thanks! I'll fix it.

src/execution/row_indexer.rs
vumichien: Thanks for pointing it out

src/execution/row_indexer.rs
```diff
@@ -525,33 +613,86 @@ pub async fn update_source_row(
...
+                        .map(|(max_ord, proc_ord)| max_ord == proc_ord)
+                        .unwrap_or(false);
+
+                let content_hash = info
+                    .memoization_info
+                    .as_ref()
+                    .and_then(|info| info.0.as_ref())
+                    .and_then(|stored_info| stored_info.content_hash.as_ref())
+                    .cloned();
+
```
badmonster0, line 647: Now seems we can move the logic to compute `can_use_optimization` and `existing_content_hash` close to where they're used. Logic will be more clear. e.g.

```rs
let existing_version = match &existing_tracking_info {
   ... 
};

// Compute content hash once if needed for both optimization and evaluation
let current_content_hash = match &source_value {
    ....
};

if let (Some(current_hash), Some(existing_tracking_info)) = (&current_content_hash, &existing_tracking_info) {
  if let Some(optimization_result) =
    try_content_hash_optimization(
                src_eval_ctx, &source_key_json, source_version, &process_time,
                current_hash, existing_tracking_info, &existing_version, update_stats, pool).await? {
    ....
  }
}
```

Then `can_use_optimization` and `existing_content_hash` can be computed based on `existing_tracking_info` and `existing_version` within `try_content_hash_optimization()` - logic related to the optimization will be kept together and easier to maintain.

src/execution/source_indexer.rs
```diff
@@ -62,11 +64,19 @@ impl SourceIndexingContext {
             rows.insert(
                 source_key,
                 SourceRowIndexingState {
-                    source_version: SourceVersion::from_stored(
-                        key_metadata.processed_source_ordinal,
-                        &key_metadata.process_logic_fingerprint,
-                        plan.logic_fingerprint,
-                    ),
+                    source_version: SourceVersion {
```
badmonster0: What's the purpose of this change? Shall we revert it?

python/cocoindex/tests/test_content_hash.py
badmonster0, line 1: I really appreciate adding so many tests! Thanks ❤️

Note that these tests are irrelevant to this content hash optimization. A _transform flow_ is fully evaluated in memory and doesn't involve the database. But it's great to keep these tests under a different file name / class name. It's even better if this test can be separated out as a different PR  (they actually covers what we want to achieve for https://github.com/cocoindex-io/cocoindex/issues/618).

src/execution/source_indexer.rs
vumichien: I think we can revert it

python/cocoindex/tests/test_content_hash.py
vumichien: removed the test

src/execution/row_indexer.rs
vumichien: Refactor related logic

src/execution/memoization.rs
```diff
@@ -84,26 +87,26 @@ pub struct EvaluationMemory {
 impl EvaluationMemory {
     pub fn new(
         current_time: chrono::DateTime<chrono::Utc>,
-        stored_info: Option<StoredMemoizationInfo>,
+        stored_info: Option<&StoredMemoizationInfo>,
```
badmonster0: Passing this by reference means additional clone, which can be a lot given `StoredMemoizationInfo` can be big. Can we revert this change? We should move `StoredMemoizationInfo` out of `existing_tracking_info` (please also see my comment left at the call site)

src/execution/row_indexer.rs
```diff
@@ -525,33 +636,68 @@ pub async fn update_source_row(
...
+            return Ok(optimization_result);
+        }
+    }
+
     let (output, stored_mem_info) = match source_value {
         interface::SourceValue::Existence(source_value) => {
+            let memoization_info = existing_tracking_info
+                .as_ref()
+                .and_then(|info| info.memoization_info.as_ref())
+                .and_then(|info| info.0.as_ref());
```
badmonster0: We don't need to take ref here. Just simplify move out the value.


```rs
            let memoization_info = existing_tracking_info
                .and_then(|info| info.memoization_info)
                .and_then(|info| info.0);
```

Then we can move `memoization_info` when calling `EvaluationMemory::new()` - which will save a lot of clone there.

src/execution/row_indexer.rs
```diff
@@ -472,6 +495,94 @@ async fn commit_source_tracking_info(
...
+        .and_then(|info| info.0.as_ref())
+        .and_then(|stored_info| stored_info.content_hash.as_ref());
+
+    let Some(existing_hash) = existing_hash else {
+        return Ok(None);
+    };
+
+    if existing_hash != current_hash {
+        return Ok(None);
+    }
```
badmonster0, line 532: The two `if` above can be merged into something like:

```rs
if existing_hash != Some(current_hash) {
  return Ok(None)
}
```

src/execution/row_indexer.rs
```diff
@@ -472,6 +495,94 @@ async fn commit_source_tracking_info(
...
+    source_version: &SourceVersion,
+    process_time: &chrono::DateTime<chrono::Utc>,
+    current_hash: &crate::utils::fingerprint::Fingerprint,
+    tracking_info: &db_tracking::SourceTrackingInfoForProcessing,
+    existing_version: &Option<SourceVersion>,
+    update_stats: &stats::UpdateStats,
+    pool: &PgPool,
+) -> Result<Option<SkippedOr<()>>> {
+    // Check if we can use content hash optimization
+    let can_use_optimization = existing_version
```
badmonster0: I think having a `can_use_optimization` here is unnecessary and its name is somewhat confusing - it doesn't really cover all cases that cannot do optimization.

It might be simpler if we directly check each condition and just return:

```rs
if !existing_version
        .as_ref()
        .map_or(false, |v| v.kind == SourceVersionKind::CurrentLogic) {
  return Ok(None);
}

if !tracking_info
            .max_process_ordinal
            .zip(tracking_info.process_ordinal)
            .map_or(false, |(max_ord, proc_ord)| max_ord == proc_ord) {
  return Ok(None);
}
```

src/execution/row_indexer.rs
```diff
@@ -472,6 +495,94 @@ async fn commit_source_tracking_info(
...
+
+    let current_tracking_info = db_tracking::read_source_tracking_info_for_precommit(
+        src_eval_ctx.import_op.source_id,
+        source_key_json,
+        &src_eval_ctx.plan.tracking_table_setup,
+        &mut *txn,
+    )
+    .await?;
+
+    if let Some(current_info) = current_tracking_info {
```
badmonster0: nit: Prefer reusing the name `current_tracking_info` to be more informative.

badmonster0: An early return pattern can help to reduce nested level of code. e.g.

```rs
let Some(current_tracking_info) = current_tracking_info else {
  return Ok(None);
};

...
```

src/execution/row_indexer.rs
```diff
@@ -472,6 +495,94 @@ async fn commit_source_tracking_info(
...
+            &current_info,
+            src_eval_ctx.plan.logic_fingerprint,
+        );
+        if current_source_version.should_skip(source_version, Some(update_stats)) {
+            return Ok(Some(SkippedOr::Skipped(current_source_version)));
+        }
+
+        // Check 2: Verify process_ordinal hasn't changed (no concurrent processing)
+        let original_process_ordinal = tracking_info.process_ordinal;
+        if current_info.process_ordinal == original_process_ordinal {
```
badmonster0: nit: We can try to do an early return here too, e.g.

```rs
if current_info.process_ordinal != original_process_ordinal {
  return Ok(None);
}

...
```

It reduces nested level, and makes all branches not applicable to optimization consistently return early, leaving the optimization logic to the end - which is easier to follow.

src/execution/db_tracking.rs
```diff
@@ -305,3 +306,31 @@ pub async fn read_source_last_processed_info(
...
+    process_ordinal: i64,
+    process_time_micros: i64,
+    db_setup: &TrackingTableSetupState,
+    db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
+) -> Result<()> {
+    let query_str = format!(
+        "UPDATE {} SET \
+         processed_source_ordinal = $3, \
+         process_ordinal = $4, \
+         process_time_micros = $5 \
```
badmonster0: We don't need to update `process_ordinal` and `process_time_micros` (we had more discussion [here](https://github.com/cocoindex-io/cocoindex/pull/629#discussion_r2159815852))

src/execution/memoization.rs
vumichien: We still need the existing tracking information to create a content hash. Therefore, I will extract the memoization information early like this.

```rust
let extracted_memoization_info = existing_tracking_info
    .as_mut()
    .and_then(|info| info.memoization_info.take()) 
    .and_then(|info| info.0);
```
    
Does it make sense?

src/execution/db_tracking.rs
```diff
@@ -311,25 +311,17 @@ pub async fn update_source_tracking_ordinal_and_logic(
...
     db_setup: &TrackingTableSetupState,
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
 ) -> Result<()> {
     let query_str = format!(
-        "UPDATE {} SET \
-         processed_source_ordinal = $3, \
-         process_ordinal = $4, \
-         process_time_micros = $5 \
-         WHERE source_id = $1 AND source_key = $2",
+        "UPDATE {} SET processed_source_ordinal = $3 WHERE source_id = $1 AND source_key = $2",
```
badmonster0, line 318: Now the function can be renamed to `update_source_tracking_ordinal`. It no longer touches `logic`.

src/execution/row_indexer.rs
```diff
@@ -499,39 +499,35 @@ async fn try_content_hash_optimization(
     src_eval_ctx: &SourceRowEvaluationContext<'_>,
     source_key_json: &serde_json::Value,
     source_version: &SourceVersion,
-    process_time: &chrono::DateTime<chrono::Utc>,
+    _process_time: &chrono::DateTime<chrono::Utc>,
```
badmonster0: It's no longer used. Just remove this.

src/execution/memoization.rs
badmonster0: Before this latest change, it was extracted right before `EvaluationMemory::new()`. If we do this there, do we still need to access `existing_tracking_info` later? I think this way is better because:

- We don't need to pass an additional `extracted_memoization_info` to 
`try_content_hash_optimization()`, as things can be retrieved from `existing_tracking_info` - it makes the function signature simpler.
- `take()` is a little bit tricky - sometimes we need it, but it means we need to manually track which field is gone. This is a large function and the live scope for `existing_tracking_info` is also large, so easily lose track especially after later changes. "move" means compiler helps us to watch out.

vumichien: Thank you for the explanation.

src/execution/row_indexer.rs
```diff
@@ -487,7 +591,7 @@ pub async fn evaluate_source_entry_with_memory(
         )
         .await?;
         existing_tracking_info
-            .and_then(|info| info.memoization_info.map(|info| info.0))
+            .and_then(|mut info| info.memoization_info.take().map(|info| info.0))
```
badmonster0: Why do we need to make this change? A simple move (as before) should just work.

src/execution/row_indexer.rs
```diff
@@ -525,46 +629,86 @@ pub async fn update_source_row(
...
+        )
+        .await?
+        {
+            return Ok(optimization_result);
+        }
+    }
+
+    let (output, stored_mem_info) = {
+        let extracted_memoization_info = existing_tracking_info
+            .and_then(|mut info| info.memoization_info.take())
```
badmonster0: A simple move (`.and_then(|info| info.memoization_info)`) should just work here

vumichien: Thank you for pointing it out

badmonster0 (approved):
All look good! Thanks for your patience! 👍

vumichien:
@badmonster0 
Thank you for supporting and guiding me through this PR.

badmonster0:
Hi @vumichien, 
Thank you so much for your contribution! We love it and made a section for you in the latest release note!
https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#vumichien

Really appreciate your great work!

Issue #627, opened 2025-06-16
[FEATURE] Support LiteLLM (Proxy) as LLM


badmonster0:
[LitelLLM](https://github.com/BerriAI/litellm) provides a hub to connect to different LLM models. It provides a [proxy server](https://docs.litellm.ai/docs/proxy/docker_quick_start).
We want to support talking to it, as a separate API in addition to Ollama, OpenAI, Gemini and Anthropic.

https://github.com/cocoindex-io/cocoindex/blob/800ceb24d11174033d187140b3f2892985d58b00/src/llm/mod.rs#L10-L16

See a similar example PR of adding Anthropic support: https://github.com/cocoindex-io/cocoindex/pull/395

Note: LiteLLM proxy API provides an OpenAI-compatible API, so it should be OK to reuse most OpenAI implementation when possible.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

par4m:
I can work on this, please assign.

PR #630, opened 2025-06-17
feat(llm): Add LiteLLM (Proxy) Support

par4m:
This PR aims to add support for LiteLLM as a Proxy to run multiple LLMs as mentioned in the issue https://github.com/cocoindex-io/cocoindex/issues/627

Defaults url to `http://localhost:4000`
To Test,

Install LiteLLM Proxy - `pip install 'litellm[proxy]'`

1. OpenAI, create `config.yml`
```bash
model_list:
  - model_name: "*"             # all requests where model not in your config go to this deployment
    litellm_params:
      model: openai/*           # set `openai/` to use the openai route
      api_key: os.environ/OPENAI_API_KEY

```
2. DeepSeek
Run DeepSeek - `ollama pull deepseek-r1`
`config.yml`
```bash
model_list:
  - model_name: "deepseek-r1"
    litellm_params:
      model: "ollama_chat/deepseek-r1"
      api_base: "http://localhost:11434"

```
Run LiteLLM 
`litellm --config config.yml`

In `main.py` set LLM Accordingly
```python
cocoindex.functions.ExtractByLlm(
                llm_spec=cocoindex.LlmSpec(
                    api_type=cocoindex.LlmApiType.LITELLM,
                    model="deepseek-r1",  # or whatever model is set on your LiteLLM proxy
                    # address can be omitted to use http://0.0.0.0:4000 by default
                ),
```

### DIFF
--- a/docs/docs/ai/llm.mdx
+++ b/docs/docs/ai/llm.mdx
@@ -121,3 +121,68 @@ cocoindex.LlmSpec(
 
 You can find the full list of models supported by Anthropic [here](https://docs.anthropic.com/en/docs/about-claude/models/all-models).
 
+### LiteLLM
+
+To use the LiteLLM API, you need to set the environment variable `LITELLM_API_KEY`.
+
+#### 1. Install LiteLLM Proxy
+
+```bash
+pip install 'litellm[proxy]'
+```
+
+#### 2. Create a `config.yml` for LiteLLM
+
+**Example for OpenAI:**
+```yaml
+model_list:
+  - model_name: "*"             
+    litellm_params:
+      model: openai/*           
+      api_key: os.environ/LITELLM_API_KEY
+```
+
+**Example for DeepSeek:**
+
+First, pull the DeepSeek model with Ollama:
+```bash
+ollama pull deepseek-r1
+```
+Then run it if it's not running:
+```bash
+ollama run deepseek-r1
+```
+
+Then, use this in your `config.yml`:
+```yaml
+model_list:
+  - model_name: "deepseek-r1"
+    litellm_params:
+      model: "ollama_chat/deepseek-r1"
+      api_base: "http://localhost:11434"
+```
+
+#### 3. Run LiteLLM Proxy
+
+```bash
+litellm --config config.yml
+```
+
+#### 4. A Spec for LiteLLM will look like this:
+
+<Tabs>
+<TabItem value="python" label="Python" default>
+
+```python
+cocoindex.LlmSpec(
+    api_type=cocoindex.LlmApiType.LITELLM,
+    model="deepseek-r1", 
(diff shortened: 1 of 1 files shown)


src/llm/litellm.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub struct Client {
+    client: async_openai::Client<OpenAIConfig>,
+    model: String,
+}
+
+impl Client {
+    pub async fn new(spec: super::LlmSpec) -> Result<Self> {
+        // For LiteLLM, use provided address or default to http://0.0.0.0:4000
+        let address = spec.address.clone().unwrap_or_else(|| "http://0.0.0.0:4000".to_string());
```
badmonster0: I think it's better to be `127.0.0.1`. `0.0.0.0` is usually used when binding to a port of arbitrary host address (e.g. a machine can have both `127.0.0.1` loopback and a `192.168.1.10` as a LAN IP address), but when making an ongoing connection, we should usually use a specific address.

src/llm/litellm.rs
```diff
@@ -0,0 +1,103 @@
...
+        let api_key = std::env::var("LITELLM_API_KEY").unwrap_or_else(|_| "anything".to_string());
```
badmonster0: `"anything"` looks a little bit arbitrary. If `api_key` not set, I think probably we can skip calling `with_api_key`  instead of using a placeholder.

src/llm/litellm.rs
```diff
@@ -0,0 +1,103 @@
...
+};
+use async_trait::async_trait;
+
+pub struct Client {
+    client: async_openai::Client<OpenAIConfig>,
+    model: String,
+}
+
+impl Client {
+    pub async fn new(spec: super::LlmSpec) -> Result<Self> {
```
badmonster0: Looks like most part of this file is the same as `openai.rs`, except `new()` method. Maybe consider reusing the `Client` implementation from `openai.rs` - we only need to provide a separate `new` method.

badmonster0:
Can you also help to update the documentation to describe how to use it there? Thanks!

par4m:
> Can you also help to update the documentation to describe how to use it there? Thanks!

Sure! Should I include both DeepSeek (ollama) and OpenAI examples ?

badmonster0:
> > Can you also help to update the documentation to describe how to use it there? Thanks!
> 
> Sure! Should I include both DeepSeek (ollama) and OpenAI examples ?

Sounds good. Thanks!

badmonster0:
Thanks for the PR!

badmonster0:
Hi @par4m Param, Thank you so much for your contribution.
We made a section for you in our latest release note, we love your contribution and thank you ❤️!

https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#par4m

Issue #634, opened 2025-06-18
[FEATURE] Support OpenRouter as LLM


par4m:
**What is the use case?**
Add Support for OpenRouter as LLM. 

**Describe the solution you'd like**

Same way as done with LiteLLM as OpenRouter supports OpenAI Config

**Additional context**

https://openrouter.ai/docs/community/frameworks#using-the-openai-sdk
https://openrouter.ai/deepseek/deepseek-r1:free/api

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

par4m:
I can work on this!

PR #635, opened 2025-06-18
feat(llm): Add OpenRouter Support

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/634

Reuses the OpenAI Client config

Sets base url to `https://openrouter.ai/api/v1` as specified by OpenRouter.

Docs are also added for the same.

### DIFF
--- a/python/cocoindex/llm.py
+++ b/python/cocoindex/llm.py
@@ -9,8 +9,8 @@ class LlmApiType(Enum):
     OLLAMA = "Ollama"
     GEMINI = "Gemini"
     ANTHROPIC = "Anthropic"
-    LITELLM = "LiteLlm"
-    OPENROUTER = "OpenRouter"
+    LITE_LLM = "LiteLlm"
+    OPEN_ROUTER = "OpenRouter"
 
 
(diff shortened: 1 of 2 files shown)


python/cocoindex/llm.py
```diff
@@ -10,6 +10,7 @@ class LlmApiType(Enum):
     GEMINI = "Gemini"
     ANTHROPIC = "Anthropic"
     LITELLM = "LiteLlm"
+    OPENROUTER = "OpenRouter"
```
cocoindex-dev: By default we should use `_` to separate different words in UPPER_CASE convention. So this should be `OPEN_ROUTER`.

BTW, the previous one needs to be `LITE_LLM`  (we didn't use `OPEN_AI` as this is how most other places reference it, but by default we should use `_` between words)

cocoindex-dev (approved):

par4m:
Renamed both LlmApiType, OPENROUTER -> OPEN_ROUTER and LITELLM -> LITE_LLM. Also changed in the docs for the usage.

PR #638, opened 2025-06-19
fix(example): ensure qdrant collections exist in startup event

### LINKED ISSUE #637: [BUG] `image_search` example never returns results
**Describe the bug**
I found this while working on #496. I used `image_search` as a test, but it never returned any results. Plus, there were some missing declared dependencies.

**To Reproduce**
Just follow the steps in https://github.com/cocoindex-io/cocoindex/blob/main/examples/image_search/README.md.

**Expected behavior**
In my case, regardless of the changes I make, it never returns a result to me. I think this should be something wrong on the Qdrant side, but I don't know if this is a universal problem or if it just happens on my machine. Ideally, it should give me the correct pic by the query.

Here are some logs I made when debugging.
```
2025-06-19 17:50:21,390 - main - INFO - Received search query: elephant
2025-06-19 17:50:21,390 - main - INFO - Loading CLIP model: openai/clip-vit-large-patch1
2025-06-19 17:50:28,023 - main - INFO - Query embedding for 'A photo of elephant' generated, length: 768
2025-06-19 17:50:28,032 - main - INFO - Search returned 0 results
INFO:     127.0.0.1:49991 - "GET /search?q=elephant HTTP/1.1" 200 OK
```

lemorage:
This PR does three things.
1. Add the missing declared dependencies in `pyproject.toml`.
2. Deprecate `@app.on_event("startup")` and use FastAPI lifespan context manager instead.
3. Always make sure the Qdrant collection already exists when starting up the app.

Fixes #637.


examples/image_search/main.py
```diff
@@ -86,8 +84,31 @@ def image_object_embedding_flow(
...
+    app.state.qdrant_client = QdrantClient(url=QDRANT_URL, prefer_grpc=True)
+
+    app.state.qdrant_client.recreate_collection(
+        collection_name=QDRANT_COLLECTION,
+        vectors_config={
+            "embedding": VectorParams(
+                size=CLIP_MODEL_DIMENSION, distance=Distance.COSINE
+            )
+        },
+    )
```
badmonster0: We want to avoid users manually do. `cocoindex setup` is for this (we have a [blog](https://cocoindex.io/blogs/schema-inference-for-qdrant) discussing the considerations around this).

CocoIndex doesn't provide an corresponding API for `setup` now. Ideally after there is such an API, we can call `image_object_embedding_flow.setup()` here (I created #639 for this with more context).

lemorage: Thanks, cool, I see. So I'll leave this PR here for now. After #639 is resolved, this PR can be used for updating the associated examples.

lemorage: actually, this can be a pure update for missing dependencies and deprecated code.

badmonster0: Sounds good. Thanks!

PR #641, opened 2025-06-21
ops: add pre-commit and hooks for code checking

### LINKED ISSUE #633: [FEATURE] Add pre-commit and use hooks for code checking
**What is the use case?**
In #632, we added a script to check all items before committing the code. This approach works by relying on users installing the required tools (`ruff`, `mypy`, etc.) and running this script manually every time before committing.

This is a good use case for [pre-commit](https://pre-commit.com/). There are several advantages from my side.
1. **Tool Installation is easy.** We can use `pre-commit` to install all the required tools in one go, and it will use exact versions of tools.
2. **Checking is automatic.** We no longer need to run the checks by ourselves. `Pre-commit` runs automatically when we make commits.
3. **Speed**. `Pre-commit` only runs relevant hooks on staged files, speeding up a bit than our current script checking.

1. Following the [pre-commit documentation](https://pre-commit.com/), create the `.pre-commit-config.yaml` file, and add necessary hooks.
2. Make sure the `pre-commit` works and then add steps to set up pre-commit in **contributing docs** accordingly.

**Additional context**

---

lemorage:
This PR adds the pre-commit and hooks for checking, replacing the previous script checking.

All the checking steps in `check.sh` are ported. Plus, a few recommended official hooks are used.

Close #633.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+        pass_filenames: true
+  
+  - repo: https://github.com/christophmeissner/pytest-pre-commit
+    rev: 1.0.0
+    hooks:
+      - id: pytest
+        language: system
+        types: [python]
+        pass_filenames: false
+        always_run: false
```
badmonster0: I think we need to make sure `pytest` is executed after `maturin develop`, as the test needs to call compiled Rust `.so`.

lemorage: Oh my, I changed to use the third-party pytest hook in the last commit, but I forgot the ordering. Initially it used local repo. Thanks for catching this!

PR #653, opened 2025-06-24
feat: support image input to LLM clients

### LINKED ISSUE #496: [FEATURE] Support passing image as input to LLM
Currently our LLM interface and `ExtractByLlm` builtin function only supports text prompt. We want it to support image inputs too.

- Extend `LlmGenerateRequest` to accept image bytes as input:
    https://github.com/cocoindex-io/cocoindex/blob/f368a9aff550b14e6a9580c200cb1f65da4e394f/src/llm/mod.rs#L34
- Implement existing LLM clients to pass this to different APIs (usually as base64)
- Extend the [`ExtractByLlm` function](https://github.com/cocoindex-io/cocoindex/blob/f368a9aff550b14e6a9580c200cb1f65da4e394f/src/ops/functions/extract_by_llm.rs) to accept an optional image as input
- Can be tested by the existing [`image_search_example`](https://github.com/cocoindex-io/cocoindex/tree/f368a9aff550b14e6a9580c200cb1f65da4e394f/examples/image_search_example) example. It's using customized code to send image to LLM now. We should be able to replace it with the `ExtractByLlm` function.

---

lemorage:
Closes #496.

### DIFF
--- a/examples/image_search/main.py
+++ b/examples/image_search/main.py
@@ -15,10 +15,9 @@ from PIL import Image
 from qdrant_client import QdrantClient
 from transformers import CLIPModel, CLIPProcessor
 
+OLLAMA_URL = os.getenv("OLLAMA_URL", "http://localhost:11434/")
 QDRANT_URL = os.getenv("QDRANT_URL", "http://localhost:6334/")
 QDRANT_COLLECTION = "ImageSearch"
-OLLAMA_URL = os.getenv("OLLAMA_URL", "http://localhost:11434/")
-OLLAMA_MODEL = "gemma3"
 CLIP_MODEL_NAME = "openai/clip-vit-large-patch14"
 CLIP_MODEL_DIMENSION = 768
 
@@ -71,22 +70,14 @@ def image_object_embedding_flow(
     )
     img_embeddings = data_scope.add_collector()
     with data_scope["images"].row() as img:
-        has_ollama_model = os.getenv("OLLAMA_MODEL") == OLLAMA_MODEL
-        if has_ollama_model:
+        ollama_model_name = os.getenv("OLLAMA_MODEL")
+        if ollama_model_name is not None:
+            # If an Ollama model is specified, generate an image caption
             img["caption"] = flow_builder.transform(
                 cocoindex.functions.ExtractByLlm(
                     llm_spec=cocoindex.llm.LlmSpec(
-                        api_type=cocoindex.LlmApiType.OLLAMA, model=OLLAMA_MODEL
+                        api_type=cocoindex.LlmApiType.OLLAMA, model=ollama_model_name
                     ),
-                    # Replace by this spec below, to use OpenAI API model instead of ollama
-                    #   llm_spec=cocoindex.LlmSpec(
-                    #       api_type=cocoindex.LlmApiType.OPENAI, model="gpt-4o"),
-                    # Replace by this spec below, to use Gemini API model
-                    #   llm_spec=cocoindex.LlmSpec(
-                    #       api_type=cocoindex.LlmApiType.GEMINI, model="gemini-2.0-flash"),
-                    # Replace by this spec below, to use Anthropic API model
-                    #   llm_spec=cocoindex.LlmSpec(
-                    #       api_type=cocoindex.LlmApiType.ANTHROPIC, model="claude-3-5-sonnet-latest"),
                     instruction=(
                         "Describe the image in one detailed sentence. "
                         "Name all visible animal species, objects, and the main scene. "
@@ -105,13 +96,11 @@ def image_object_embedding_flow(
             "embedding": img["embedding"],
         }
 
-        if has_ollama_model:
-            print(
-                f"Ollama model '{OLLAMA_MODEL}' is available — captions will be extracted."
-            )
+        if ollama_model_name is not None:
+            print(f"Using Ollama model '{ollama_model_name}' for captioning.")
             collect_fields["caption"] = img["caption"]
         else:
-            print(f"Ollama model '{OLLAMA_MODEL}' not found — skipping captioning.")
+            print(f"No Ollama model '{ollama_model_name}' found — skipping captioning.")
 
         img_embeddings.collect(**collect_fields)
 


src/llm/mod.rs
```diff
@@ -135,3 +139,11 @@ pub fn new_llm_embedding_client(
     };
     Ok(client)
 }
+
+pub fn detect_mime_type(bytes: &[u8]) -> Result<&'static str> {
```
badmonster0: `detect_image_mime_type` is a better name.

src/llm/mod.rs
```diff
@@ -1,9 +1,12 @@
 use crate::prelude::*;
 
 use crate::base::json_schema::ToJsonSchemaOptions;
+use infer::Infer;
 use schemars::schema::SchemaObject;
 use std::borrow::Cow;
 
+static INFER: OnceLock<Infer> = OnceLock::new();
```
badmonster0: Slightly prefer `LazyLock` for this case (initialization logic stays together).

src/llm/ollama.rs
```diff
@@ -18,6 +19,7 @@ enum OllamaFormat<'a> {
 struct OllamaRequest<'a> {
     pub model: &'a str,
     pub prompt: &'a str,
+    pub images: Option<Vec<String>>,
```
badmonster0: We can change this to a `Option<&'a [u8]>`. We can use https://docs.rs/serde_with/latest/serde_with/base64/struct.Base64.html to serialize bytes as base64 directly. This will save one copy.

src/ops/functions/extract_by_llm.rs
badmonster0, line 1: `get_system_prompt()` also needs to be updated

src/ops/functions/extract_by_llm.rs
```diff
@@ -110,14 +129,28 @@ impl SimpleFunctionFactoryBase for Factory {
...
-            },
-            spec.output_type.clone(),
-        ))
+        let mut args = Args {
+            text: None,
+            image: None,
+        };
+
+        // Handle positional argument
+        if let Some(arg) = args_resolver.next_optional_arg("")? {
```
badmonster0: I'm a little bit hesitant about how we want to resolve the argument here. Empty arg name looks a little bit weird, and using data type to distinguish text/image won't always work (e.g. there may be more media types later).

If it's a Python function, the signature will be something like:

```py
def extract_by_llm(text: str = None, image: bytes = None):
  ...
```

So users can call `extract_by_llm('text')`, `extract_by_llm(text='text')`, `extract_by_llm(image=some_image)`, or even pass in both like `extract_by_llm(text='text', image=some_image)` (it's reasonable for `extract_by_llm` to accept both). 

I know we have a limitation that cocoindex function can only be called in chain style (`arg0.transform(FunctionSpec(...), remaining_args...)`). I thought about introducing another API to allow a non-chain style of function call, but didn't have a clear idea about the style of the API. Maybe this is the time to make a decision.

I created #654 for this purpose (if you prefer, welcome to make take that one too, then the current one will be easier to resolve, otherwise I can also resolve that myself :)

lemorage: Thanks! I'll look into it.

src/ops/functions/extract_by_llm.rs
```diff
@@ -30,10 +30,11 @@ struct Executor {
...
     let mut message =
-        "You are a helpful assistant that extracts structured information from text. \
-Your task is to analyze the input text and output valid JSON that matches the specified schema. \
-Be precise and only include information that is explicitly stated in the text. \
-Output only the JSON without any additional messages or explanations."
+        "You are a helpful assistant that processes user-provided inputs (text, images, or both) to produce structured outputs. \
+Your task is to follow the provided instructions to generate or extract information and output valid JSON matching the specified schema. \
+Base your response solely on the content of the input. \
+For generative tasks, respond accurately and relevantly based on what is provided. \
+Unless explicitly instructed otherwise, output only the JSON—do not include explanations, descriptions, or formatting outside the JSON."
```
badmonster0: nit: here the hyphen more like a in-word connector. Maybe rephrase like this to be more clear:

> ... output only the JSON. DO NOT include explanations, descriptions, or formatting outside the JSON.

src/ops/functions/extract_by_llm.rs
```diff
@@ -113,8 +133,11 @@ impl SimpleFunctionFactoryBase for Factory {
         Ok((
             Args {
                 text: args_resolver
-                    .next_arg("text")?
+                    .next_optional_arg("text")?
                     .expect_type(&ValueType::Basic(BasicValueType::Str))?,
+                image: args_resolver
+                    .next_optional_arg("image")?
+                    .expect_type(&ValueType::Basic(BasicValueType::Bytes))?,
```
badmonster0: We want to check here to ensure one of `text` or `image` must be provided (this will raise error on flow building time instead of execution time)

examples/image_search/main.py
```diff
@@ -69,10 +69,35 @@ def image_object_embedding_flow(
...
+                instruction=(
+                    "Describe the image in one detailed sentence. "
+                    "Name all visible animal species, objects, and the main scene. "
+                    "Be specific about type, color, and notable features. "
+                    "Mention what each animal is doing."
+                ),
+                output_type=str,
+            ),
+            image=img["content"],
+        )
```
badmonster0: Now users have to set a `GEMINI_API_KEY` when running this example. The bar to launch this example is slightly higher now.

Probably we should make this one optional. e.g. check if `GEMINI_API_KEY` is set, and only compute caption / export the additional field when it's set. 

Also the `search()` below may be modified to output the caption field (if exists) - to make this new field really useful.

Besides, would you also update `README.md` for this example to mention that `GEMINI_API_KEY` can be provided optionally to enable caption?

src/llm/gemini.rs
```diff
@@ -70,10 +71,28 @@ impl LlmGenerationClient for Client {
...
-        // Compose the prompt/messages
+        let mut user_parts: Vec<serde_json::Value> = Vec::new();
+
+        // Add text part first
+        user_parts.push(serde_json::json!({ "text": request.user_prompt }));
+
+        // Add image part if present
+        if let Some(image_bytes) = &request.image {
+            let base64_image =
+                base64::engine::general_purpose::STANDARD.encode(image_bytes.as_ref());
```
badmonster0: We can consider `use base64::prelude::*` (like [this](https://github.com/cocoindex-io/cocoindex/blob/22cab1db264cca0a9e90c07991e64f5ba81d6e88/src/utils/fingerprint.rs#L2)), then directly use `BASE64_STANDARD` here.

examples/image_search/main.py
lemorage: Thanks! I checked the old `image_search` example (https://github.com/cocoindex-io/cocoindex/tree/f368a9aff550b14e6a9580c200cb1f65da4e394f/examples/image_search_example). I'd like to stick to the original docs and use the same `ollama/gemma3` model for this example. Differently, I made it optional.

examples/image_search/main.py
```diff
@@ -17,10 +18,21 @@
...
 CLIP_MODEL_DIMENSION = 768
 
 
+def ollama_has_model(model) -> bool:
+    try:
+        r = requests.get(f"{OLLAMA_URL}/api/tags", timeout=1)
+        r.raise_for_status()
+        return any(m.get("name") == model for m in r.json().get("models", []))
+    except Exception:
+        return False
```
badmonster0: Instead of doing this check, how about simply check a environment variable? For example, we may ask users to set `OLLAMA_MODEL` as an environment variable. If it's not set, we skip the new logic.

We want to keep examples simple if possible. It'll make users easier to learn how to use cocoindex by reading these examples.

examples/image_search/main.py
```diff
@@ -69,12 +81,49 @@ def image_object_embedding_flow(
     )
     img_embeddings = data_scope.add_collector()
     with data_scope["images"].row() as img:
+        has_gemma3 = ollama_has_model(OLLAMA_MODEL)
```
badmonster0: Maybe use something like `has_ollama_model` instead of `has_gemma3` instead.

examples/image_search/main.py
```diff
@@ -69,12 +71,49 @@ def image_object_embedding_flow(
     )
     img_embeddings = data_scope.add_collector()
     with data_scope["images"].row() as img:
+        has_ollama_model = os.getenv("OLLAMA_MODEL") == OLLAMA_MODEL
```
badmonster0: This seems weird. Now if users set `OLLAMA_MODEL` env to something different from `gemma3` we're just ignoring it?

Can we simply use whatever `OLLAMA_MODEL` set by users, and skip related logic when it's not set?

lemorage: Sure!

badmonster0:
thank you @lemorage ! new release note is out and we made a section for you, we love your contribution!!
https://cocoindex.io/blogs/cocoindex-changelog-2025-08-18#lemorage ❤️

PR #662, opened 2025-06-25
add contributing guide to CONTRIBUTING.md

aayo-python:
Description of the changes
updated the contributing.md to include information such as quick start for new contributors


badmonster0 (commented):
Thanks for making this PR!

We already have a contributing guide at https://github.com/cocoindex-io/cocoindex/blob/main/docs/docs/about/contributing.md. We intentionally make this file simply containing a URL to that, to avoid duplicating content at two different places. 

Would you like to update this PR to revise https://github.com/cocoindex-io/cocoindex/blob/main/docs/docs/about/contributing.md directly for things not covered there?

Thank you!

PR #663, opened 2025-06-25
Migrated custom UUID conversion logic to pyo3.uuid package

### LINKED ISSUE #652: [FEATURE] Migrate custom UUID conversion logic to `pyo3.uuid` package
Currently we're passing uuid as bytes between Rust and Python, so we have special conversion logic on both sides:

- Rust: https://github.com/cocoindex-io/cocoindex/blob/4296c2d1f52796e795362ccab37c5b5ba2e1095b/src/py/convert.rs#L140 
 https://github.com/cocoindex-io/cocoindex/blob/4296c2d1f52796e795362ccab37c5b5ba2e1095b/src/py/convert.rs#L72

- Python: https://github.com/cocoindex-io/cocoindex/blob/4296c2d1f52796e795362ccab37c5b5ba2e1095b/python/cocoindex/convert.py#L45-L46 https://github.com/cocoindex-io/cocoindex/blob/4296c2d1f52796e795362ccab37c5b5ba2e1095b/python/cocoindex/convert.py#L95-L96

And an existing test for Union type actually cannot handle Uuid correctly: https://github.com/cocoindex-io/cocoindex/blob/4296c2d1f52796e795362ccab37c5b5ba2e1095b/python/cocoindex/tests/test_convert.py#L580-L583 (in the expected test, we should use `value = uuid.uuid4()` without `.bytes`)

PyO3 has a [`uuid` module](https://pyo3.rs/main/doc/pyo3/uuid/), which can directly bind Python's UUID with Rust's UUID without manually converting from/to bytes. We should switch to this, and all code above can be simplified / fixed.

---

SaiSakthidar:
# # Summary
 - Enable the uuid feature in PyO3 to support UUID type conversions

- Update Rust and Python code to pass and receive uuid.UUID objects directly

- Refactor tests to use real uuid.UUID instances instead of string representations

Closes #652

### DIFF
--- a/python/cocoindex/convert.py
+++ b/python/cocoindex/convert.py
@@ -42,8 +42,6 @@ def encode_engine_value(value: Any) -> Any:
         return [
             [encode_engine_value(k)] + encode_engine_value(v) for k, v in value.items()
         ]
-    if isinstance(value, uuid.UUID):
-        return value
     return value
 
 
@@ -91,9 +89,6 @@ def make_engine_value_decoder(
                 f"Type mismatch for `{''.join(field_path)}`: "
                 f"passed in {src_type_kind}, declared {dst_annotation} ({dst_type_info.kind})"
             )
-
-    if src_type_kind == "Uuid":
-        return lambda value: value
 
     if dst_type_info is None:
         if src_type_kind == "Struct" or src_type_kind in TABLE_TYPES:


python/cocoindex/convert.py
```diff
@@ -43,7 +43,7 @@ def encode_engine_value(value: Any) -> Any:
             [encode_engine_value(k)] + encode_engine_value(v) for k, v in value.items()
         ]
     if isinstance(value, uuid.UUID):
-        return value.bytes
+        return value
```
badmonster0: We can remove this entire `if` statement. Just fall through to the default.

python/cocoindex/convert.py
```diff
@@ -93,7 +93,7 @@ def make_engine_value_decoder(
             )
 
     if src_type_kind == "Uuid":
-        return lambda value: uuid.UUID(bytes=value)
+        return lambda value: value
```
badmonster0: Same as above, we can remove this `if` statement and fall through to the default.

SaiSakthidar:
@badmonster0 
Made the required changes now

badmonster0 (approved):

badmonster0:
hi @SaiSakthidar thank you for you contribution!
We made a section for you in the latest release note, thanks you!
https://cocoindex.io/blogs/cocoindex-changelog-2025-07-07/#saisakthidar

PR #674, opened 2025-06-29
Make sentence_transformers an optional dependency

### LINKED ISSUE #636: Make sentence_transformers an optional dependency
`sentence_transformers` is included as a builtin function - it's not needed for all cases, and its dependency is heavy. We want to make it an optional dependency, and only provide the `SentenceTransformerEmbed` function  when the dependency is installed.

Existing examples / documentations also need to be updated.

---

theparthgupta:
## Summary
This PR makes `sentence_transformers` an optional dependency as requested in #636. The dependency is heavy and not needed for all use cases, so users can now choose whether to install it.

## Changes Made
- **pyproject.toml**: Moved `sentence_transformers` from required dependencies to optional dependencies under the `sentence-transformers` extra
- **python/cocoindex/features.py**: Added conditional import logic to only define `SentenceTransformerEmbed` when the dependency is available
- **python/cocoindex/__init__.py**: Updated to conditionally export `SentenceTransformerEmbed` only when available
- **examples/**: Added installation instructions to all example files that use `SentenceTransformerEmbed`

## Installation Options
Users can now install with:
```bash
# With sentence-transformers support
pip install 'cocoindex[sentence-transformers]'

# Or install the dependency separately
pip install sentence-transformers

Fixes #636

### DIFF
--- a/docs/docs/getting_started/quickstart.md
+++ b/docs/docs/getting_started/quickstart.md
@@ -1,294 +1,294 @@
-    ---
-    title: Quickstart
-    description: Get started with CocoIndex in 10 minutes
-    ---
+---
+title: Quickstart
+description: Get started with CocoIndex in 10 minutes
+---
 
-    import ReactPlayer from 'react-player'
+import ReactPlayer from 'react-player'
 
-    # Build your first CocoIndex project
+# Build your first CocoIndex project
 
-    This guide will help you get up and running with CocoIndex in just a few minutes, that does:
-    *   Read files from a directory
-    *   Perform basic chunking and embedding
-    *   loads the data into a vector store (PG Vector)
+This guide will help you get up and running with CocoIndex in just a few minutes, that does:
+*   Read files from a directory
+*   Perform basic chunking and embedding
+*   loads the data into a vector store (PG Vector)
 
-    <ReactPlayer controls url='https://www.youtube.com/watch?v=gv5R8nOXsWU' />
+<ReactPlayer controls url='https://www.youtube.com/watch?v=gv5R8nOXsWU' />
 
-    ## Prerequisite: Install CocoIndex environment
+## Prerequisite: Install CocoIndex environment
 
-    We'll need to install a bunch of dependencies for this project.
+We'll need to install a bunch of dependencies for this project.
 
-    1.  Install CocoIndex:
-
-        ```bash
-        pip install -U 'cocoindex[sentence-transformers]'
-        ```
-
-    2.  You can skip this step if you already have a Postgres database with pgvector extension installed.
-        If not, the easiest way is to bring up a Postgres database using docker compose:
-
-        - Make sure Docker Compose is installed: [docs](https://docs.docker.com/compose/install/)
-        - Start a Postgres SQL database for cocoindex using our docker compose config:
-
-        ```bash
-        docker compose -f <(curl -L https://raw.githubusercontent.com/cocoindex-io/cocoindex/refs/heads/main/dev/postgres.yaml) up -d
-        ```
-
-    ## Step 1: Prepare directory for your project
-
-    1.  Open the terminal and create a new directory for your project:
-
-        ```bash
-        mkdir cocoindex-quickstart
-        cd cocoindex-quickstart
-        ```
-
-    2.  Prepare input files for the index. Put them in a directory, e.g. `markdown_files`.
-        If you don't have any files at hand, you may download the example [markdown_files.zip](markdown_files.zip) and unzip it in the current directory.
-
-    ## Step 2: Define the indexing flow
-
-    Create a new file `quickstart.py` and import the `cocoindex` library:
-
-    ```python title="quickstart.py"
-    import cocoindex
-    ```
-
-    Then we'll create the indexing flow as follows.
-
-    ```python title="quickstart.py"
-    @cocoindex.flow_def(name="TextEmbedding")
-    def text_embedding_flow(flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope):
-        # Add a data source to read files from a directory
-        data_scope["documents"] = flow_builder.add_source(
-            cocoindex.sources.LocalFile(path="markdown_files"))
-
-        # Add a collector for data to be exported to the vector index
-        doc_embeddings = data_scope.add_collector()
-
-        # Transform data of each document
-        with data_scope["documents"].row() as doc:
-            # Split the document into chunks, put into `chunks` field
-            doc["chunks"] = doc["content"].transform(
-                cocoindex.functions.SplitRecursively(),
-                language="markdown", chunk_size=2000, chunk_overlap=500)
-
-            # Transform data of each chunk
-            with doc["chunks"].row() as chunk:
-                # Embed the chunk, put into `embedding` field
-                chunk["embedding"] = chunk["text"].transform(
-                    cocoindex.functions.SentenceTransformerEmbed(
-                        model="sentence-transformers/all-MiniLM-L6-v2"))
-
-                # Collect the chunk into the collector.
-                doc_embeddings.collect(filename=doc["filename"], location=chunk["location"],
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: examples/amazon_s3_embedding/main.py [hunk 1]
```diff
@@ -1,3 +1,13 @@
+"""
+This example uses SentenceTransformerEmbed which requires the optional sentence-transformers dependency.
+
+Install with:
+    pip install 'cocoindex[sentence-transformers]'
+
+Or install the dependency directly:
+    pip install sentence-transformers
+"""
+
```
badmonster0: How about update `pyprojects.py` for each example, so the dependency will be installed automatically when users do `pip install .`.

lemorage:
Hey, personally, I feel that for the extra name, `sentence-transformers` is a bit long and does not quite make sense. Following [PEP 621](https://peps.python.org/pep-0621/#dependencies-optional-dependencies), **Extras** collect specific subsets of optional dependencies based on **features**. Maybe we can change it to some descriptive names like `ml` or `embeddings`. This way, it is also quite extensible for future changes.

examples/amazon_s3_embedding/main.py
theparthgupta: Sure! I'll update the documentation and add sentence-transformers to the pyproject.toml files for the relevant examples. Give me a bit to make those changes.

theparthgupta:
I've updated all 7 examples that use `SentenceTransformerEmbed` with the new optional dependency approach.

I noticed that the main README.md and several docs files also reference `SentenceTransformerEmbed`. Should I update those files with the new installation instructions as well, or would you prefer to handle the documentation updates yourself?

Files that might need updates:
- README.md (main example uses SentenceTransformerEmbed)  
- docs/quickstart.md
- docs/basics.md
- docs/functions.md
- docs/custom_function.mdx

examples/amazon_s3_embedding/README.md
```diff
@@ -36,6 +36,17 @@ Install dependencies:
 ```sh
 pip install -e .
 ```
+**Note**: This example uses 'SentenceTransformerEmbed' which requires the sentence-transformers library. If you encounter import errors, install it with:
```
badmonster0: Since we already put this as dependency in `pyproject.toml`, and `pip install -e .` above will cover it, I don't think we need a separate instruction to manually install it.

examples/amazon_s3_embedding/main.py
badmonster0: Same for comments here, since we already put this as dependency in `pyproject.toml`, we don't need a separate instruction to manually install it. After following the instruction in `README.md` (`pip install -e .`) things will work.

badmonster0:
> Hey, personally, I feel that for the extra name, `sentence-transformers` is a bit long and does not quite make sense. Following [PEP 621](https://peps.python.org/pep-0621/#dependencies-optional-dependencies), **Extras** collect specific subsets of optional dependencies based on **features**. Maybe we can change it to some descriptive names like `ml` or `embeddings`. This way, it is also quite extensible for future changes.

Agree, `sentence-transformers` is indeed too long and probably too specific. +1 for using `embedding` as the feature name.

badmonster0:
> I've updated all 7 examples that use `SentenceTransformerEmbed` with the new optional dependency approach.
> 
> I noticed that the main README.md and several docs files also reference `SentenceTransformerEmbed`. Should I update those files with the new installation instructions as well, or would you prefer to handle the documentation updates yourself?
> 
> Files that might need updates:
> 
> * README.md (main example uses SentenceTransformerEmbed)
> * docs/quickstart.md
> * docs/basics.md
> * docs/functions.md
> * docs/custom_function.mdx

It'll be great if you can help to update these docs in the current PR:

* docs/quickstart.md (in the instruction for `pip install cocoindex`, we need to add this new optional feature)
* docs/functions.md (we need to tell users this `SentenceTransformerEmbed` function needs this new optional feature)

Other docs may be just mentioning `SentenceTransformerEmbed`. Seems they don't need to be updated.

`README.md` usually doesn't need to be updated either - after the dependency is added into `pyproject.toml`, the dependency will be automatically installed in `pip install`.

Thanks!

examples/amazon_s3_embedding/README.md
theparthgupta: Ok will remove these from all 7 examples

examples/amazon_s3_embedding/main.py
theparthgupta: will remove these too

theparthgupta:
I've addressed all the feedback regarding documentation updates:
Changes made:
- ✅ **docs/quickstart.md**: Updated pip install command to `pip install -U 'cocoindex[sentence-transformers]'`
- ✅ **docs/functions.md**: Added optional dependency note for `SentenceTransformerEmbed` function with clear installation instructions
- ✅ **All 7 examples**: Removed manual installation instructions from README.md and main.py files as requested

docs/docs/getting_started/quickstart.md
```diff
@@ -1,294 +1,294 @@
----
-title: Quickstart
-description: Get started with CocoIndex in 10 minutes
----
+    ---
```
badmonster0: We don't need to add additional indents to every single line of this file. Please revert the changes and only keep changes relevant to the purpose of this PR. Thanks!

badmonster0:
> > Hey, personally, I feel that for the extra name, `sentence-transformers` is a bit long and does not quite make sense. Following [PEP 621](https://peps.python.org/pep-0621/#dependencies-optional-dependencies), **Extras** collect specific subsets of optional dependencies based on **features**. Maybe we can change it to some descriptive names like `ml` or `embeddings`. This way, it is also quite extensible for future changes.
> 
> Agree, `sentence-transformers` is indeed too long and probably too specific. +1 for using `embedding` as the feature name.

@theparthgupta what do you think about this? Would you also change the feature name to `embedding`? Thanks!

docs/docs/getting_started/quickstart.md
theparthgupta: reverted this

theparthgupta:
> > > Hey, personally, I feel that for the extra name, `sentence-transformers` is a bit long and does not quite make sense. Following [PEP 621](https://peps.python.org/pep-0621/#dependencies-optional-dependencies), **Extras** collect specific subsets of optional dependencies based on **features**. Maybe we can change it to some descriptive names like `ml` or `embeddings`. This way, it is also quite extensible for future changes.
> > 
> > 
> > Agree, `sentence-transformers` is indeed too long and probably too specific. +1 for using `embedding` as the feature name.
> 
> @theparthgupta what do you think about this? Would you also change the feature name to `embedding`? Thanks!

Agreed on changing it! I'd lean toward embeddings (plural) since it typically handles multiple embeddings, but either embedding or embeddings works well. Much better than the current long name. If you agree i can work on this too

badmonster0:
> embeddings

`embeddings` looks good. Thanks!

theparthgupta:
> > embeddings
> 
> `embeddings` looks good. Thanks!

Sounds good! I'll get started on the implementation.

theparthgupta:
I've updated the optional dependency group name from sentence-transformers to embeddings as discussed. Let me know if you'd like any other naming adjustments or additions

badmonster0:
Thanks for this change!

theparthgupta:
> Thanks for this change!

thanks for merging the PR!

 I noticed CI is failing on main after my PR was merged — looks like mypy can't find sentence_transformers or numpy. I think it's due to the optional import. Should I open a quick fix to update mypy.ini or CI deps?

lemorage:
@theparthgupta The CI failure is largely due to the missing `numpy` dep, after making `sentence_transformers` optional. I've made the fix, and feel free to check out #676.

theparthgupta:
> @theparthgupta The CI failure is largely due to the missing `numpy` dep, after making `sentence_transformers` optional. I've made the fix, and feel free to check out #676.

thanks for doing it

PR #675, opened 2025-06-29
feat: add `transform` method to `FlowBuilder`

### LINKED ISSUE #654: [FEATURE] Support non-chain-style cocoindex function call
Currently cocoindex function can only be called in chain style (`arg0.transform(FunctionSpec(...), remaining_args...)`). It's convenient for many cases, but there're situations that not fits into this. e.g. in Python we may have a function like:

```py
def extract_from_llm(text: str = None, image: bytes = None):
  ...
```

Here both `text` and `image` are optional. There's no clear equivalence in chain style

For the non-chain-style function calls, two options in my mind for now:

1. `FunctionSpec(...).call(args...)`
    * More explicit
    * But more verbose, and `FunctionSpec` cannot have `call` field

2. `FunctionSpec(...)(args...)`
    * More concise, no restriction
    * But two adjacent `(...)(...)` looks a little bit uncommon
  
Would like to hear some feedback about the desired style before proceeds.

---

lemorage:
This PR implements the suggested regime mentioned in #654.

Relates to #653, and closes #654.

### DIFF
--- a/python/cocoindex/op.py
+++ b/python/cocoindex/op.py
@@ -5,13 +5,13 @@ Facilities for defining cocoindex operations.
 import asyncio
 import dataclasses
 import inspect
-
-from typing import Protocol, Any, Callable, Awaitable, dataclass_transform
 from enum import Enum
+from typing import Any, Awaitable, Callable, Protocol, dataclass_transform
 
-from .typing import encode_enriched_type, resolve_forward_ref
-from .convert import encode_engine_value, make_engine_value_decoder
 from . import _engine  # type: ignore
+from .convert import dump_engine_object, encode_engine_value, make_engine_value_decoder
+from .flow import DataSlice, _create_data_slice, _spec_kind, _to_snake_case
+from .typing import encode_enriched_type, resolve_forward_ref
 
 
 class OpCategory(Enum):
@@ -49,7 +49,51 @@ class SourceSpec(metaclass=SpecMeta, category=OpCategory.SOURCE):  # pylint: dis
 
 
 class FunctionSpec(metaclass=SpecMeta, category=OpCategory.FUNCTION):  # pylint: disable=too-few-public-methods
-    """A function spec. All its subclass can be instantiated similar to a dataclass, i.e. ClassName(field1=value1, field2=value2, ...)"""
+    """A function spec. Can be instantiated and called like a function: spec(args...).
+    For non-chain-style calls, use spec(args...) with at least one DataSlice argument.
+    For chain-style calls, use data_slice.transform(spec, args...).
+    """
+
+    def __call__(self, *args: Any, **kwargs: Any) -> DataSlice[Any]:
+        """Execute the function, returning a DataSlice."""
+
+        data_slice_args = [arg for arg in args if isinstance(arg, DataSlice)]
+        data_slice_kwargs = {
+            k: v for k, v in kwargs.items() if isinstance(v, DataSlice)
+        }
+        if not data_slice_args and not data_slice_kwargs:
+            raise ValueError(
+                "At least one DataSlice argument is required to provide flow context"
+            )
+
+        first_data_slice = (
+            data_slice_args[0]
+            if data_slice_args
+            else list(data_slice_kwargs.values())[0]
+        )
+        flow_builder_state = first_data_slice._state.flow_builder_state
+
+        transform_args: list[tuple[Any, str | None]] = [
+            (flow_builder_state.get_data_slice(v), None) for v in args if v is not None
+        ]
+        transform_args += [
+            (flow_builder_state.get_data_slice(v), k)
+            for k, v in kwargs.items()
+            if v is not None
+        ]
+
+        return _create_data_slice(
+            flow_builder_state,
+            lambda target_scope, name: flow_builder_state.engine_flow_builder.transform(
+                _spec_kind(self),
+                dump_engine_object(self),
+                transform_args,
+                target_scope,
+                flow_builder_state.field_name_builder.build_name(
+                    name, prefix=_to_snake_case(_spec_kind(self)) + "_"
+                ),
+            ),
+        )
 
 
 class TargetSpec(metaclass=SpecMeta, category=OpCategory.TARGET):  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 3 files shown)


python/cocoindex/op.py
```diff
@@ -311,8 +355,7 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
                 return fn(*args, **kwargs)
 
         class _Spec(FunctionSpec):
-            def __call__(self, *args: Any, **kwargs: Any) -> Any:
-                return fn(*args, **kwargs)
+            pass
```
badmonster0: Allowing users to call the function directly was intentional. It allows users directly call it from other code, and easily test it.

python/cocoindex/op.py
```diff
@@ -49,7 +49,51 @@ class SourceSpec(metaclass=SpecMeta, category=OpCategory.SOURCE):  # pylint: dis
...
+        """Execute the function, returning a DataSlice."""
+
+        data_slice_args = [arg for arg in args if isinstance(arg, DataSlice)]
+        data_slice_kwargs = {
+            k: v for k, v in kwargs.items() if isinstance(v, DataSlice)
+        }
+        if not data_slice_args and not data_slice_kwargs:
+            raise ValueError(
+                "At least one DataSlice argument is required to provide flow context"
+            )
```
badmonster0: This seems a little bit restrictive. Ideally we should also all arguments are literal values.

I also have some little concern around the change around `@function` below (see my comment there).

I'm just thinking about another alternative approach, i.e. expose a new method for `FlowBuilder`. e.g.

```py
class FlowBuilder:
  ...
  def transform(self, fn_spec, *args, **kwargs):
    ...
```

Pros:
- No special case: Even if all inputs are not `DataSlice`, this API will work. It'll even work on no data input case (it's probably possible, just like a function can have no argument; the current `FlowBuilder` on the engine side doesn't support this, but changing API to this form will keep the door open for its support whenever deemed necessary).
- Reuse existing pattern: This is aligned with existing `add_source()` method for `FlowBuilder`.
- Keep `@function` to its current shape (users can still directly call it for testing, etc.)

Cons:
- Slightly more verbose: `flow_builder.transform(FunctionSpec(...), arg1=ds1, args=ds2)` vs `FunctionSpec(...)(arg1=ds1, arg2=ds2)`

I know we had discussion in the issue and we agreed on using direct `__call__()` for it. But after reading the implementation I started to realize more nuances here. Really sorry for the back and forth.

What do you think about this? I'm open for more discussions on this. Thanks!

python/cocoindex/op.py
```diff
@@ -49,7 +49,51 @@ class SourceSpec(metaclass=SpecMeta, category=OpCategory.SOURCE):  # pylint: dis
...
+            lambda target_scope, name: flow_builder_state.engine_flow_builder.transform(
+                _spec_kind(self),
+                dump_engine_object(self),
+                transform_args,
+                target_scope,
+                flow_builder_state.field_name_builder.build_name(
+                    name, prefix=_to_snake_case(_spec_kind(self)) + "_"
+                ),
+            ),
+        )
```
badmonster0: Now the `DataSlice.transform()` is more like a syntax sugar for this generic form. We can consider rewrite it to call this one directly, so less code to maintain.

python/cocoindex/op.py
lemorage: Really clever idea. I haven’t had a chance to dig deeply into it yet, but switching to work on `FlowBuilder` looks like helps sidestep lots of issues we had to deal with the previous direction. I haven’t spotted any concerns so far. I’ll take a closer look at the implementation details when I get a bit more time.

lemorage:
We opened the issue #654, because there was a limitation of the current chain-style API (`arg0.transform(...)`), which assumed a clear primary input. This breaks down in cases like `extract_from_llm(text=None, image=None)`, where inputs are optional and no single argument acts as “self”. Our proposed solution to this was to add a non-chain-style API.

Here we propose another solution, which is to add `transform` to `FlowBuilder`. So here is the diff in terms of usage:

```py
@cocoindex.flow_def(name="TextEmbedding")
def text_embedding_flow(
    flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope
) -> None:
    """
    Define an example flow that embeds text into a vector database.
    """
    data_scope["documents"] = flow_builder.add_source(
        cocoindex.sources.LocalFile(path="markdown_files")
    )

    doc_embeddings = data_scope.add_collector()

    with data_scope["documents"].row() as doc:
        # Chain-style
        doc["chunks"] = doc["content"].transform(
            cocoindex.functions.SplitRecursively(),
            language="markdown",
            chunk_size=2000,
            chunk_overlap=500,
        )
        
        # Non-chain-style
        func = cocoindex.functions.SplitRecursively()
        chunks = func(
            text=doc["content"],
            language="markdown",
            chunk_size=2000,
            chunk_overlap=500,
        )
        
        # FlowBuilder Transform
        doc["chunks"] = flow_builder.transform(
            cocoindex.functions.SplitRecursively(),
            doc["content"],
            language="markdown",
            chunk_size=2000,
            chunk_overlap=500,
        )
```

Compared to the **Non-chain-style approach**, one more con for **FlowBuilder Transform approach** I'd like to add here is, we could not get the intelligent editor hints for which param should be input when using the built-in CocoIndex functions, like `ExtractByLlm`, `SplitRecursively`, etc.

badmonster0 (approved):

badmonster0:
> Compared to the Non-chain-style approach, one more con for FlowBuilder Transform approach I'd like to add here is, we could not get the intelligent editor hints for which param should be input when using the built-in CocoIndex functions, like `ExtractByLlm`, `SplitRecursively`, etc.

Yes, noted. This is indeed a limitation.

I think there may be some tricks in Python's generic type annotations / meta programming that may be helpful to resolve this. I don't have time to think it carefully yet, but [typing.Concatenate](https://docs.python.org/3/library/typing.html#typing.Concatenate) may be related. With this, probably we may be able to define a placeholder method (or some sort of annotations) for function specs with parameter types, and annotate `transform()` in a way to correctly forward these type annotations?

Will merge this PR for now. We can resolve this separately.

PR #676, opened 2025-06-30
fix(ci): add numpy to dependencies and pass typing checks

lemorage:
This PR fixes the CI failure.

1. The `numpy` dep is missing, after making `sentence-transformers` optional since #674. Because `sentence-transformers` pulled in numpy as one of its dependencies, our code used numpy implicitly.
2. The `mypy` type checks failed, cuz imports won't be found, after `sentence-transformers` is optional.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
     "rich>=14.0.0",
     "python-dotenv>=1.1.0",
     "watchfiles>=1.1.0",
-    "numpy>=2.3.1",
+    "numpy>=1.23.2",
 ]
 license = "Apache-2.0"
 urls = { Homepage = "https://cocoindex.io/" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
     "rich>=14.0.0",
     "python-dotenv>=1.1.0",
     "watchfiles>=1.1.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "rich>=14.0.0",
     "python-dotenv>=1.1.0",
     "watchfiles>=1.1.0",
+    "numpy>=2.3.1",
```
badmonster0: probably `"numpy"` is sufficient here.

Seems numpy 1.x and 2.x are not compatible (https://numpy.org/doc/stable/numpy_2_0_migration_guide.html) and some older projects that not migrated yet may still need numpy 1.x and there will be a conflict. So relax the version restriction may be better here.

lemorage: That's a valid point. I agree with relaxing the version restriction here, but `numpy` seems too broad to me. I checked the numpy release history, and found 1.23.2 was the first release starting support for Python 3.11[^1]. `numpy <= 1.23.1` only worked for Python 3.8 - 3.10[^2]. Since our package denotes that we only support Python 3.11 ~ 3.13, I think we can change it like this.

[^1]: https://github.com/numpy/numpy/releases/tag/v1.23.2
[^2]: https://github.com/numpy/numpy/releases/tag/v1.23.1

PR #680, opened 2025-07-02
patient intake structured extraction example

badmonster0:


examples/patient_intake_extraction/main.py
```diff
@@ -0,0 +1,145 @@
...
+
+    patients_index.export(
+        "patients",
+        cocoindex.storages.Postgres(table_name="patients_info"),
+        primary_key_fields=["filename"],
+    )
+
+@cocoindex.main_fn()
+def _run():
+    pass
```
badmonster0: This is outdated and we don't need this now.

examples/patient_intake_extraction/.env
```diff
@@ -0,0 +1,4 @@
+# Postgres database address for cocoindex
+COCOINDEX_DATABASE_URL=postgres://cocoindex:cocoindex@localhost/cocoindex
+
+OPENAI_API_KEY=
```
badmonster0: We should probably keep `.env.example` and ignore `.env`, similar to https://github.com/cocoindex-io/cocoindex/tree/main/examples/amazon_s3_embedding. This file contains API key and too easy to be committed otherwise.

examples/patient_intake_extraction/README.md
```diff
@@ -0,0 +1,59 @@
...
+   ```bash
+   pip install -U cocoindex
+   ```
+
+3. Install MarkItDown
+   ```bash
+   pip install 'markitdown[all]'
+   ```
+
+4. Create a .env file from .env.example, and fill OPENAI_API_KEY.
```
badmonster0: Backquote `.env`, `.env.example` and `OPENAI_API_KEY`

Issue #659, opened 2025-06-24
[FEATURE] Support vLLM for text generation


badmonster0:
vLLM provides an [HTTP API](https://docs.vllm.ai/en/v0.8.3/serving/openai_compatible_server.html) compatible to OpenAI.
We want to support talking to it, as an additional [LLM API type](https://cocoindex.io/docs/ai/llm#llm-api-types).

Please also update the docs.

The task is similar to #627.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

par4m:
I can work on this!

PR #683, opened 2025-07-02
feat(llm): Add vLLM Support

par4m:
Implements https://github.com/cocoindex-io/cocoindex/issues/659

Since vLLM doesn't support MacOS, models can only be run on CPU.

Choose a model with less than 1.5B Parameters and at least 2400 token support, for testing DeepSeek Coder 1.3B is used on MacOS M4

Install vLLM: (python < 3.13)

```bash
pip install vllm
```

Run vLLM Server

```bash
vllm serve deepseek-ai/deepseek-coder-1.3b-instruct
```

```python
cocoindex.LlmSpec(
    api_type=cocoindex.LlmApiType.VLLM,
    model="deepseek-ai/deepseek-coder-1.3b-instruct",
    address="http://127.0.0.1:8000/v1",                    # /v1 is mandatory 
)
```

Also vLLM Supports Embedding the same way as OpenAI, a new PR for it would be better - https://docs.vllm.ai/en/v0.6.6/serving/openai_compatible_server.html#embeddings-api

More info - https://docs.vllm.ai/en/v0.6.6/serving/openai_compatible_server.html#

### DIFF
--- a/docs/docs/ai/llm.mdx
+++ b/docs/docs/ai/llm.mdx
@@ -26,6 +26,7 @@ We support the following types of LLM APIs:
 | [Voyage](#voyage) | `LlmApiType.VOYAGE` | ❌ | ✅ |
 | [LiteLLM](#litellm) | `LlmApiType.LITE_LLM` | ✅ | ❌ |
 | [OpenRouter](#openrouter) | `LlmApiType.OPEN_ROUTER` | ✅ | ❌ |
+| [vLLM](#vllm) | `LlmApiType.VLLM` | ✅ | ❌ |
 
 ## LLM Tasks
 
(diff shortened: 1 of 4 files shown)


docs/docs/ai/llm.mdx
badmonster0, line 1: Please also add it to the table at the beginning of the file:

https://github.com/cocoindex-io/cocoindex/blob/2ea160149adc650f4898e7d267f3710544701f3b/docs/docs/ai/llm.mdx?plain=1#L20-L28

par4m:
Not sure why formatting failed can you please rerun the checks

badmonster0:
> Not sure why formatting failed can you please rerun the checks

Should be caused by a format issue of some example code. fixed in https://github.com/cocoindex-io/cocoindex/pull/687.

The check still fails as it's not merged with the latest main yet, but it's safe to merge. I'll merge now.

badmonster0:
thank you @par4m ! new release note is out and we made a section for you, we love your contribution!!
https://cocoindex.io/blogs/cocoindex-changelog-2025-08-18#par4m ❤️

PR #723, opened 2025-07-09
feat(rs): add utilities to streamline tests for flow functions

### LINKED ISSUE #352: [FEATURE] Create a util to streamline test for flow functions in Rust
Our framework should have a reusable utils to easily test flow functions.

It takes the following input:
- a [`SimpleFunctionFactory`](https://github.com/cocoindex-io/cocoindex/blob/00f923ad515369b564bb24d2a464fc9234b57f2b/src/ops/interface.rs#L112)
- a spec
- input argument / types
- input row(s)

It returns output rows, and users can asserts on its output values.

lemorage:
Closes #352.

### DIFF
--- a/src/ops/functions/embed_text.rs
+++ b/src/ops/functions/embed_text.rs
@@ -100,22 +100,18 @@ pub fn register(registry: &mut ExecutorFactoryRegistry) -> Result<()> {
 mod tests {
     use super::*;
     use crate::ops::functions::test_utils::{build_arg_schema, test_flow_function};
-    use serde_json::json;
 
     #[tokio::test]
     #[ignore = "This test requires OpenAI API key or a configured local LLM and may make network calls."]
-    async fn test_embed_text_with_util() {
-        let context = Arc::new(FlowInstanceContext {
-            flow_instance_name: "test_embed_text_flow".to_string(),
-            auth_registry: Arc::new(AuthRegistry::default()),
-            py_exec_ctx: None,
-        });
-
+    async fn test_embed_text() {
         // Using OpenAI as an example.
-        let spec_json = json!({
-            "api_type": "OpenAi",
-            "model": "text-embedding-ada-002",
-        });
+        let spec = Spec {
+            api_type: LlmApiType::OpenAi,
+            model: "text-embedding-ada-002".to_string(),
+            address: None,
+            output_dimension: None,
+            task_type: None,
+        };
 
         let factory = Arc::new(Factory);
         let text_content = "CocoIndex is a performant data transformation framework for AI.";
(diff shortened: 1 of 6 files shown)


src/ops/functions/embed_text.rs
```diff
@@ -95,3 +95,74 @@ impl SimpleFunctionFactoryBase for Factory {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::ops::functions::test_utils::{build_arg_schema, test_flow_function};
+    use serde_json::json;
+
+    #[tokio::test]
+    #[ignore = "This test requires OpenAI API key or a configured local LLM and may make network calls."]
+    async fn test_embed_text_with_util() {
```
badmonster0: nit: I would just call it `test_embed_text`. "with util" sounds like its internal implementation.

src/ops/functions/test_utils.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// This function tests a flow function by providing a spec, input argument schemas, and values.
+pub async fn test_flow_function<S, R, F>(
+    factory: Arc<F>,
+    spec_json: JsonValue,
+    input_arg_schemas: Vec<OpArgSchema>,
+    input_arg_values: Vec<Value>,
+    context: Arc<FlowInstanceContext>,
+) -> Result<Value>
```
badmonster0, line 26: We can make this slightly simpler to use by:

- Take `S` instead of `JsonValue`. I think it's usually slightly easier to construct a struct than a JSON.
- Don't need to take `context` as input. I think we can build it internally.

src/ops/functions/test_utils.rs
```diff
@@ -0,0 +1,65 @@
...
+    name: Option<&str>,
+    value: Value,
+    value_type: EnrichedValueType,
+) -> OpArgSchema {
+    OpArgSchema {
+        name: name.map_or(crate::base::spec::OpArgName(None), |n| {
+            crate::base::spec::OpArgName(Some(n.to_string()))
+        }),
+        value_type,
+        analyzed_value: AnalyzedValueMapping::Constant { value },
```
badmonster0: We don't expect this to be `Constant` for most cases. This `analyzed_value` was only intended to fill in some attributes - it should never be directly consumed by a function to get the value. I think we can fill in something trivial e.g. use `AnalyzedValueMapping::Field(...)` and use the position of this argument to fill `fields_idx[0]`. To do this, we can consider making `test_flow_function` accepts a `Vec<(Option<&str>, EnrichedValueType)>` as `input_arg_schemas`, so we can construct each `AnalyzedValueMapping::Field` with a different index internally.

badmonster0 (approved):

PR #725, opened 2025-07-10
Try to implement notion backend

wykrrr:
This implementation has error:

[2025-07-09T05:32:58Z ERROR cocoindex_engine::execution::source_indexer] JoinError::Panic(Id(33), "Field is already set, violating single-definition rule: KTable({Range(RangeValue { start: 0, end: 5 }): ScopeValueBuilder { fields: [OnceLock(Basic(Str("test1"))), OnceLock(Struct(FieldValues { fields: [Basic(Int64(0)), Basic(Int64(1)), Basic(Int64(1))] })), OnceLock(Struct(FieldValues { fields: [Basic(Int64(5)), Basic(Int64(1)), Basic(Int64(6))] })), OnceLock(<uninit>)] }})", ...)

Please double check.

### DIFF
--- a/src/ops/sources/notion.rs
+++ b/src/ops/sources/notion.rs
@@ -1,0 +1,502 @@
+use crate::fields_value;
+use async_stream::try_stream;
+use reqwest::Client;
+use serde::Deserialize;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use crate::base::field_attrs;
+use crate::base::value::FieldValues;
+use crate::ops::sdk::*;
+
+#[derive(Debug, Deserialize)]
+pub struct Spec {
+    token: String,
+    source_type: String, // "database" or "page"
+    database_ids: Option<Vec<String>>,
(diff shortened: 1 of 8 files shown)


src/ops/sources/notion.rs
```diff
@@ -0,0 +1,502 @@
...
+        } else {
+            None
+        };
+
+        let value = if options.include_value {
+            let title = self.extract_title_from_properties(&page.properties);
+            let content = self.fetch_page_content(page_id.as_ref()).await?;
+
+            let fields = vec![
+                page.id.into(),
```
georgeh0, line 427: page ID is the key column, not a value column. This`get_value()` function takes key as input and return value columns. So page ID shouldn't be here.

The error is because fields put here doesn't match output schema returned by `get_output_schema()` below (for a KTable, first is key column, the remaining are value columns, so the value here is expected to match the remaining columns).

We will need to have validation to check schema of values returned here matching the declared schema (created https://github.com/cocoindex-io/cocoindex/issues/735 for this), so such problem can be easily caught and understandable in the future.

PR #727, opened 2025-07-10
feat(convert): enhance dict and struct encoding with type awareness

### LINKED ISSUE #670: [FEATURE] Allow CocoIndex Struct binding to `dict` in Python
Currently we allow CocoIndex struct to bind to dataclasses and namedtuples ([doc](https://cocoindex.io/docs/core/data_types#struct-types)). We want to support binding struct to `dict` (or `dict[str, Any]`) too - this is more handy as users don't need to define a type first.

`dict` cannot be used when a *specific type* is required, e.g. for return type as custom functions ([existing doc](https://cocoindex.io/docs/core/data_types#struct-types:~:text=use%20the%20specific%20type)).

Users are even allowed to use `Any` or omit type annotation to bind Struct values when a specific type is not required (e.g. for custom function arguments), just like for basic typed values type annotation is not required for function arguments.

---

vumichien:
Resolve #670

### DIFF
--- a/python/cocoindex/convert.py
+++ b/python/cocoindex/convert.py
@@ -86,6 +86,14 @@ def make_engine_value_decoder(
         or dst_annotation is inspect.Parameter.empty
         or dst_annotation is Any
     )
+    # Handle struct -> dict binding for explicit dict annotations
+    is_dict_annotation = False
+    if dst_annotation is dict:
+        is_dict_annotation = True
+    elif getattr(dst_annotation, "__origin__", None) is dict:
+        args = getattr(dst_annotation, "__args__", ())
+        if args == (str, Any):
+            is_dict_annotation = True
     if dst_is_any:
         if src_type_kind == "Union":
             return lambda value: value[1]
@@ -97,20 +105,10 @@ def make_engine_value_decoder(
                 f"It's required for {src_type_kind} type."
             )
         return lambda value: value
+    if is_dict_annotation and src_type_kind == "Struct":
+        return _make_engine_struct_to_dict_decoder(field_path, src_type["fields"])
 
     dst_type_info = analyze_type_info(dst_annotation)
-
-    # Handle struct -> dict binding for explicit dict annotations
-    if (
-        src_type_kind == "Struct"
-        and dst_type_info.kind == "KTable"
-        and dst_type_info.elem_type
-        and isinstance(dst_type_info.elem_type, tuple)
-        and len(dst_type_info.elem_type) == 2
-        and dst_type_info.elem_type[0] is str
-        and dst_type_info.elem_type[1] is Any
-    ):
-        return _make_engine_struct_to_dict_decoder(field_path, src_type["fields"])
 
     if src_type_kind == "Union":
         dst_type_variants = (
@@ -311,7 +309,7 @@ def _make_engine_struct_value_decoder(
 def _make_engine_struct_to_dict_decoder(
     field_path: list[str],
     src_fields: list[dict[str, Any]],
-) -> Callable[[list[Any]], dict[str, Any]]:
+) -> Callable[[list[Any] | None], dict[str, Any] | None]:
     """Make a decoder from engine field values to a Python dict."""
 
     field_decoders = []
@@ -326,12 +324,17 @@ def _make_engine_struct_to_dict_decoder(
         field_path.pop()
         field_decoders.append((field_name, field_decoder))
 
-    def decode_to_dict(values: list[Any]) -> dict[str, Any]:
-        result = {}
-        for i, (field_name, field_decoder) in enumerate(field_decoders):
-            if i < len(values):
-                result[field_name] = field_decoder(values[i])
-        return result
+    def decode_to_dict(values: list[Any] | None) -> dict[str, Any] | None:
+        if values is None:
+            return None
+        if len(field_decoders) != len(values):
+            raise ValueError(
+                f"Field count mismatch: expected {len(field_decoders)}, got {len(values)}"
+            )
+        return {
+            field_name: field_decoder(value)
+            for value, (field_name, field_decoder) in zip(values, field_decoders)
+        }
 
     return decode_to_dict
 
--- a/python/cocoindex/convert.py
+++ b/python/cocoindex/convert.py
@@ -86,14 +86,6 @@ def make_engine_value_decoder(
         or dst_annotation is inspect.Parameter.empty
         or dst_annotation is Any
     )
-    # Handle struct -> dict binding for explicit dict annotations
-    is_dict_annotation = False
-    if dst_annotation is dict:
-        is_dict_annotation = True
-    elif getattr(dst_annotation, "__origin__", None) is dict:
-        args = getattr(dst_annotation, "__args__", ())
-        if args == (str, Any):
-            is_dict_annotation = True
     if dst_is_any:
         if src_type_kind == "Union":
             return lambda value: value[1]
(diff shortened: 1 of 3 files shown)


python/cocoindex/convert.py
```diff
@@ -22,6 +22,52 @@
...
 
+def _is_ktable_dict(value: dict[Any, Any]) -> bool:
+    """Check if a dict is being used as a KTable (dict[key, struct])."""
+    if not value:
+        return False
+    first_val = next(iter(value.values()))
+    return is_struct_type(type(first_val))
+
+
+def encode_engine_value_with_type(value: Any, target_type: Any = None) -> Any:
```
badmonster0: This function is not called by any other functions (other than itself). How are we going to use it?

Note that for conversions between Python values and engine values, there're two directions:

- decode (engine -> Python), e.g. for a custom function argument we need to do this. Engine already has the type information, so users don't need to annotate with the exact type for this case. This is also the original intention of #670 

- encode (Python -> engine), e.g. for the return type of the custom function. In this case, the custom function's annotation decides its return type - it provides the ground truth for the type. So here there's no situation like users provides a dict but we convert it to a struct. 

So I doubt if there's a real scenario that we want to encode a `dict` as a Struct?

python/cocoindex/tests/test_convert.py
```diff
@@ -1229,3 +1229,113 @@ class MixedStruct:
...
+        {"name": "city", "type": {"kind": "Str"}},
+    ]
+
+    # Test decoding struct to dict
+    decoder = _make_engine_struct_value_decoder([], src_fields, Dict[str, Any])
+    result = decoder(["Alice", 30, "New York"])
+    assert result == {"name": "Alice", "age": 30, "city": "New York"}
+
+    # Test decoding struct to untyped dict
+    decoder = _make_engine_struct_value_decoder([], src_fields, dict)
```
badmonster0: I think in most cases we should focus on testing public interfaces instead of private ones (related discussions: [[1]](https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice?utm_source=chatgpt.com) [[2]](https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones/47401015)).

For encoding/decoding logic, the util `validate_full_roundtrip()` has very a good end-to-end coverage and we can use it for most cases. For example, we can do something like:

```py
validate_full_roundtrip(
  struct_value, SomeNamedTupleType,
  (some_dict_value, dict),
  (some_dict_value, Any))
``` 

Then it can validate the specific struct value can be decoded under both `dict` and `Any` annotations.

(and also feel free to reuse existing tests and add new arguments like `(some_dict_value, dict)` to them)

python/cocoindex/convert.py
```diff
@@ -258,6 +298,19 @@ def _make_engine_struct_value_decoder(
             )
             for name in fields
         }
+    elif dst_struct_type is dict or (
+        hasattr(dst_struct_type, "__origin__")
+        and getattr(dst_struct_type, "__origin__") is dict
```
badmonster0: This expression to check if it's a `dict` appears twice. We may keep this into a local variable.

badmonster0: Note that we also would like to support the case that users write no type annotation or use `Any` as type annotation (just like they're allowed for basic types in custom function parameters).

badmonster0 (commented):
Thanks for working on this!

vumichien:
Thank you for your instructions. I just revised the logic and test cases

python/cocoindex/convert.py
```diff
@@ -294,6 +308,34 @@ def make_closure_for_value(
...
+        field_path.append(f".{field_name}")
+        field_decoder = make_engine_value_decoder(
+            field_path,
+            field_schema["type"],
+            Any,  # Use Any for recursive decoding
+        )
+        field_path.pop()
+        field_decoders.append((field_name, field_decoder))
+
+    def decode_to_dict(values: list[Any]) -> dict[str, Any]:
```
badmonster0: `values` should be `list[Any] | None`, and we want to check and return `None` when it's `None`.

python/cocoindex/convert.py
```diff
@@ -294,6 +308,34 @@ def make_closure_for_value(
...
+        result = {}
+        for i, (field_name, field_decoder) in enumerate(field_decoders):
+            if i < len(values):
```
badmonster0: We may check `len(field_decoders) == len(values)` and raise an exception before the loop.

For the loop, using `zip(values, field_decoders)` is slightly simpler. (also consider a dict comprehension)

python/cocoindex/convert.py
```diff
@@ -98,6 +100,18 @@ def make_engine_value_decoder(
 
     dst_type_info = analyze_type_info(dst_annotation)
 
+    # Handle struct -> dict binding for explicit dict annotations
+    if (
+        src_type_kind == "Struct"
+        and dst_type_info.kind == "KTable"
```
badmonster0: After we support binding `struct` to `dict`, this case is no longer a _KTable_.

I think here we may directly check if `dst_annotation` is one of `dict` and `dict[str, Any]`. We can check it before calling `analyze_type_info()`. Note that this is a case similar to `dst_is_any` case, which is also handled without `dst_type_info`.

python/cocoindex/typing.py
```diff
@@ -241,7 +241,10 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
 
     elif base_type is collections.abc.Mapping or base_type is dict:
         args = typing.get_args(t)
-        elem_type = (args[0], args[1])
+        if len(args) == 0:  # Handle untyped dict
+            elem_type = (str, Any)
+        else:
+            elem_type = (args[0], args[1])
```
badmonster0, line 250: The untyped dict should not be treated as a _KTable_.

Actually the current `analyze_type_info()` is intended to extract specific CocoIndex type information out of the type annotation. For example, if the annotation is `Any` or empty, it raises an exception today.
(this is not clear from the docstring of this function, you may consider helping add this as you're on this file, thanks!)

So if it's an untyped `dict`, I think we may raise an exception too.

python/cocoindex/typing.py
```diff
@@ -282,6 +285,9 @@ def analyze_type_info(t: Any) -> AnalyzedTypeInfo:
             kind = "OffsetDateTime"
         elif t is datetime.timedelta:
             kind = "TimeDelta"
+        elif t is dict:
+            elem_type = (str, Any)
+            kind = "KTable"
```
badmonster0: Same as I commented above, we may raise an exception in this case instead of treating it as a _KTable_. Also consider merging the check with the check above for untyped `dict`.

python/cocoindex/convert.py
vumichien: revised

python/cocoindex/convert.py
vumichien: revised

python/cocoindex/typing.py
vumichien: revised

python/cocoindex/convert.py
vumichien: revised

python/cocoindex/convert.py
```diff
@@ -86,15 +86,27 @@ def make_engine_value_decoder(
...
         or dst_annotation is Any
     )
+    # Handle struct -> dict binding for explicit dict annotations
+    is_dict_annotation = False
+    if dst_annotation is dict:
+        is_dict_annotation = True
+    elif getattr(dst_annotation, "__origin__", None) is dict:
+        args = getattr(dst_annotation, "__args__", ())
+        if args == (str, Any):
+            is_dict_annotation = True
```
badmonster0: It'll be slightly better by moving these below, right above `if` statements using `is_dict_annotation`, as they're closer related.

badmonster0 (approved):

PR #736, opened 2025-07-12
add Azure Blob Storage source with full authentication support

wykrrr:
- Implement Azure Blob Storage source following S3 pattern
  - Support multiple authentication methods with priority:
    * Connection string (highest priority)
    * SAS token (with proper permissions validation)
    * Account key (full access)
    * Anonymous access (public containers)
  - Include file pattern filtering (include/exclude glob patterns)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "RustyXML"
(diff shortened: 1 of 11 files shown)


badmonster0:
Thanks! We'll merge into a branch and make further changes before merging into main.

PR #737, opened 2025-07-12
test: add unit tests to execute transform flow e2e

### LINKED ISSUE #618: Add test on Python side to execute transform flow end to end

lemorage:
resolve #618.

### DIFF
--- a/python/cocoindex/tests/test_transform_flow.py
+++ b/python/cocoindex/tests/test_transform_flow.py
@@ -36,17 +36,19 @@ def simple_transform(text: cocoindex.DataSlice[str]) -> cocoindex.DataSlice[str]
 
 
 @cocoindex.op.function()
-def extract_child_values(parent: Parent) -> list[int]:
-    """Extract values from each child in the Parent's children list."""
-    return [child.value for child in parent.children]
+def extract_value(child: int) -> int:
+    """Extracts the value from a Child object."""
+    return child
 
 
 @cocoindex.transform_flow()
 def for_each_transform(
     data: cocoindex.DataSlice[Parent],
-) -> cocoindex.DataSlice[list[int]]:
+) -> cocoindex.DataSlice[Parent]:
     """Transform flow that processes child rows to extract values."""
-    return data.transform(extract_child_values)
+    with data["children"].row() as child:
+        child["new_field"] = child["value"].transform(extract_value)
+    return data
 
 
 def test_simple_transform_flow() -> None:
@@ -71,12 +73,12 @@ def test_for_each_transform_flow() -> None:
     """Test the complex transform flow with child rows."""
     input_data = Parent(children=[Child(1), Child(2), Child(3)])
     result = for_each_transform.eval(input_data)
-    expected = [1, 2, 3]
+    expected = Parent(children=[Child(1), Child(2), Child(3)])
     assert result == expected, f"Expected {expected}, got {result}"
 
     input_data = Parent(children=[])
     result = for_each_transform.eval(input_data)
-    assert result == [], f"Expected [], got {result}"
+    assert result == Parent(children=[]), f"Expected [], got {result}"
 
 
 @pytest.mark.asyncio
@@ -84,5 +86,5 @@ async def test_for_each_transform_flow_async() -> None:
     """Test the complex transform flow asynchronously."""
     input_data = Parent(children=[Child(4), Child(5)])
     result = await for_each_transform.eval_async(input_data)
-    expected = [4, 5]
+    expected = Parent(children=[Child(4), Child(5)])
     assert result == expected, f"Expected {expected}, got {result}"
--- a/python/cocoindex/tests/test_transform_flow.py
+++ b/python/cocoindex/tests/test_transform_flow.py
@@ -1,0 +1,88 @@
+import typing
+from dataclasses import dataclass
+
+import pytest
+
+import cocoindex
+
+
+@dataclass
+class Child:
+    value: int
+
+
+@dataclass
+class Parent:
+    children: list[Child]
+
+
+# Fixture to initialize CocoIndex library
+@pytest.fixture(scope="session", autouse=True)
+def init_cocoindex() -> typing.Generator[None, None, None]:
+    cocoindex.init()
+    yield
+
+
+@cocoindex.op.function()
+def add_suffix(text: str) -> str:
+    """Append ' world' to the input text."""
+    return f"{text} world"
+
+
+@cocoindex.transform_flow()
+def simple_transform(text: cocoindex.DataSlice[str]) -> cocoindex.DataSlice[str]:
+    """Transform flow that applies add_suffix to input text."""
+    return text.transform(add_suffix)
+
+
+@cocoindex.op.function()
+def extract_child_values(parent: Parent) -> list[int]:
+    """Extract values from each child in the Parent's children list."""
+    return [child.value for child in parent.children]
+
+
+@cocoindex.transform_flow()
+def for_each_transform(
+    data: cocoindex.DataSlice[Parent],
+) -> cocoindex.DataSlice[list[int]]:
+    """Transform flow that processes child rows to extract values."""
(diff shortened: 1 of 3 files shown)


python/cocoindex/tests/test_transform_flow.py
```diff
@@ -0,0 +1,88 @@
...
+    """Extract values from each child in the Parent's children list."""
+    return [child.value for child in parent.children]
+
+
+@cocoindex.transform_flow()
+def for_each_transform(
+    data: cocoindex.DataSlice[Parent],
+) -> cocoindex.DataSlice[list[int]]:
+    """Transform flow that processes child rows to extract values."""
+    return data.transform(extract_child_values)
```
badmonster0: To test the "for each", I think we want to have a "for each" within this flow like

```py
with data["children"].row() as child:
  ...
```
 instead of doing the iteration within the function's internal implementation.

lemorage: How should we collect the child values in the flow if we change it like this (if not get it wrong):
```py
@cocoindex.transform_flow()
def for_each_transform(data: cocoindex.DataSlice[Parent]) -> cocoindex.DataSlice[list[int]]:
    with data["children"].row() as child:
        child_values = child["value"].transform(extract_value)
```

badmonster0: We don't need to collect it ("collect" is only for "export" now, and transform flow doesn't do export). What's in my mind is return the entire table, e.g.

```py
...
chunks = data.transform(...)
with chunks.row() as chunk:
  chunk["new_field"] = chunk["field"].transform(...)
return chunks
```

(Actually we didn't do similar things within a transform flow before; there's a chance that it doesn't work, if that's the case, we can comment this test out for now, and resolve this as a separate issue)

lemorage: Thanks for clarifying, I see!

python/cocoindex/tests/test_transform_flow.py
```diff
@@ -0,0 +1,90 @@
...
+    input_text = "async"
+    result = await simple_transform.eval_async(input_text)
+    assert result == "async world", f"Expected 'async world', got {result}"
+
+
+def test_for_each_transform_flow() -> None:
+    """Test the complex transform flow with child rows."""
+    input_data = Parent(children=[Child(1), Child(2), Child(3)])
+    result = for_each_transform.eval(input_data)
+    expected = Parent(children=[Child(1), Child(2), Child(3)])
```
badmonster0: This is not the value we really expect it to return - the new field isn't here.
The reason is the annotated return type doesn't contain the new field.

(This will need to define another set of `Parent` + `Child` for this additional field. This may look a little bit awkward for now, but it's OK as it's only a test, and after #758 is resolved we won't need to define types to annotate return values for transform functions)

lemorage: Yeah right, that can be a valid workaround. But aside from that, I would say there is one more concern if I were right.

Unlike regular `@flow_def` flows that use `DataCollector` to explicitly capture the output, here our transform flows can only rely on the direct return value. The `row()` context creates a temporary scope for iteration, but changes within that scope aren't automatically merged back.

So, if we update it as follows, I don't think the `new_field` is attached to the data.

```py
@dataclass
class NewChild:
    value: int
    new_field: int = 0

@dataclass
class NewParent:
    children: list[NewChild]

@cocoindex.op.function()
def extract_value(value: int) -> int:
    return value

@cocoindex.transform_flow()
def for_each_transform(
    data: cocoindex.DataSlice[Parent],
) -> cocoindex.DataSlice[NewParent]:
    with data["children"].row() as child:
        child["new_field"] = child["value"].transform(extract_value)
    return data

@pytest.mark.asyncio
async def test_for_each_transform_flow_async() -> None:
    input_data = Parent(children=[Child(4), Child(5)])
    result = await for_each_transform.eval_async(input_data)
    expected = NewParent(
        children=[NewChild(value=4, new_field=4), NewChild(value=5, new_field=5)]
    )
    assert result == expected, f"Expected {expected}, got {result}"
```

badmonster0: > The `row()` context creates a temporary scope for iteration

This is not the intention of the design. In `with data["children"].row() as child`, think `child` as a reference. When doing `child["new_field"] = ...`, it should add a new field to the original table. CocoInsight also shows this idea (it presents the entire table with all new fields merged).

Did you run the code with this change? If the returned table indeed doesn't have `new_field`, it's a bug. Maybe comment out this test for now and merge, and after we fix the bug we can uncomment it.

lemorage: Thanks for the explanation! That's quite unusual; the `new_field` in the result from the above code snippet was always 0. I'd like to make some other minimal examples and delve into this deeper to see what's wrong with it.

lemorage:
> and after https://github.com/cocoindex-io/cocoindex/issues/758 is resolved we won't need to define types to annotate return values for transform functions

Sorry for keeping this for so long. Said here, does that mean now we can write the code without defining NewParent here, after #758?
```py
@cocoindex.transform_flow()
def for_each_transform(
    data: cocoindex.DataSlice[Parent],
) -> cocoindex.DataSlice[typing.Any]: # No need to specify cocoindex.DataSlice[NewParent]?
    with data["children"].row() as child:
        child["new_field"] = child["value"].transform(extract_value)
    return data
    ```

badmonster0:
> > and after #758 is resolved we won't need to define types to annotate return values for transform functions
> 
> Sorry for keeping this for so long. Said here, does that mean now we can write the code without defining NewParent here, after #758?
> 
> ```python
> @cocoindex.transform_flow()
> def for_each_transform(
>     data: cocoindex.DataSlice[Parent],
> ) -> cocoindex.DataSlice[typing.Any]: # No need to specify cocoindex.DataSlice[NewParent]?
>     with data["children"].row() as child:
>         child["new_field"] = child["value"].transform(extract_value)
>     return data
>     ```
> ```

No worries. Yes, that's true. We can rewrite the code without defining NewParent here.

Besides, because of #818, it's still possible that even if returning a `dict` it still doesn't work as intended (I'll assess #818 more closely later when got time). So if it still doesn't work after a quick try, feel free to comment out the related test first and we can merge it (it means the test is already helping us identifying issues!). Thanks!

georgeh0:
> > > and after #758 is resolved we won't need to define types to annotate return values for transform functions
> > 
> > 
> > Sorry for keeping this for so long. Said here, does that mean now we can write the code without defining NewParent here, after #758?
> > ```python
> > @cocoindex.transform_flow()
> > def for_each_transform(
> >     data: cocoindex.DataSlice[Parent],
> > ) -> cocoindex.DataSlice[typing.Any]: # No need to specify cocoindex.DataSlice[NewParent]?
> >     with data["children"].row() as child:
> >         child["new_field"] = child["value"].transform(extract_value)
> >     return data
> >     ```
> > ```
> 
> No worries. Yes, that's true. We can rewrite the code without defining NewParent here.
> 
> Besides, because of #818, it's still possible that even if returning a `dict` it still doesn't work as intended (I'll assess #818 more closely later when got time). So if it still doesn't work after a quick try, feel free to comment out the related test first and we can merge it (it means the test is already helping us identifying issues!). Thanks!

#818 has been fixed in #825 :)

badmonster0:
Perfect! Thanks a lot for adding the tests!

PR #756, opened 2025-07-15
update readme with translations and badge

cocoindex-dev:


cocoindex-dev:
preview here https://github.com/cocoindex-dev/cocoindex/blob/patch-6/README.md

PR #767, opened 2025-07-17
feat(conversion): add LTable and KTable decoders for list and dict bindings

### LINKED ISSUE #758: [FEATURE] Allow CocoIndex Table bind to `dict` / `list` in Python
This is a continuation of #670, which supports binding _Struct_ to `dict`, so users don't need to explicitly define a dataclass, and can even omit type annotations for Engine->Python value binding (e.g. argument of custom functions)

We also want to support it for _Table_ types, in particular:
- For _LTable_, supports binding to `list[dict[String, Any]]`
- For _KTable_, supports binding to `dict[K, dict[String, Any]]` (`K` is key type; use a tuple if multiple key parts)
(it's the runtime type we support, users don't have to annotate)

With this, we'll support omit type annotations for Engine->Python value binding for all data types.

---

vumichien:
Resolved #758

### DIFF
--- a/python/cocoindex/convert.py
+++ b/python/cocoindex/convert.py
@@ -357,10 +357,13 @@ def _make_engine_ltable_to_list_dict_decoder(
         if values is None:
             return None
         result = []
-        for row_values in values:
+        for i, row_values in enumerate(values):
             decoded_row = row_decoder(row_values)
-            if decoded_row is not None:
-                result.append(decoded_row)
+            if decoded_row is None:
+                raise ValueError(
+                    f"LTable row at index {i} decoded to None, which is not allowed."
+                )
+            result.append(decoded_row)
         return result
 
     return decode_to_list_dict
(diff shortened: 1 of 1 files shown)


python/cocoindex/convert.py
```diff
@@ -340,6 +344,72 @@ def decode_to_dict(values: list[Any] | None) -> dict[str, Any] | None:
...
+    # Create a decoder for each row (struct) to dict
+    row_decoder = _make_engine_struct_to_dict_decoder(field_path, src_fields)
+
+    def decode_to_list_dict(values: list[Any] | None) -> list[dict[str, Any]] | None:
+        if values is None:
+            return None
+        result = []
+        for row_values in values:
+            decoded_row = row_decoder(row_values)
+            if decoded_row is not None:
```
badmonster0: I think we may raise an exception if a row is `None`

python/cocoindex/convert.py
```diff
@@ -340,6 +344,72 @@ def decode_to_dict(values: list[Any] | None) -> dict[str, Any] | None:
...
+    value_decoder = _make_engine_struct_to_dict_decoder(field_path, value_fields_schema)
+
+    def decode_to_dict_dict(
+        values: list[Any] | None,
+    ) -> dict[Any, dict[str, Any]] | None:
+        if values is None:
+            return None
+        result = {}
+        for row_values in values:
+            if len(row_values) < 2:
```
badmonster0: actually a key-only KTable is valid. In this case the value can be an empty dict.

python/cocoindex/convert.py
vumichien: I have added a way to raise an exception in here

python/cocoindex/convert.py
vumichien: Revised: if the row has only one element (the key), then the value part should be an empty dict.

badmonster0 (approved):

badmonster0:
thank you @vumichien ! new release note is out and we made a section for you, we love your contribution!!
https://cocoindex.io/blogs/cocoindex-changelog-2025-08-18#vumichien ❤️

PR #783, opened 2025-07-20
Add naming validation for flow identifiers (#779)

### LINKED ISSUE #779: [FEATURE] Enforce restrictions for flow names, field names, target names, app namespace names
We want to enforce restrictions on flow names, field names, target names, app namespace names.

- Field names, target names: allow letters, digits and underscores, starting with letter/underscore, no more than 64 characters
- Full flow names (app namespace + `.` + flow name), app namespace name: allow letters, digits, underscores and `.`, starting with letter/underscore, no more than 64 characters
- Flow name: allow letters, digits and underscores, starting with letter/underscore, no more than 64 characters

None should start from double underscores (reserved for internal usage)

Where to enforce?
- All of them should be enforced in Python SDK which provides more user friendly error message.
- Field names, target names and full flow names should also be enforced in the core engine as an extra guard.

---

theparthgupta:
## Overview
Implements naming validation for CocoIndex flow identifiers as requested in #779.

## Changes
- New validation module with comprehensive naming rules
- Integration with flow definition decorator
- Public API exports for external validation
- Full test coverage (11 test cases)

## Validation Rules
- Flow/field/target names: `[a-zA-Z_][a-zA-Z0-9_]*`
- Maximum 64 characters
- No double-underscore prefixes

## Testing
All existing tests pass. No breaking changes.

Fixes #779

### DIFF
--- a/python/cocoindex/__init__.py
+++ b/python/cocoindex/__init__.py
@@ -5,6 +5,12 @@ Cocoindex is a framework for building and running indexing pipelines.
 from . import functions, sources, targets, cli, utils
 
 from . import targets as storages  # Deprecated: Use targets instead
+from .validation import (
+    NamingError,
+    validate_field_name,
+    validate_flow_name,
+    validate_target_name,
+)
 
 from .auth_registry import AuthEntryReference, add_auth_entry, add_transient_auth_entry
 from .flow import FlowBuilder, DataScope, DataSlice, Flow, transform_flow
@@ -61,6 +67,11 @@ __all__ = [
     "update_all_flows_async",
     "setup_all_flows",
     "drop_all_flows",
+    # Validation
+    "NamingError",
+    "validate_field_name",
+    "validate_flow_name",
+    "validate_target_name",
     # Lib
     "init",
     "start_server",
--- a/python/cocoindex/flow.py
+++ b/python/cocoindex/flow.py
@@ -10,6 +10,8 @@ import functools
 import inspect
 import re
 
+from .validation import validate_flow_name, NamingError
+
 from dataclasses import dataclass
 from enum import Enum
 from threading import Lock
@@ -832,10 +834,11 @@ def get_flow_full_name(name: str) -> str:
 
 def add_flow_def(name: str, fl_def: Callable[[FlowBuilder, DataScope], None]) -> Flow:
     """Add a flow definition to the cocoindex library."""
-    if not all(c.isalnum() or c == "_" for c in name):
-        raise ValueError(
-            f"Flow name '{name}' contains invalid characters. Only alphanumeric characters and underscores are allowed."
-        )
+    try:
+        validate_flow_name(name)
+    except NamingError as e:
+        raise ValueError(str(e)) from e
+
     with _flows_lock:
         if name in _flows:
             raise KeyError(f"Flow with name {name} already exists")
--- a/python/cocoindex/validation.py
+++ b/python/cocoindex/validation.py
@@ -1,0 +1,101 @@
+"""
+Naming validation for CocoIndex identifiers.
+
+This module enforces naming conventions for flow names, field names,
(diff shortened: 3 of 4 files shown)


python/cocoindex/__init__.py
```diff
@@ -61,6 +67,11 @@
     "update_all_flows_async",
     "setup_all_flows",
     "drop_all_flows",
+    # Validation
+    "NamingError",
+    "validate_field_name",
+    "validate_flow_name",
+    "validate_target_name",
```
badmonster0: Functions exported from this file are APIs exposed to cocoindex users. I think these functions are internal ones and shouldn't be exposed to users.

Since the validations are intended to happen internally, we should also update code to call these functions (only `validate_flow_name` is called now, others are not)

python/cocoindex/flow.py
```diff
@@ -832,10 +834,11 @@ def get_flow_full_name(name: str) -> str:
...
 def add_flow_def(name: str, fl_def: Callable[[FlowBuilder, DataScope], None]) -> Flow:
     """Add a flow definition to the cocoindex library."""
-    if not all(c.isalnum() or c == "_" for c in name):
-        raise ValueError(
-            f"Flow name '{name}' contains invalid characters. Only alphanumeric characters and underscores are allowed."
-        )
+    try:
+        validate_flow_name(name)
+    except NamingError as e:
+        raise ValueError(str(e)) from e
```
badmonster0: This re-raise seems not really necessary. I think we don't need to wrap a `NamingError` by a `ValueError`.

python/cocoindex/validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if name.startswith("__"):
+        return f"{identifier_type} name '{name}' cannot start with double underscores (reserved for internal usage)"
+
+    # Define allowed pattern
+    if allow_dots:
+        pattern = r"^[a-zA-Z_][a-zA-Z0-9_.]*$"
+        allowed_chars = "letters, digits, underscores, and dots"
+    else:
+        pattern = r"^[a-zA-Z_][a-zA-Z0-9_]*$"
```
badmonster0: We only have two fixed regular expression patterns. I think we can consider pre-compile them (in the global scope): https://docs.python.org/3/library/re.html#re.compile

theparthgupta:
please review if the changes are correct now or any other thing is needed to be changed

python/cocoindex/flow.py
```diff
@@ -381,6 +386,11 @@ def export(
 
         `vector_index` is for backward compatibility only. Please use `vector_indexes` instead.
         """
+        from .validation import (
+            validate_target_name,
+        )  # Import locally to avoid circular imports
```
badmonster0: Curious why do need to do local import? The `validation` module doesn't import any other module from cocoindex.

python/cocoindex/flow.py
badmonster0, line 1: `validate_app_namespace_name` and `validate_full_flow_name` are still not used.

python/cocoindex/flow.py
theparthgupta: Right! since the validation module doesn't import from cocoindex, I can move these to the top-level imports. I'll clean that up.

python/cocoindex/flow.py
theparthgupta: its done

python/cocoindex/flow.py
```diff
@@ -660,6 +672,7 @@ class Flow:
     def __init__(
         self, name: str, full_name: str, engine_flow_creator: Callable[[], _engine.Flow]
     ):
+        validate_full_flow_name(full_name)
```
badmonster0, line 676: Probably we may move the check of `validate_flow_name(name)` to here too, so similar checks is triggered in the location, which makes code easier to maintain.

georgeh0 (approved):

PR #792, opened 2025-07-23
added default values for field decoding (#788)

### LINKED ISSUE #788: [FEATURE] Enhance field decoding by adding type-based defaults
**What is the use case?**
Specifically, this refers to the inner function `make_closure_for_value`. In L298~L301, it raises an error if the field has no default value. As an improvement, this could be extended to support auto-defaulting based on common types behind a feature flag.

https://github.com/cocoindex-io/cocoindex/blob/9491acf7bd90a737056f33cc1bde045c218d388a/python/cocoindex/convert.py#L281-L307

Pros:
 - **User-Friendly**: Prevents crashes for missing fields, allowing the program to continue with sensible defaults.
 - **Reduces Boilerplate**: Users don’t need to specify defaults for every field.

Cons:
- **Hides Errors**: Auto-defaulting might mask potential bugs.

Instead of raising a ValueError, issue a warning and assign a default value based on the data type (e.g., 0 for `int`, 0.0 for `float`, [] for `list`, etc.).

**Additional context**

---

kingkushal16:
fixed #788 

## Overview

Implements robust auto-defaulting for missing fields in the CocoIndex engine value decoder, improving error handling and data conversion flexibility. This addresses **[#788](https://github.com/your-org/your-repo/issues/788)**.

---

## Changes

- Auto-defaulting is **always enabled** for supported types (no user-facing option).
- **Type-aware defaults**:
  - Only nullable types and table structures (`LTable` / `KTable`) are auto-defaulted.
- **Clear error messages**: When auto-defaulting is not supported, errors include field names and type information.
- **Warning system**: Users are notified via warnings when auto-defaults are applied.
- **Test coverage**: New tests ensure correct auto-defaulting and error reporting.

---

### Auto-Default Rules

| Type                        | Default Value     | Notes                                  |
|----------------------------|-------------------|----------------------------------------|
| `Optional[T]` / `T \| None`| `None`            | Nullable types                         |
| `LTable`                   | `[]`              | Defaults to an empty list              |
| `KTable`                   | `{}`              | Defaults to an empty dictionary        |
| All other types            | ❌ No default      | Missing fields raise an error          |

---

### Enhanced Error Handling

- **Field-specific** error messages: Include field name and expected type.
- **Warnings** are issued when auto-defaults are applied.

---

## Implementation Details

- Introduced `_get_auto_default_for_type()` helper:
  - Returns the default value.
  - Indicates whether auto-defaulting is supported.
- All decoder logic updated to use this helper.
- Removed `auto_default_missing_fields` parameter.
  - Auto-defaulting is **always on** for supported types and **off** for others.

---

## Files Modified

- `python/cocoindex/convert.py`: Core implementation and error/warning improvements
- `python/cocoindex/tests/test_convert.py`: New tests for auto-defaulting and error reporting


python/cocoindex/convert.py
```diff
@@ -249,10 +271,44 @@ def decode(value: Any) -> Any | None:
...
+    """
+    if annotation is None or annotation is inspect.Parameter.empty or annotation is Any:
+        return None
+
+    try:
+        type_info = analyze_type_info(annotation)
+
+        # Case 1: Nullable types (Optional[T] or T | None)
+        if type_info.nullable:
+            return None
```
badmonster0: Now, in both `T | None` case, and for cases where default value is unsupported, we're returning the same thing `None`. We should distinguish them. Consider return an additional `bool`, e.g. we may return `tuple[Any, bool]` and using the `bool` value to indicate if auto default is supported.

python/cocoindex/convert.py
```diff
@@ -67,6 +68,7 @@ def make_engine_value_decoder(
     field_path: list[str],
     src_type: dict[str, Any],
     dst_annotation: Any,
+    auto_default_missing_fields: bool = False,
```
badmonster0: I think we don't need to guard it as an option, as we already picked the types to support auto default very carefully. We don't have a place to set the option any way. Users get different behavior on different options. Removing the option makes the code simpler.

python/cocoindex/convert.py
badmonster0, line 1: Would you please also update `test_convert.py` to test the new feature?
You can leverage `validate_full_roundtrip()`, and pass in dataclasses with extra fields as `other_decoded_values`. Thanks!

python/cocoindex/tests/test_convert.py
```diff
@@ -1468,3 +1468,39 @@ class Team:
...
+    @dataclass
+    class Base:
+        a: int
+        b: int
+
+    @dataclass
+    class ExtraFieldSupported:
+        a: int
+        b: int
+        c: list[int] = field(default_factory=list)
```
georgeh0: This is using an explicit default value - which is not an auto default case we start to support by this PR. I think we want to test with the following cases: a field with `int | None` type, and a field with `list[Base]` type (as LTable), and a field with `dict[str, Base]` type (as KTable), and all shouldn't have explicit default value.

kingkushal16: @dataclass
    class Base:
        a: int

    @dataclass
    class NullableField:
        a: int
        b: int | None

    @dataclass
    class LTableField:
        a: int
        b: list[Base]

    @dataclass
    class KTableField:
        a: int
        b: dict[str, Base]

    @dataclass
    class UnsupportedField:
        a: int
        b: int
        
   would these dataclasses be fine?

badmonster0: These look good. Thanks!

kingkushal16: For the KtableField I am getting an error

TypeError: 'dict' object cannot be converted to 'Sequence' when using validate_full_roundtrip(KTableField(1, {}), KTableField) Can you guys help me out with this. Attaching the dumps below.


=================================== FAILURES ===================================
____________ test_auto_default_for_supported_and_unsupported_types _____________

    def test_auto_default_for_supported_and_unsupported_types() -> None:
        @dataclass
        class Base:
            a: int
    
        @dataclass
        class NullableField:
            a: int
            b: int | None
    
        @dataclass
        class LTableField:
            a: int
            b: list[Base]
    
        @dataclass
        class KTableField:
            a: int
            b: dict[str, Base]
    
        @dataclass
        class UnsupportedField:
            a: int
            b: int
    
        engine_val = [1]
    
        validate_full_roundtrip(NullableField(1, None), NullableField)
    
        validate_full_roundtrip(LTableField(1, []), LTableField)
    
        decoder = build_engine_value_decoder(KTableField)
        result = decoder(engine_val)
        assert result == KTableField(1, {})
    
>       validate_full_roundtrip(KTableField(1, {}), KTableField)

python/cocoindex/tests/test_convert.py:1508: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
python/cocoindex/tests/test_convert.py:122: in validate_full_roundtrip
    validate_full_roundtrip_to(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = test_auto_default_for_supported_and_unsupported_types.<locals>.KTableField(a=1, b={})
value_type = <class 'cocoindex.tests.test_convert.test_auto_default_for_supported_and_unsupported_types.<locals>.KTableField'>
decoded_values = ((test_auto_default_for_supported_and_unsupported_types.<locals>.KTableField(a=1, b={}), <class 'cocoindex.tests.test_convert.test_auto_default_for_supported_and_unsupported_types.<locals>.KTableField'>),)
_engine = <module 'cocoindex._engine' from '/home/kushal/Desktop/Open-Source/cocoindex/python/cocoindex/_engine.cpython-312-x86_64-linux-gnu.so'>
eq = <function validate_full_roundtrip_to.<locals>.eq at 0x7d7a6d6d9940>
encoded_value = [1, {}]

    def validate_full_roundtrip_to(
        value: Any,
        value_type: Any,
        *decoded_values: tuple[Any, Any],
    ) -> None:
        """
        Validate the given value becomes specific values after encoding, sending to engine (using output_type), receiving back and decoding (using input_type).
    
        `decoded_values` is a tuple of (value, type) pairs.
        """
        from cocoindex import _engine  # type: ignore
    
        def eq(a: Any, b: Any) -> bool:
            if isinstance(a, np.ndarray) and isinstance(b, np.ndarray):
                return np.array_equal(a, b)
            return type(a) is type(b) and not not (a == b)
    
        encoded_value = encode_engine_value(value)
        value_type = value_type or type(value)
        encoded_output_type = encode_enriched_type(value_type)["type"]
>       value_from_engine = _engine.testutil.seder_roundtrip(
            encoded_value, encoded_output_type
        )
E       TypeError: 'dict' object cannot be converted to 'Sequence'

python/cocoindex/tests/test_convert.py:101: TypeError

lemorage: Interesting. The test case is `validate_full_roundtrip(KTableField(1, {}), KTableField)`. Here the empty dict is being passed. When encoding struct fields that contained empty dicts (KTables), the function was preserving them as `{}`, but the expected engine format for KTables was `[]`.

badmonster0: Thank you both! Since fix for the issue (PR #807) may still need more time to be resolved, we can disable this specific test case (e.g. comment out the specific line) and merge the PR first. After the fix is in, we can re-enable it.

kingkushal16: Thanks guys, I've commented out the line for now and have added a test case just for the decoder.

python/cocoindex/convert.py
```diff
@@ -349,20 +409,27 @@ def make_closure_for_value(
...
+
+                return field_value_getter
 
+        default_value = param.default
+        if default_value is not inspect.Parameter.empty:
             return lambda _: default_value
 
+        auto_default = _handle_missing_field_with_auto_default(param, name, field_path)
+        return lambda _: auto_default
+
```
badmonster0: these lines need to be indented under the `with ChildFieldPath(field_path, f".{name}"):`, because it's for this specific field (the `.{name}` need to be included in error messages)

kingkushal16: Have addressed this. Thanks.

PR #796, opened 2025-07-24
Support using ColPali library to compute embedding

badmonster0:


examples/image_search_colpali/README.md
```diff
@@ -0,0 +1,71 @@
...
+  ```
+- You can also override these with environment variables:
+  ```sh
+  export QDRANT_URL="localhost:6334"           # for gRPC (default)
+  export QDRANT_PREFER_GRPC=true                # for gRPC (default)
+  # or for HTTP:
+  # export QDRANT_URL="http://localhost:6333/"
+  # export QDRANT_PREFER_GRPC=false
+  ```
+
```
athrael-soju: Just a minor suggestion to add, which may save some troubleshooting time.

Recent Nvidia GPUs (RTX 5090) will not work with the Stable pytorch version up to 2.7.1

If we get this error:
```
The current PyTorch install supports CUDA capabilities sm_37 sm_50 sm_60 sm_61 sm_70 sm_75 sm_80 sm_86 sm_90 compute_37.
```
we can install the nightly pytorch build here: https://pytorch.org/get-started/locally/
`pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu129`

cu128 and cu129 work fine.

examples/image_search_colpali/main.py
```diff
@@ -0,0 +1,212 @@
...
+        _colpali_model_cache[model] = {
+            "model": ColPali.from_pretrained(model),
+            "processor": ColPaliProcessor.from_pretrained(model),
+        }
+    return _colpali_model_cache[model]["model"], _colpali_model_cache[model][
+        "processor"
+    ]
+
+
+def colpali_embed_image(
```
georgeh0: I think it will be great if we can expose it as part of CocoIndex standard function, similar to the existing `SentenceTransformerEmbed`:

https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/functions.py#L51-L101

`model` may be part of the spec, so that this information is fixed at the instantiation time, and the output vector size can be determined based on the model (similar to the `SentenceTransformerEmbed` example).

examples/image_search_colpali/frontend/src/App.jsx
georgeh0, line 1: Seems the frontend code and also example images are the same with the existing `examples/image_search`? So I think instead of duplicating them, we may share the existing parts as much as possible. e.g. one way is to create a separate `main.py` (e.g. `colapi_main.py`) under `examples/image_search`, so most other files can be directly shared. What do you think?

python/cocoindex/functions.py
```diff
@@ -99,3 +99,227 @@ def __call__(self, text: str) -> NDArray[np.float32]:
         assert self._model is not None
         result: NDArray[np.float32] = self._model.encode(text, convert_to_numpy=True)
         return result
+
+
+# Global ColPali model cache to avoid reloading models
+_COLPALI_MODEL_CACHE = {}
```
georgeh0: We can use [`@functools.cache`](https://docs.python.org/3/library/functools.html#functools.cache) for this purpose. It's thread safe and reduces boilerplates.

python/cocoindex/functions.py
```diff
@@ -99,3 +99,227 @@ def __call__(self, text: str) -> NDArray[np.float32]:
...
+        return _COLPALI_MODEL_CACHE[model_name]
+
+    @staticmethod
+    def _detect_dimension() -> int:
+        """Detect ColPali embedding dimension using FastEmbed API."""
+        try:
+            from fastembed import LateInteractionMultimodalEmbedding
+
+            # Use the standard FastEmbed ColPali model for dimension detection
+            standard_colpali_model = "Qdrant/colpali-v1.3-fp16"
```
georgeh0: When `_detect_dimension()` is already called, we already know the specific mode. Why not use the specific model for this purpose?

pyproject.toml
```diff
@@ -31,11 +31,8 @@ features = ["pyo3/extension-module"]
 [project.optional-dependencies]
 dev = ["pytest", "ruff", "mypy", "pre-commit"]
 
-embeddings = ["sentence-transformers>=3.3.1"]
-
-# We need to repeat the dependency above to make it available for the `all` feature.
-# Indirect dependencies such as "cocoindex[embeddings]" will not work for local development.
```
georgeh0, line 37: Please bring back these comments. They provide useful guidance.

pyproject.toml
```diff
@@ -31,11 +31,8 @@ features = ["pyo3/extension-module"]
...
-all = ["sentence-transformers>=3.3.1"]
+embeddings = ["sentence-transformers>=3.3.1", "colpali-engine", "fastembed"]
```
georgeh0: I'm thinking we want to expose `colpali` as a separate feature, as they're specific models. People who need `sentence-transformers` may not need `colpali` and vice versa.

athrael-soju: Also, why fastembed?

badmonster0: thanks for the detailed review! removed fastembed with bunch of cleanups :)

examples/image_search_colpali/README.md
badmonster0: Thanks! Added this instruction to `README.md`:

https://github.com/cocoindex-io/cocoindex/pull/796/files#diff-4aa405bff6e9511426cb010b49e7495c937ed3e5161282400f56e4252d7b2aa4R79-R91

PR #807, opened 2025-07-26
feat(convert): support explicit type hints in engine value encoding

lemorage:
### Context
https://github.com/cocoindex-io/cocoindex/pull/792#discussion_r2232790662

### Cause
The problem was that `encode_engine_value` couldn't distinguish between:
  1. Empty dicts that should be KTables (should become [])
  2. Empty dicts that should be JSON objects (should stay {})

### Solution

~Make `encode_engine_value` function context-aware by adding an optional `_in_struct` and `type_hint` params.~

~The logic now:~
  ~- When encoding struct fields, if the type-hint denotes it is a JSON, it stays what it is, otherwise becomes [] (assuming KTable)~
  ~- When encoding standalone values (_in_struct=False): empty dicts stay {}~
  
Pass type hints when calling `encode_engine_value` function, so it will be explicitly known how we handle each value.


python/cocoindex/convert.py
```diff
@@ -6,52 +6,72 @@
...
+        ]
     if isinstance(value, np.number):
         return value.item()
     if isinstance(value, np.ndarray):
         return value
     if isinstance(value, (list, tuple)):
-        return [encode_engine_value(v) for v in value]
+        return [encode_engine_value(v, in_struct) for v in value]
     if isinstance(value, dict):
+        is_json_type = type_hint and any(
```
badmonster0: we may use `typing.analyze_type_info()` for this purpose. It takes care of different situations, e.g. `typing.Optional[cocoindex.Json]`

python/cocoindex/convert.py
```diff
@@ -6,52 +6,72 @@
...
+            isinstance(arg, TypeKind) and arg.kind == "Json"
+            for arg in get_args(type_hint)[1:]
+        )
+
+        # For empty dicts, check type hints if in a struct context
+        # when no contexts are provided, return an empty dict as default
         if not value:
+            if in_struct:
+                return value if is_json_type else []
             return {}
```
badmonster0: The top-level value can be either a Struct or a Json. To solve this problem robustly, I think we may want to always pass in the type annotation or engine-side type to the function.

Currently encoding happens in two places:
1. To encode return value of a custom function
2. To encode arguments passed to a transform flow when `eval()` method is called

For both, actually type annotation is required ([for 1](https://github.com/cocoindex-io/cocoindex/blob/0eaa7a277506f36912e0dffbe741580d683ee910/python/cocoindex/op.py#L126), [for 2](https://github.com/cocoindex-io/cocoindex/blob/0eaa7a277506f36912e0dffbe741580d683ee910/python/cocoindex/flow.py#L1028C49-L1028C59)). We just didn't pass it to this encode function (we took a short cut in the initial implementation).

We may also want to have a style similar to the decode functions, i.e. have a layer of function to analyze types and return encoder closures, and run encoder closures on actual data (so we save the computation to analyze type info for each value).

This is a larger change though. If you're interested and would like to take this, that will be very great. Otherwise, no worries, I'll create an issue and leave for later. Let me know what do you think. Thanks!

badmonster0 (commented):
Thanks a lot for taking care of this, @lemorage!

python/cocoindex/convert.py
lemorage: Thanks, sounds good. I'll try to see if I can implement this soon; otherwise, maybe I'll consider making this an issue and leave this unresolved.

python/cocoindex/convert.py
```diff
@@ -49,45 +49,146 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+    def encode_value(value: Any, in_struct: bool = False) -> Any:
+        return _encode_engine_value_core(
+            value, in_struct=in_struct, type_variant=type_info
+        )
+
+    return lambda value: encode_value(value, in_struct=False)
+
+
+def encode_engine_value(
+    value: Any, in_struct: bool = False, type_hint: Type[Any] | str | None = None
```
badmonster0: Since all callers have been updated with a `type_hint`, we always have a `type_hint`, and `in_struct` seems no longer needed?

python/cocoindex/convert.py
```diff
@@ -49,45 +49,146 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+        value: The Python value to encode
+        in_struct: Whether this value is being encoded within a struct context
+        type_hint: Type annotation for the value. When provided, enables optimized
+                  type-aware encoding. For top-level calls, this should always be provided.
+
+    Returns:
+        The encoded engine value
+    """
+    if type_hint is not None:
+        encoder = make_engine_value_encoder(type_hint)
```
badmonster0: Ideally we don't want to create this encoder for each value, as there're some complex logic in analyzing the type information and they only need to happen once and can be reused for all values (similar to the decode logic).

badmonster0 (commented):
Thanks a lot for starting this deep refactoring, which solves the problem more systematically!

python/cocoindex/convert.py
```diff
@@ -50,45 +50,153 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+            }
+
     return value
 
 
-_CONVERTIBLE_KINDS = {
-    ("Float32", "Float64"),
-    ("LocalDateTime", "OffsetDateTime"),
-}
+def encode_engine_value(value: Any, type_hint: Type[Any] | str | None = None) -> Any:
```
badmonster0: We may remove `= None` as it should always be provided.

python/cocoindex/convert.py
```diff
@@ -50,45 +50,153 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+
+
+def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool:
+    return (
+        src_type_kind == dst_type_kind
+        or (src_type_kind, dst_type_kind) in _CONVERTIBLE_KINDS
+    )
+
+
+def _get_cached_type_info(
```
badmonster0: We may use `@functools.cache`. It's threadsafe.

https://docs.python.org/3/library/functools.html#functools.cache

python/cocoindex/convert.py
```diff
@@ -50,45 +50,153 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+    """Retrieve or compute and cache the type information for a given type."""
+    if type_to_analyze not in cache:
+        cache[type_to_analyze] = analyze_type_info(type_to_analyze)
+    return cache[type_to_analyze]
+
+
+def _encode_engine_value_core(
+    value: Any,
+    type_hint: Type[Any] | str | None = None,
+    type_variant: AnalyzedTypeInfo | None = None,
```
badmonster0: We shouldn't need to pass both `type_hint` and `type_variant`. Only take one, to be more clear. Consider keep `type_variant` (and `type_info` is a better name for it), and caller can call `analyze_type_info()` if they only have the type hint (similar to the decode functions).

python/cocoindex/convert.py
lemorage: With this, I assume we'll then update all the tests accordingly.

python/cocoindex/convert.py
```diff
@@ -50,45 +51,138 @@ def __exit__(self, _exc_type: Any, _exc_val: Any, _exc_tb: Any) -> None:
...
+
+    @functools.cache
+    def _get_cached_type_info() -> AnalyzedTypeInfo:
+        """cache the computed type information for a given type."""
+        return analyze_type_info(type_to_analyze)
+
+    try:
+        return _get_cached_type_info(type_to_analyze)
+    except TypeError:
+        return analyze_type_info(type_to_analyze)
```
badmonster0: This starts to be heavy. When the type is not hashable, we always need to execute `analyze_type_info()` repeatedly, and a exception handling flow is also heavy (exceptions are designed for exceptional cases, so whenever an exception needs to be created and propagated, significantly heavier computation will happen).

Besides, within the logic of `_encode_engine_value_core()`, we need to go through multiple `if` to check the type, which is also something can be avoided.

The ideal implementation in my mind is something similar to the decoding side: the type inspection on `type_info` only happens once when `make_engine_value_decoder()` is called, and it returns a closure, and the closure is reused many times for different data (i.e. when the same function is invoked on values for different rows). There're a few examples of these closures:

https://github.com/cocoindex-io/cocoindex/blob/d77ae249c45442788ede2e23b0570839e5c55245/python/cocoindex/convert.py#L281-L292

https://github.com/cocoindex-io/cocoindex/blob/d77ae249c45442788ede2e23b0570839e5c55245/python/cocoindex/convert.py#L294

There's a lot of type checking happening before creating these closures, but they only happen once (for closure preparation), not at the time of execution on actual values. So it's quite light in using the prepared closures to convert values.

With this said, I'm also OK with merging this PR first and resolve this later as a separate issue. Let me know what you prefer. Thanks!

python/cocoindex/flow.py
```diff
@@ -1100,10 +1096,15 @@ async def eval_async(self, *args: Any, **kwargs: Any) -> T:
         flow_info = await self._flow_info_async()
         params = []
         for i, arg in enumerate(self._param_names):
+            param_type = (
+                self._flow_arg_types[i] if i < len(self._flow_arg_types) else None
```
badmonster0: I think it's clearer to directly use `else Any` here (and we don't need another `or Any` for two branches below). 

Actually the length of `self._param_names` and `self._flow_arg_types` are expected to have the same length. So it's also OK to use

```py
for i, (arg, arg_type) in enumerate(zip(self._param_names, self._flow_arg_types)):
```

python/cocoindex/convert.py
lemorage: Thanks for the detailed clarifications. Quite convinced. No worries, I'll try to refactor this logic asap.

badmonster0: thanks! take your time!

python/cocoindex/flow.py
lemorage: > ```python
> for i, (arg, arg_type) in enumerate(zip(self._param_names, self._flow_arg_types)):
> ```
Love this change!

python/cocoindex/convert.py
```diff
@@ -91,6 +64,224 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+        """cache the computed type information for a given type."""
+        return analyze_type_info(t)
+
+    try:
+        return _get_cached_type_info(type_to_analyze)
+    except TypeError:  # The type is not hashable
+        return analyze_type_info(type_to_analyze)
+
+
+def _make_encoder_closure(type_info: AnalyzedTypeInfo | None) -> Callable[[Any], Any]:
```
badmonster0: I think there should be no case that `type_info` is `None`.

`analyze_type_info()` already handles the case when the type hint is missing (the annotation type will be `inspect.Parameter.empty` in this case - equivalent to `Any`).

https://github.com/cocoindex-io/cocoindex/blob/c1c2a15f5d3561735f8ecf86100c6221b28b78e3/python/cocoindex/typing.py#L265-L266

`None`  in type annotation means the `None` value in Python (e.g. if a function is annotated as `def foo() -> None: ...` it means it return value is always `None`, which is different from empty or `Any`), and we're representing it as the `AnalyzedUnknownType` variant now as we don't support it.

python/cocoindex/convert.py
```diff
@@ -91,6 +64,224 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+
+    if isinstance(variant, AnalyzedListType):
+        elem_encoder = (
+            _make_encoder_closure(None)
+            if not variant.elem_type
+            else _make_encoder_closure(_get_type_info(variant.elem_type))
+        )
+
+        def encode_list(value: Any) -> Any:
+            if isinstance(value, (list, tuple)):
```
badmonster0: This type check seems not needed.

When this condition doesn't hold, it means user code has bug (e.g. they declared to return a list but the value is something else), 
- It should be caught by things like `mypy` ideally. If not, I also wouldn't proactively check the type and will just let it fail (just like when we write a Python function already with type hint like `str`, I wouldn't check it's really a `str` at runtime).
- If the type is not expected, `return value` won't solve the problem either and will trigger failure on engine side.

Only a `None` value check is necessary here, as `value` can be `None`.

python/cocoindex/convert.py
```diff
@@ -91,6 +64,224 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+                return [elem_encoder(v) for v in value]
+            return value
+
+        return encode_list
+
+    if isinstance(variant, AnalyzedDictType):
+        if not variant.value_type:
+            return lambda value: value
+
```
badmonster0, line 91: We also want to encode the keys.

python/cocoindex/convert.py
```diff
@@ -91,6 +64,224 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+    def _encode_ktable_dict(value: dict[Any, Any]) -> Any | None:
+        if not value:
+            return value
+        first_val = next(iter(value.values()))
+        if is_struct_type(type(first_val)):
+            encoder = _make_encoder_closure(None)
+            return [[encoder(k)] + encoder(v) for k, v in value.items()]
+        return None
+
```
badmonster0: Note that when we want to encode a value for the engine, the specific type annotation should always be available ([more explanation](https://cocoindex.io/docs/core/data_types#:~:text=When%20you%20pass%20a%20Python%20value%20to%20the%20engine%20(e.g.%20return%20values%20of%20a%20custom%20function)%2C%20a%20specific%20type%20annotation%20is%20required.)).

But later I just realized there's a complication caused by [union types](https://cocoindex.io/docs/core/data_types#union-types) - the specific type may be unknown (there're multiple variations and not so easy to decide). But union type variances are only [basic types](https://cocoindex.io/docs/core/data_types#basic-types).

So I think we may do type-hint-based encoding for Struct and Table types, and for basic types we only encode based on values, similar to the old way (among basic types, only numpy number/array types need to be handled specially, others don't need any change).

```
# Handle AnalyzedListType (only when element is an AnalyzedStructType) and return
...
# Handle AnalyzedDictType and return
...
# Handle AnalyzedStructType and return
...
# Now it's expected to be basic type, we can return a simple closure to encode the value based on its runtime type, only numpy number/array types need to be handled specially
...
```

python/cocoindex/convert.py
```diff
@@ -91,6 +64,209 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+            return value
+
+        return encode_any_type
+
+    def encode_basic_value(value: Any) -> Any:
+        if isinstance(value, np.number):
+            return value.item()
+        if isinstance(value, np.ndarray):
+            return value
```
badmonster0, line 165: If `value` is `list` or `tuple`, we want to call `encode_basic_value()` recursively on its elements (similar to before).

python/cocoindex/convert.py
```diff
@@ -91,6 +64,209 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
...
+                if not is_namedtuple_type(type(value)):
+                    return value
+                return [
+                    encoder(getattr(value, name))
+                    for encoder, name in zip(field_encoders, field_names)
+                ]
+
+            return encode_namedtuple
+
+    if isinstance(variant, AnalyzedAnyType):
```
badmonster0: We don't need this branch, as specific type annotation is required for python->engine. This case will fail in `encode_enriched_type_info()` first. We can assert `variant` is neither `AnalyzedAnyType` nor `AnalyzedUnknownType` here.

python/cocoindex/convert.py
```diff
@@ -91,6 +64,209 @@ def _is_type_kind_convertible_to(src_type_kind: str, dst_type_kind: str) -> bool
     )
 
 
+def _get_type_info(type_to_analyze: Any) -> AnalyzedTypeInfo:
```
badmonster0: I think we no longer needs this cache, and we can just always call `analyze_type_info()` directly, as we no longer call it for each specific row/element of data.

python/cocoindex/flow.py
```diff
@@ -1099,11 +1095,16 @@ async def eval_async(self, *args: Any, **kwargs: Any) -> T:
...
+        ):
+            param_type = (
+                self._flow_arg_types[i] if i < len(self._flow_arg_types) else Any
+            )
             if i < len(args):
-                params.append(encode_engine_value(args[i]))
+                params.append(encode_engine_value(args[i], type_hint=param_type))
             elif arg in kwargs:
-                params.append(encode_engine_value(kwargs[arg]))
+                params.append(encode_engine_value(kwargs[arg], type_hint=param_type))
```
badmonster0, line 1107: Now the heavy type analyzing part is still being executed for each specific data. We should do the similar thing as the [`result_decoder`](https://github.com/cocoindex-io/cocoindex/blob/c2b09e115fb90c0d8c6e156ca025c448af443ce5/python/cocoindex/flow.py#L1072-L1074). Specifically, in [this loop](https://github.com/cocoindex-io/cocoindex/blob/c2b09e115fb90c0d8c6e156ca025c448af443ce5/python/cocoindex/flow.py#L1038-L1053) we may create the encoder for each param and keep them as a member of `TransformFlowInfo`, then here we only need to call the encoder.

python/cocoindex/op.py
```diff
@@ -317,7 +317,7 @@ async def __call__(self, *args: Any, **kwargs: Any) -> Any:
                     output = await self._acall(*decoded_args, **decoded_kwargs)
             else:
                 output = await self._acall(*decoded_args, **decoded_kwargs)
-            return encode_engine_value(output)
+            return encode_engine_value(output, type_hint=expected_return)
```
badmonster0, line 320: Similar here, we may build the encoder and keep it before we run it on real data, maybe around [here](https://github.com/cocoindex-io/cocoindex/blob/c2b09e115fb90c0d8c6e156ca025c448af443ce5/python/cocoindex/op.py#L304). Then for each specific data we only run the encoder without analyzing type info.

python/cocoindex/flow.py
lemorage: Thanks for the advice! Do u think it would be ok to leave the following two as a separate issue?

badmonster0: Sounds good!

My only concern was this may cause performance regression, as doing this for each instance of data will be even slower than before. But I can take care of the remaining change.

I'll merge this PR once I'm available to make this change, so that this PR and the improvement will be out in the same release.

Thanks for all the effort in this PR! This really deeply touches a complex aspect of the codebase!

lemorage: Thanks for the reviews all the way through! Actually, I attempted to refactor these as well, but I found it hard to maintain a consistent mental flow when my time is scattered, as I have been busy with some other stuff lately. Thanks for your time in taking care of this!

badmonster0:
hi @lemorage latest release note out and we had a section for you, thanks for your contribution! https://cocoindex.io/blogs/cocoindex-changelog-2025-10-19#lemorage

PR #861, opened 2025-08-11
visual - doc spacing fix for algolia search

badmonster0:


docs/package-lock.json
georgeh0, line 1: We already have `yarn.lock`. I think we don't need this file?

docs/yarn.lock
```diff
@@ -4,229 +4,197 @@
 
 "@algolia/autocomplete-core@1.17.9":
   version "1.17.9"
-  resolved "https://registry.yarnpkg.com/@algolia/autocomplete-core/-/autocomplete-core-1.17.9.tgz#83374c47dc72482aa45d6b953e89377047f0dcdc"
+  resolved "https://registry.npmjs.org/@algolia/autocomplete-core/-/autocomplete-core-1.17.9.tgz"
```
georgeh0: are these updates intended? we're using `yarn` (as guided by `README` and also in the github workflows) instead of `npm`.

badmonster0:
done

georgeh0 (approved):

PR #883, opened 2025-08-19
organize docs for contributing guide

badmonster0:


docs/docs/contributing/guide.md
georgeh0, line 1: Can we add a redirect from the old URL to here?

https://github.com/cocoindex-io/cocoindex/blob/e8bd0e205ba0ae12924670aa80a8cec58dc3cb8f/docs/docusaurus.config.ts#L55-L68

badmonster0: done, ty

PR #938, opened 2025-08-29
examples: simplify postgres source example to be a single flow

badmonster0:


georgeh0 (approved):
can you also help to delete `.env.example`?

PR #958, opened 2025-09-02
docs: postgres example

badmonster0:
postgre example docs


docs/docs/examples/examples/postgres_source.md
```diff
@@ -0,0 +1,298 @@
...
+            ordinal_column="modified_time",
+            notification=cocoindex.sources.PostgresNotification(),
+        ),
+    )
+```
+This step adds source data from PostgreSQL table `source_products` to the flow as a `KTable`.
+
+![Add PostgreSQL Source](/img/examples/postgres_source/source.png)
+
+- Incremental Sync: When new or updated rows are found, only those rows are run through the pipeline, so downstream indexes and search results reflect the latest data while unchanged rows are untouched.
```
georgeh0, line 56: s/are run through/run through/

docs/docs/examples/examples/postgres_source.md
```diff
@@ -0,0 +1,298 @@
...
+- `ordinal_column` is recommended for change detection so the pipeline processes what's changed.
+- `notification`: when present, enable change capture based on Postgres LISTEN/NOTIFY.
```
georgeh0, line 58: These three bullets are related to each other. Users may not fully understand their relationships. We may rewrite to something like this to make it more clear:

```md
CocoIndex incrementally sync data from Postgres. When new or updated rows are found, only those rows run through the pipeline, so downstream indexes and search results reflect the latest data while unchanged rows are untouched. The following two arguments (both are optional) make this more efficient:

- `notification` enables change capture based on Postgres LISTEN/NOTIFY. Each change triggers an incremental processing on the specific row immediately.
- Regardless if `notification` is provided or not, CocoIndex still needs to scan the full table to detect changes in some scenarios (e.g. between two `update` invocation), and the `ordinal_column` provides a field that CocoIndex can use to quickly detect which row  has changed without reading value columns.
```

docs/docs/examples/examples/postgres_source.md
```diff
@@ -0,0 +1,298 @@
...
+    ],
+)
+```
+
+All transformed rows are collected and exported to a new PostgreSQL table with a vector index, ready for semantic search.
+
+
+## Field lineage
+When the transform flow starts to getting complex, it's hard to understand how each field is derived.
+CocoIndex provides a way to visualize the lineage of each field, to make it easier to trace and troubleshoot field origins and downstream dependencies.
```
georgeh0, line 173: s/CocoIndex/CocoInsight/?

docs/docs/examples/examples/postgres_source.md
```diff
@@ -0,0 +1,298 @@
...
+### Run as a Service
+
+This [example](https://cocoindex.io/docs/examples/image_search#fast-api-application) runs as a service using Fast API.
+
+
+## Why One Framework for Structured + Unstructured?
+
+- One mental model: Treat files, APIs, and databases uniformly; AI steps are ordinary ops.
+- Incremental by default: Use an ordinal column to sync only changes; no fragile glue jobs.
+- Consistency: Embeddings are always derived from the exact transformed row state.
```
georgeh0, line 296: What does "state" mean here exactly?

Don't quite understand the point it wants to emphasize on.

docs/docs/examples/examples/postgres_source.md
georgeh0: This is not resolved yet.

docs/docs/examples/examples/postgres_source.md
georgeh0: And this is also not resolved yet.

PR #1010, opened 2025-09-17
feat(ops): SplitBySeparators — regex-only splitter (Rust+Py)

### LINKED ISSUE #1003: [FEATURE] Create a simple `SplitBySeparators` function (don't account chunk size)
Today we have a function [`SplitRecursively`](https://cocoindex.io/docs/ops/functions#splitrecursively) which consider both chunk size and separators during splitting.

Sometimes users may need a simpler behavior: just split by specified separators without considering chunk size.
For example, when users specify `\n\n+`, it simply split the text into paragraphs (considering paragraphs always separated by empty lines in between).

You can look into the [existing implementation of `SplitRecursively`](https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex+SplitRecursively+-path%3Aexample&type=code) for the overall structure, but this `SplitBySeparators` will be much simpler.

---

MrAnayDongre:
Adds `SplitBySeparators` (regex-only, no chunk-size planning).
Rust executor + registration; Python spec; schema matches `SplitRecursively`.
143 passed / 1 skipped (DB). Closes #1003. cc @georgeh0

### DIFF
--- a/src/ops/functions/split_by_separators.rs
+++ b/src/ops/functions/split_by_separators.rs
@@ -1,0 +1,299 @@
+use anyhow::{Context, Result};
+use regex::Regex;
+use std::sync::Arc;
+
+use crate::base::field_attrs;
+use crate::ops::registry::ExecutorFactoryRegistry;
+use crate::ops::shared::split::{Position, set_output_positions};
+use crate::{fields_value, ops::sdk::*};
+
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+#[serde(rename_all = "UPPERCASE")]
+enum KeepSep {
+    NONE,
+    LEFT,
+    RIGHT,
+}
+
+#[derive(Serialize, Deserialize)]
+struct Spec {
+    // Python SDK provides defaults/values.
+    separators_regex: Vec<String>,
+    keep_separator: KeepSep,
+    include_empty: bool,
+    trim: bool,
+}
+
+struct Args {
+    text: ResolvedOpArg,
+}
+
+struct Executor {
+    spec: Spec,
+    regex: Option<Regex>,
+    args: Args,
+}
+
+impl Executor {
+    fn new(args: Args, spec: Spec) -> Result<Self> {
+        let regex = if spec.separators_regex.is_empty() {
+            None
+        } else {
+            // OR-join all separators, multiline
+            let pattern = format!(
+                "(?m){}",
+                spec.separators_regex
+                    .iter()
+                    .map(|s| format!("(?:{s})"))
+                    .collect::<Vec<_>>()
+                    .join("|")
+            );
+            Some(Regex::new(&pattern).context("failed to compile separators_regex")?)
+        };
+        Ok(Self { args, spec, regex })
+    }
+}
+
+struct ChunkOutput<'s> {
+    start_pos: Position,
+    end_pos: Position,
+    text: &'s str,
+}
+
+#[async_trait]
+impl SimpleFunctionExecutor for Executor {
+    async fn evaluate(&self, input: Vec<Value>) -> Result<Value> {
+        let full_text = self.args.text.value(&input)?.as_str()?;
+        let bytes = full_text.as_bytes();
+
+        // add_range applies trim/include_empty and records the text slice
+        let mut chunks: Vec<ChunkOutput<'_>> = Vec::new();
+        let mut add_range = |mut s: usize, mut e: usize| {
+            if self.spec.trim {
+                while s < e && bytes[s].is_ascii_whitespace() {
+                    s += 1;
+                }
+                while e > s && bytes[e - 1].is_ascii_whitespace() {
+                    e -= 1;
+                }
+            }
+            if self.spec.include_empty || e > s {
+                chunks.push(ChunkOutput {
+                    start_pos: Position::new(s),
+                    end_pos: Position::new(e),
+                    text: &full_text[s..e],
+                });
+            }
+        };
+
+        if let Some(re) = &self.regex {
+            let mut start = 0usize;
+            for m in re.find_iter(full_text) {
+                let end = match self.spec.keep_separator {
+                    KeepSep::LEFT => m.end(),
+                    KeepSep::NONE | KeepSep::RIGHT => m.start(),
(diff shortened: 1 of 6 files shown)


src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[derive(Serialize, Deserialize)]
+struct Spec {
+    #[serde(default)]
+    separators_regex: Vec<String>,
+    #[serde(default)]
+    keep_separator: KeepSep,
+    #[serde(default)]
+    include_empty: bool,
+    #[serde(default = "default_trim")]
```
georgeh0: I guess we don't have to explicitly set these default values here.

Python SDK already has the default values. It's better to keep default values in a single place (no need to worry about out of sync).

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,432 @@
+use anyhow::{Result, anyhow};
+use regex::Regex;
+use std::sync::Arc;
+
+use crate::base::field_attrs;
+use crate::{fields_value, ops::sdk::*};
+
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
```
georgeh0: Prefer `UPPERCASE` here (or `SCREAMING_SNAKE_CASE`, but no difference for this case). It's also the naming convention for enum variants in Python.

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,432 @@
...
+            line,
+            column,
+        });
+        if let Some(p) = positions_iter.next() {
+            next_position = p;
+        } else {
+            return;
+        }
+    }
+}
```
georgeh0: It's great if we can extract these common part into a shared library. We can put it in `src/ops/functions/shared/split.rs`.

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[async_trait]
+impl SimpleFunctionExecutor for Executor {
+    async fn evaluate(&self, input: Vec<Value>) -> Result<Value> {
+        let full_text = self.args.text.value(&input)?.as_str()?;
+
+        // If no separators were provided, emit whole text once.
+        if self.regex.is_none() {
+            return emit_single_chunk_kv(full_text, self.spec.trim, self.spec.include_empty);
```
georgeh0: This function has ~40 lines of code, but I feel we don't need it.

We can reuse the logic below. Only the way to compute `ranges` are different.

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,432 @@
...
+                while s < e && bytes[s].is_ascii_whitespace() {
+                    s += 1;
+                }
+                while e > s && bytes[e - 1].is_ascii_whitespace() {
+                    e -= 1;
+                }
+            }
+            if self.spec.include_empty || e > s {
+                filtered.push((s, e));
+            }
```
georgeh0: (optional) we may consider putting this filter logic into a closure, e.g. `add_range`. Then in the loop above we can directly add filtered range, e.g. `add_range(prev, m.start())`. We can avoid one intermediate `Vec`.

badmonster0:
btw great PR!! Thanks a lot @MrAnayDongre

MrAnayDongre:
Applied feedback: enum UPPERCASE, removed Rust-side defaults, extracted shared split helpers, simplified range logic. 
Re-tested: 143/1 + cargo tests green

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,299 @@
...
+use crate::ops::registry::ExecutorFactoryRegistry;
+use crate::ops::shared::split::{Position, set_output_positions};
+use crate::{fields_value, ops::sdk::*};
+
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+#[serde(rename_all = "UPPERCASE")]
+enum KeepSep {
+    NONE,
+    LEFT,
+    RIGHT,
```
georgeh0, line 15: We should still use PascalCase for enum variants in Rust. That's the convention

(and that's why `rename_all` is needed to change the case for serialization)

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,299 @@
...
+    ) -> Result<(Args, EnrichedValueType)> {
+        // one required arg: text: Str
+        let args = Args {
+            text: args_resolver
+                .next_arg("text")?
+                .expect_type(&ValueType::Basic(BasicValueType::Str))?
+                .required()?,
+        };
+
+        // start/end structs exactly like SplitRecursively
```
georgeh0, line 157: This hints that we should factor out the common schema construction logic between `SplitRecursively` and this one into `shared/split.rs` too.

src/ops/functions/split_by_separators.rs
```diff
@@ -0,0 +1,299 @@
...
+    NONE,
+    LEFT,
+    RIGHT,
+}
+
+#[derive(Serialize, Deserialize)]
+struct Spec {
+    // Python SDK provides defaults/values.
+    separators_regex: Vec<String>,
+    keep_separator: KeepSep,
```
georgeh0, line 22: I think it's more idiomatic if we leave only 2 variants in `KeepSep` (left and right) and make this one a `Option<KeepSep>`. And on Python side the type can be `Literal["LEFT", "RIGHT"] | None`. In this way we just reuse the `None` value in both languages as the default one.

src/ops/shared/split.rs
georgeh0, line 1: Thanks for extracting the common code here!

Can we also update `split_recursively.rs` accordingly to use this library and avoid duplication?

georgeh0:
Thanks for contributing! I'll merge it for now and take over the remaining comments.

badmonster0:
hi @MrAnayDongre https://cocoindex.io/blogs/cocoindex-changelog-2025-10-19#mranaydongre latest release out and we had a section for you. thanks for your contribution!

Issue #1049, opened 2025-09-25
[FEATURE] Allow users to configure specific types of vector index


georgeh0:
The request comes from #1043 originally.

There're different vector index types, e.g. Postgres support IVFFlat / HNSW. Currently we default to HNSW. We can expose an option to change the default and use IVFFlat instead of HNSW. We should also expose extra options for each type, e.g. `m` and `ef_construction` for HNSW.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

banrovegrie:
Hi @georgeh0, I was trying to look into this and the other issue at #1043. Put up a draft PR at #1050. Hoping that helps?

georgeh0:
Thanks a lot for contributing, @banrovegrie. Will make a new release soon on this new feature.

Issue #1043, opened 2025-09-24
[BUG] Explicitly specifying `output_dimension` doesn't work for Gemini embedding API


AurumnPegasus:
**What is the use case?**
Have some way of native implementation of either IVFFlat or HNSW (both of which are natively supported by pgvector) or allow for sending metadata to EmbedText class (say, to use MRL which [Gemini provides](https://ai.google.dev/gemini-api/docs/embeddings)  via parameter `config=types.EmbedContentConfig(output_dimensionality=768)`  )

**Describe the solution you'd like**
ideally as simple implementation where we can add config or params to EmbedText or Vector to implement these changes. 

**Additional context**

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

georgeh0:
Hi @AurumnPegasus , thanks for filing this feature request!

To make sure we understand correctly about your request. Two parts are mentioned here:

1. IVFFlat / HNSW. Currently we default to HNSW. We can expose an option to change the default and  use IVFFlat instead of HNSW, by adding an option to [`VectorIndexDef`](https://cocoindex.io/docs/core/flow_def#vector-index).

2. For extra parameters passing to embedding, we already support specifying `output_dimensionality` in the [`EmbedText` function](https://cocoindex.io/docs/ops/functions#embedtext). Does this already serve for your need? Or is there any other parameters you're looking for?

Thanks!

AurumnPegasus:
For 1, extra option would be perfect (maybe additional param to also define number of clusters but a appropriate default should be fine too)

For 2, `output_dimension` doesn't work as expected. Here, what I wanted was to use [MRL in gemini](https://ai.google.dev/gemini-api/docs/embeddings#control-embedding-size) via the `output_dimensionality` parameter which is the input to the gemini client. In cocoindex's case, I dont understand how `output_dimension` works but I dont think it acts as an input to the client (since I constantly get 

```
[2025-09-24T18:34:09Z ERROR cocoindex_engine::execution::source_indexer] Error in processing row from flow `CodeEmbedding` source `files` with key: {"utils.py"}

    Caused by:
[...]

PR #1050, opened 2025-09-25
feat: fix Gemini embedding config and expose Postgres index tuning

banrovegrie:
This PR attempts at rolling in two related improvements. It addresses both #1049 and #1043.

First, Gemini embeddings now mirror `outputDimensionality` inside the `config` block, so `gemini-embedding-*` honors custom dimensions as requested by AurumnPegasus.

Second, we expose vector index tuning for Postgres: callers can choose between **HNSW** and **IVFFlat** (with parameters) from Python, and the engine creates the matching index with pgvector options and updated naming.

**Changes**
- `src/llm/gemini.rs`: build the embed payload via a helper so Gemini gets both top-level and `config` `outputDimensionality`.
- `src/base/spec.rs`: add `VectorIndexMethod` (HNSW/IVFFlat with optional params), default existing indexes to HNSW, extend `VectorIndexDef` formatting.
- `src/ops/targets/postgres.rs`: honor the selected method when generating `CREATE INDEX`, emit `WITH` options (`m`, `ef_construction`, `lists`), suffix non-default methods in the index name.
- `python/cocoindex/index.py`, `python/cocoindex/__init__.py`: expose matching Python dataclasses (`HnswVectorIndexMethod`, `IvfFlatVectorIndexMethod`) so flows can configure index methods/options.
- `docs/docs/core/flow_def.mdx`, `docs/docs/examples/examples/simple_vector_index.md`: document the new `method` field and show an IVFFlat example.

P.S. most of the changes are schema/SQL plumbing but I couldn't find existing tests covering the vector-index SQL paths (not sure if I missed anything).

### DIFF
--- a/src/base/spec.rs
+++ b/src/base/spec.rs
@@ -406,16 +406,6 @@ impl VectorIndexMethod {
             Self::IvfFlat { .. } => "IvfFlat",
         }
     }
-
-    pub fn is_default(&self) -> bool {
-        matches!(
-            self,
-            Self::Hnsw {
-                m: None,
-                ef_construction: None,
-            }
-        )
-    }
 }
 
 impl fmt::Display for VectorIndexMethod {
@@ -458,9 +448,6 @@ impl fmt::Display for VectorIndexDef {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &self.method {
             None => write!(f, "{}:{}", self.field_name, self.metric),
-            Some(method) if method.is_default() => {
-                write!(f, "{}:{}", self.field_name, self.metric)
-            }
             Some(method) => write!(f, "{}:{}:{}", self.field_name, self.metric, method),
         }
     }
--- a/src/ops/targets/postgres.rs
+++ b/src/ops/targets/postgres.rs
@@ -501,7 +501,7 @@ fn to_vector_index_name(table_name: &str, vector_index_def: &spec::VectorIndexDe
         vector_index_def.field_name,
         to_vector_similarity_metric_sql(vector_index_def.metric)
     );
-    if let Some(method) = vector_index_def.method.as_ref().filter(|m| !m.is_default()) {
+    if let Some(method) = vector_index_def.method.as_ref() {
         name.push_str("__");
         name.push_str(&method.kind().to_ascii_lowercase());
     }


banrovegrie:
Update: fixed the rust formatting error in `spec.rs`.

Someone kindly run the checks once again for this.

python/cocoindex/index.py
```diff
@@ -17,6 +37,7 @@ class VectorIndexDef:
 
     field_name: str
     metric: VectorSimilarityMetric
+    method: VectorIndexMethod = field(default_factory=HnswVectorIndexMethod)
```
georgeh0: Let's make this field optional and set the default to `None`. `None` means using the default configs decided by the target. A few considerations:

- Not all targets support configuring indexing method.
- This will also avoid a change of the generated spec if they stay with the default - which will trigger other changes down the road (e.g. invalidate intermediate data caches, update setup states in db, etc.)

On rust side the corresponding field should also be optional.

src/base/spec.rs
```diff
@@ -384,15 +384,92 @@ impl fmt::Display for VectorSimilarityMetric {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "type", rename_all = "snake_case")]
```
georgeh0: Use `"kind"` as tag here. It's the existing convention of other cocoindex code.

(we pick "kind" instead of "type" as "type" is usually already a keyword or builtin symbol in various languages)

src/base/spec.rs
```diff
@@ -384,15 +384,92 @@ impl fmt::Display for VectorSimilarityMetric {
...
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        ef_construction: Option<u32>,
+    },
+    IvfFlat {
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        lists: Option<u32>,
+    },
+}
+
+impl Default for VectorIndexMethod {
```
georgeh0: After we use `Option<VectorIndexMethod>`, we no longer need to implement `Default` for it.

src/base/spec.rs
georgeh0: We don't need `rename_all = "snake_case"` (we don't do it for other types too). PascalCase is good as the literal value of kind string.

python/cocoindex/index.py
```diff
@@ -9,6 +9,26 @@ class VectorSimilarityMetric(Enum):
     INNER_PRODUCT = "InnerProduct"
 
 
+@dataclass
+class HnswVectorIndexMethod:
+    """HNSW vector index parameters."""
+
+    type: str = field(init=False, default="hnsw")
```
georgeh0: After we change the field name to `kind` (please refer to other comments on Rust side), this can be:

```py
kind = "Hnsw"
```
(in our load/dump logic we handle class field "kind")

georgeh0:
Thanks a lot for contributing!

banrovegrie:
@georgeh0 thankyou for the review. I pushed a change addressing the comments. Updates: 

- made `method` optional
- switched the serde `tag` to kind with PascalCase values
- aligned the Python dataclasses to use `kind = "Hnsw"` / `"IvfFlat"`
- letting the Postgres SQL helper default to HNSW when method is None while still lowercasing the suffix for non-default cases
- updated the serialized "hnsw"/"ivfflat" strings

src/base/spec.rs
banrovegrie: Yep yep makes sense.

python/cocoindex/index.py
banrovegrie: I pushed the latest change so now we have:

- `VectorIndexDef.method` is now `Option<VectorIndexMethod>` in Rust (and None by default in the Python dataclass), so leaving it unset preserves the target’s own default behavior.
- Targets that don’t support extra tuning continue to see `None` and therefore behave exactly as before; specs are unchanged unless someone explicitly sets a method, so there’s no unintended cascade (no cache invalidation, no setup churn).
- Postgres, for example, falls back to `USING hnsw` when `method` is `None`, and only emits the extra SQL when a method/parameters are provided.

src/base/spec.rs
```diff
@@ -384,15 +384,85 @@ impl fmt::Display for VectorSimilarityMetric {
...
+
+impl VectorIndexMethod {
+    pub fn kind(&self) -> &'static str {
+        match self {
+            Self::Hnsw { .. } => "Hnsw",
+            Self::IvfFlat { .. } => "IvfFlat",
+        }
+    }
+
+    pub fn is_default(&self) -> bool {
```
georgeh0: We should no longer need this, as we don't have a global default. For example, in implementation for `fmt::Display`, I think we may always explicitly print out the method is explicitly set.

banrovegrie: Yep yep.

badmonster0:
Great PR, thanks a lot @banrovegrie  for your contribution!

banrovegrie:
@badmonster0 @georgeh0 thankyou. We have been using cocoindex for our startup and it has been a pleasure!

PR #1066, opened 2025-09-29
 fix: make example paths os-friendly using os.path.join()

### LINKED ISSUE #973: Make paths in examples friendly to Windows
We use `/` as path separators, which works well for Linux and macOS, e.g.

https://github.com/cocoindex-io/cocoindex/blob/074bf52a5b39d482ed8a56fde36d75822225f288/examples/docs_to_knowledge_graph/main.py#L67-L71

We want to update all examples to make the paths also work for Windows, e.g. here we may use `os.path.join('..', '..', 'docs', 'docs', 'core')` so it works under all operating systems.

thisisharsh7:
## Description

Replaces hardcoded forward slash path separators with `os.path.join()` in 3 examples and their corresponding documentation files 

## Related issues
  Fixes #973

### DIFF
--- a/docs/docs/examples/examples/codebase_index.md
+++ b/docs/docs/examples/examples/codebase_index.md
@@ -55,10 +55,12 @@ The flow is composed of the following steps:
 We will index the CocoIndex codebase. Here we use the `LocalFile` source to ingest files from the CocoIndex codebase root directory.
 
 ```python
+import os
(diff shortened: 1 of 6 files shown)


thisisharsh7:
kindly review it and Let me know if any adjustments are needed!! thanks.

PR #1087, opened 2025-10-01
feat: Add field-level descriptions to FieldSchema

### LINKED ISSUE #1074: [FEATURE] Enable users providing field-level description and carry it over to JSON schema
Since #795 is resolved, we can enable users providing field-level description and carry it over to JSON schema!

We want to add a `description` field to `FieldSchema`:
- [Rust type](https://github.com/cocoindex-io/cocoindex/blob/d3090761a4a8ea786e1daa5535719c07d04454b7/src/base/schema.rs#L321-L328). Should be in `Arc<str>` type.
- [Python type](https://github.com/cocoindex-io/cocoindex/blob/d3090761a4a8ea786e1daa5535719c07d04454b7/python/cocoindex/typing.py#L623-L635)
- Create `FieldSchema` based on pydantic model: [code](https://github.com/cocoindex-io/cocoindex/blob/d3090761a4a8ea786e1daa5535719c07d04454b7/python/cocoindex/typing.py#L381-L389)
- Then we can put that into the Json schema: [Json schema builder code](https://github.com/cocoindex-io/cocoindex/blob/5a550dcb84d5e0958fc9ab74bc8fb8e605b8dcc6/src/base/json_schema.rs#L206-L224)
 

---

belloibrahv:
- Add description field to Rust FieldSchema struct (`Arc<str>`)
- Update Python FieldSchema to include description field
- Extract Pydantic field descriptions during schema creation
- Update JSON schema builder to include field descriptions
- Add comprehensive tests for field description functionality

Resolves #1074

### DIFF
--- a/src/base/json_schema.rs
+++ b/src/base/json_schema.rs
@@ -45,10 +45,27 @@ impl JsonSchemaBuilder {
         if self.options.extract_descriptions {
             let mut fields: Vec<_> = field_path.iter().map(|f| f.as_str()).collect();
             fields.reverse();
-            self.extra_instructions_per_field
-                .insert(fields.join("."), description.to_string());
+            let field_path_str = fields.join(".");
+            
+            // Check if we already have a description for this field
+            if let Some(existing_description) = self.extra_instructions_per_field.get(&field_path_str) {
+                // Concatenate descriptions with newline separator
+                let combined_description = format!("{}\n{}", existing_description, description.to_string());
+                self.extra_instructions_per_field
+                    .insert(field_path_str, combined_description);
+            } else {
+                self.extra_instructions_per_field
+                    .insert(field_path_str, description.to_string());
+            }
         } else {
-            schema.metadata.get_or_insert_default().description = Some(description.to_string());
+            let metadata = schema.metadata.get_or_insert_default();
+            if let Some(existing_description) = &metadata.description {
+                // Concatenate descriptions with newline separator
+                let combined_description = format!("{}\n{}", existing_description, description.to_string());
+                metadata.description = Some(combined_description);
+            } else {
+                metadata.description = Some(description.to_string());
+            }
         }
     }
 
@@ -334,6 +351,7 @@ pub fn build_json_schema(
             fields: Arc::new(vec![schema::FieldSchema {
                 name: object_wrapper_field_name.clone(),
                 value_type: value_type.clone(),
+                description: None,
             }]),
             description: None,
         };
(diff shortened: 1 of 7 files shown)


belloibrahv:
@georgeh0 I've implemented the field-level descriptions feature for `FieldSchema` as requested in #1074. Kindly review when you have time.

src/base/json_schema.rs
```diff
@@ -219,6 +219,10 @@ impl JsonSchemaBuilder {
                             *instance_type = SingleOrVec::Vec(types);
                         }
                     }
+                    // Set field description if available
+                    if let Some(description) = &f.description {
+                        self.set_description(&mut schema, description, field_path.prepend(&f.name));
```
georgeh0, line 245: We have a problem of description overwrite here.

The schema returned by `for_enriched_value_type()` above may already have a description. e.g. `self.description()` is already called in many places based on the type ([example](https://github.com/belloibrahv/cocoindex/blob/99ef439dc2bcf9f9e29def183406868e30d33dbd/src/base/json_schema.rs#L115-L119)).

I think we may concatenate both descriptions (field level + type level, separated by a newline).

Here's one way to do this:

- Change `set_description()` to "append" instead of "overwrite" if we already have a description.

- Instead of calling `set_description()` here (which is after the type is already processed, so field description are placed after type description), we can pass the description as a argument down to `for_enriched_value_type()` / `for_value_type()` / `for_basic_value_type()` / etc. Then we can call `set_description()` with the passed-in description right after we create `SchemaObject` (which is before `set_description()` called based on type) - this will make sure the type description is added after field descriptions.

Please let me know if there's questions on this. Thanks!

georgeh0 (commented):
Nice! Mostly looks good - just one minor thing to fix.

src/base/json_schema.rs
belloibrahv: @georgeh0 ,I have added the changes concatenate field and type descriptions instead of overwriting, kindly review it

georgeh0:
Thanks! There're two checks failed. You may install `pre-commit` hooks and use it to fix automatically:

```sh
pre-commit install
pre-commit run --all-files
```

badmonster0:
ops sorry closed by accident, this is a great PR, just a final touch on the format. Thanks @belloibrahv !

belloibrahv:
> ops sorry closed by accident, this is a great PR, just a final touch on the format. Thanks @belloibrahv !

No problem, thanks for opening the PR back

belloibrahv:
> Thanks! There're two checks failed. You may install `pre-commit` hooks and use it to fix automatically:
> 
> ```shell
> pre-commit install
> pre-commit run --all-files
> ```

I have updated the PR with the required changes as requested, and I ensured that all failing checks were resolved locally. Please help me review when you have time.

georgeh0:
Thanks for the PR! Merging.

PR #1096, opened 2025-10-02
generate full CLI docs from click's metadata and also added it in precommit hooks

Davda-James:
Generate full CLI docs from click's metadata and also added it in pre commit hooks

Ran precommit hooks.

Added one Readme for clearance what /(dev contains if not needed will remove it pls tell me)

docs will be generated at docs/docs/core/cli-reference.md also added it in pre  commit hooks (can check in image, name: generate-cli-docs)

image

### DIFF
--- a/dev/generate_cli_docs.py
+++ b/dev/generate_cli_docs.py
@@ -27,9 +27,13 @@ except ImportError as e:
 
 
 def clean_usage_line(usage: str) -> str:
-    """Clean up the usage line to remove 'cli' and make it generic."""
-    # Replace 'cli' with 'cocoindex' in usage lines
-    return usage.replace("Usage: cli ", "Usage: cocoindex ")
+    """Clean up the usage line to remove 'cli' and make it generic, and remove the 'Usage:' prefix."""
+    # Replace 'cli' with 'cocoindex' in usage lines and remove 'Usage:' prefix
+    cleaned = usage.replace("Usage: cli ", "cocoindex ")
+    # Handle case where it might be "Usage: cocoindex" already
+    if cleaned.startswith("Usage: cocoindex "):
+        cleaned = cleaned.replace("Usage: cocoindex ", "cocoindex ")
+    return cleaned
 
 
 def escape_html_tags(text: str) -> str:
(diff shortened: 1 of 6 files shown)


Davda-James:
@georgeh0 I have added a Readme /dev was earlier thinking no need to, but added for clarity, pls confirm do we need it or not ?

Davda-James:
@georgeh0 resolved the test-docs ci failing issue now, all checks passed

dev/generate_cli_docs.py
```diff
@@ -0,0 +1,337 @@
...
+import re
+from typing import Dict, List, Any
+
+# Add the cocoindex python directory to the path
+project_root = Path(__file__).parent.parent
+python_path = project_root / "python"
+sys.path.insert(0, str(python_path))
+
+try:
+    import md_click
```
georgeh0, line 21: I think we want to add this as an optional dependency at https://github.com/cocoindex-io/cocoindex/blob/661f8d746175a3d08e4253f6b587a16131aa1a2a/pyproject.toml#L65

docs/docs/core/cli.mdx
```diff
@@ -73,3 +73,7 @@ Use `--help` to see the full list of subcommands, and `subcommand --help` to see
 cocoindex --help       # Show all subcommands
 cocoindex show --help  # Show usage of "show" subcommand
 ```
+
+## Detailed Command Reference
```
georgeh0: Wondering will it work if we directly import the file?

https://stackoverflow.com/questions/65741307/how-to-include-import-markdown-file-content-in-docusaurus-v2?utm_source=chatgpt.com

This can be tested out by `yarn start` under the `docs/` directory.

Also if import works, the "Subcommands" section above can also be removed as it's already generated. And top-level headers / overviews from the generated markdown can also be removed.

Let me know if import doesn't work. We can submit it as is. Thanks!

georgeh0 (commented):
This looks great! Thanks!

georgeh0:
> @georgeh0 I have added a Readme /dev was earlier thinking no need to, but added for clarity, pls confirm do we need it or not ?

This provides more background. We can keep it. Thanks!

docs/docs/core/cli.mdx
Davda-James: @georgeh0 done its working, will soon add commit.

Davda-James: Fixing some formatting so markdownlinter do not cry..

Davda-James: Done @georgeh0, pls check.

Issue #998, opened 2025-09-14
[FEATURE] Support transient flows


georgeh0:
CocoIndex treats transformation as persistent / keep-going tasks that keep maintaining and evolving the outcome on source data / flow change.

Sometimes users want the other way: treat data transformations as transient tasks (fire and forget).

Specifically:

- Add a new argument `transient` to the `@cocoindex.flow_def` decorator, e.g.

  ```py
  @cocoindex.flow_def(name="MyFlow", transient=True)
  def my_flow(flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope):
    ...
  ```

- [Live update](https://cocoindex.io/docs/core/flow_methods#live-update) is not supported for a transient flow

- In this mode, both _setup_ and _update_ are transient tasks (fir and forget) without keeping any states in CocoIndex's internal storage.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

shresthashim:
hey @georgeh0 , can you assign me this issue?

shresthashim:
hey @georgeh0  and @badmonster0 , I have raised a PR for this issue. Please review the PR and have it merged please

georgeh0:
Hi @shresthashim, 

Thanks for creating the PR! The functionality achieved there is different from what we expect for this feature request - sorry it's my fault and I think I probably should document the requirement more clearly and in more details to avoid confusion.

*Transient flows* requested by this feature should support both _setup_ (i.e. create the target) and _update_ (i.e. after processing data from the source, write to the target) operations.
It's called _transient_ as the flow itself is gone after processing (fire and forget). It doesn't track any history, i.e. it doesn't handle any changes - assume source + processing logic are at a snapshot and process it only once and forget.

To implement it, multiple CocoIndex engine components (in Rust) need to be touched.

Please let me know if you still want to work on this. If so, I can share more details about potential ways to implement it. Thanks!

badmonster0:
Thanks @georgeh0 I'll take it out from hacktoberfest due to the complexity of this issue, and may take long time to resolve. @shresthashim let us know your preference, thanks

[...]

PR #1100, opened 2025-10-02
feat: implemented transient arguement for the coindex.flow_def decorator

shresthashim:
Issue : https://github.com/cocoindex-io/cocoindex/issues/998

I have successfully implemented the transient argument for the @cocoindex.flow_def decorator as requested.

Updated flow_def decorator
Updated open_flow function
Updated _create_lazy_function
Created TransientFlowWrapper class

Now users can create transient flows like this : 
@cocoindex.flow_def(name="MyFlow", transient=True)
def my_flow(flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope):
    # Flow definition that works with inputs and produces outputs
    # but doesn't maintain persistent state
    pass

### DIFF
--- a/python/cocoindex/flow.py
+++ b/python/cocoindex/flow.py
@@ -909,14 +909,192 @@ class Flow:
         return _inner
 
 
+class TransientFlowWrapper(Flow):
+    """
+    A wrapper for transient flows that doesn't support persistence operations.
+    """
+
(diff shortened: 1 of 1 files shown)


badmonster0:
Hi @shresthashim thank you so much for your contribution! We cannot accept this one since it is different than the expected behavior . Detailed explanation here https://github.com/cocoindex-io/cocoindex/issues/998#issuecomment-3359557716 . thanks

PR #1101, opened 2025-10-02
organize targets in documentation

badmonster0:


badmonster0:
https://github.com/cocoindex-io/cocoindex/blob/doc1/docs/docs/targets/index.md

this file probably requires most of the review + routes

docs/docs/targets/index.md
```diff
@@ -0,0 +1,344 @@
...
+
+# CocoIndex Built-in Targets
+
+For each target, data are exported from a data collector, containing data of multiple entries, each with multiple fields.
+The way to map data from a data collector to a target depends on data model of the target.
+
+
+| Target   | Documentation | Type                    |
+|----------|---------------|-------------------------|
+| Postgres | [Postgres](./targets/postgres) | Entry-oriented Targets        |
```
georgeh0: Seems it should be `./postgres`

docs/docs/targets/index.md
```diff
@@ -0,0 +1,344 @@
...
+| Qdrant   | [Qdrant](./targets/qdrant)     | Entry-oriented Targets        |
+| LanceDB  | [LanceDB](./targets/lancedb)   | Entry-oriented Targets        |
+| Neo4j    | [Neo4j](./targets/neo4j)       | [Property graph Targets](#property-graph-targets)        |
```
georgeh0: s/Property graph Targets/Property Graph/

badmonster0:
merged column ty for the suggestion

docs/docs/targets/index.md
```diff
@@ -0,0 +1,344 @@
...
+# CocoIndex Built-in Targets
+
+For each target, data are exported from a data collector, containing data of multiple entries, each with multiple fields.
+The way to map data from a data collector to a target depends on data model of the target.
+
+
+| Target   | Documentation | Type                    |
+|----------|---------------|-------------------------|
+| Postgres | [Postgres](./targets/postgres) | Entry-oriented Targets        |
+| Qdrant   | [Qdrant](./targets/qdrant)     | Entry-oriented Targets        |
```
georgeh0: I'm thinking probably we don't need the "Entry-oriented Targets" term here, as they're not standard terms any way.

Two options:
- Use more specific paradigm names like: Postgres -> Relational Database, LanceDB -> Vector Database, Qdrant -> Vector Database, Neo4j -> Property Graph, Kuzu -> Property Graph
  We don't need sections for Relational / Vector below as they're well-known.
  (but the paradigm for some DBMS are actually arguable, e.g. LanceDB also has relational aspect, and Qdrant/LanceDB can also do keyword search which is beyond vector)

- Don't try to exhaustively put paradigms for everything. Just change the header from "Type" to something like "See Also".

docs/docs/targets/index.md
```diff
@@ -0,0 +1,344 @@
...
+---
+import { ExampleButton } from '../../src/components/GitHubButton';
+
+# CocoIndex Built-in Targets
+
+For each target, data are exported from a data collector, containing data of multiple entries, each with multiple fields.
+The way to map data from a data collector to a target depends on data model of the target.
+
+
+| Documentation    | Type                    |
```
georgeh0: Change this column header to "Target Type"?

badmonster0:
ptal, thanks

https://github.com/cocoindex-io/cocoindex/blob/doc1/docs/docs/targets/index.md

PR #1102, opened 2025-10-02
build(cli-docs): consolidate CLI generation script, md-click didn't work

georgeh0:


Davda-James:
@georgeh0 in this at top subcommands reference should have # heading start and increment additively over then, as current has changed to ## markdownlinter will give warning not satisfying  "MD041/first-line-heading/first-line-h1: First line in a file should be a top-level heading", if we are okay with warnings then its okay but if not we have to change it (according to what we did in [PR](https://github.com/cocoindex-io/cocoindex/pull/1096)

georgeh0:
> @georgeh0 in this at top subcommands reference should have # heading start and increment additively over then, as current has changed to ## markdownlinter will give warning not satisfying "MD041/first-line-heading/first-line-h1: First line in a file should be a top-level heading", if we are okay with warnings then its okay but if not we have to change it (according to what we did in [PR](https://github.com/cocoindex-io/cocoindex/pull/1096)

@Davda-James Thanks for pointing it out!  I demote the levels of headings so that it'll be rendered well after being imoprted by `cli.mdx`. Otherwise it'll look weird.

I'm wondering which specific linter are you using? I'm not seeing this lint problem. We can also consider disable lints for generated files if it's bothering.

Davda-James:
Yes we can, using markdownlint https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint

Davda-James:
Also if we not care about linting it just gave us warnings, we can disable it.

georgeh0:
Thanks for the suggestion! Having #1107

Issue #340, opened 2025-04-18
[FEATURE] collect additional counter for number of rows that are being processed - for stats


badmonster0:
[sources_update_stats](https://github.com/cocoindex-io/cocoindex/blob/774a296f502f6614df07276da104dcf783660d4c/src/execution/live_updater.rs#L13)

The counter keeps track of the following [total stats](https://github.com/cocoindex-io/cocoindex/blob/774a296f502f6614df07276da104dcf783660d4c/src/execution/stats.rs#L55-L59)

We want to add one more counter to keep track of 
1)  num_in_process - total rows that are are being processed on going
2)  in addition, for each [operations](https://cocoindex.io/docs/core/basics#operations), what are the number of rows that are being processed (on going)

Davda-James:
@badmonster0  on it .

PR #1105, opened 2025-10-02
feat: collect additional counter for number of rows that are being processed for stats

Davda-James:
### Description
can now collect additional counter for number of rows that are being processed - for stats

- [x] Ran precommit hooks
- [x] tests written for new counter added 

Written 7 tests status
image

### DIFF
--- a/src/builder/analyzer.rs
+++ b/src/builder/analyzer.rs
@@ -969,13 +969,40 @@ impl AnalyzerContext {
             op_futs.push(self.analyze_reactive_op(op_scope, reactive_op).await?);
         }
         let collector_len = op_scope.states.lock().unwrap().collectors.len();
+        let scope_qualifier = self.build_scope_qualifier(op_scope);
         let result_fut = async move {
             Ok(AnalyzedOpScope {
                 reactive_ops: try_join_all(op_futs).await?,
                 collector_len,
+                scope_qualifier,
             })
         };
         Ok(result_fut)
+    }
+
+    fn build_scope_qualifier(&self, op_scope: &Arc<OpScope>) -> String {
+        let mut scope_names = Vec::new();
+        let mut current_scope = Some(op_scope.as_ref());
+
+        // Walk up the parent chain to collect scope names
+        while let Some(scope) = current_scope {
+            if let Some((parent, _)) = &scope.parent {
+                scope_names.push(scope.name.clone());
+                current_scope = Some(parent.as_ref());
+            } else {
+                break;
+            }
+        }
+
+        // Reverse to get the correct order (root to leaf)
+        scope_names.reverse();
+
+        // Build the qualifier string: "" for root, "name." for single level, "parent.child." for nested
+        if scope_names.is_empty() {
+            String::new()
+        } else {
+            format!("{}.", scope_names.join("."))
+        }
     }
 }
 
--- a/src/builder/analyzer.rs
+++ b/src/builder/analyzer.rs
@@ -982,13 +982,13 @@ impl AnalyzerContext {
 
     fn build_scope_qualifier(&self, op_scope: &Arc<OpScope>) -> String {
         let mut scope_names = Vec::new();
-        let mut current_scope = Some(op_scope.as_ref());
+        let mut current_scope = op_scope.as_ref();
 
         // Walk up the parent chain to collect scope names
-        while let Some(scope) = current_scope {
-            if let Some((parent, _)) = &scope.parent {
-                scope_names.push(scope.name.clone());
-                current_scope = Some(parent.as_ref());
+        loop {
+            if let Some((parent, _)) = &current_scope.parent {
+                scope_names.push(current_scope.name.clone());
+                current_scope = parent.as_ref();
             } else {
                 break;
             }
--- a/src/builder/analyzer.rs
+++ b/src/builder/analyzer.rs
@@ -985,24 +985,21 @@ impl AnalyzerContext {
         let mut current_scope = op_scope.as_ref();
 
         // Walk up the parent chain to collect scope names
-        loop {
-            if let Some((parent, _)) = &current_scope.parent {
-                scope_names.push(current_scope.name.clone());
-                current_scope = parent.as_ref();
-            } else {
-                break;
-            }
+        while let Some((parent, _)) = &current_scope.parent {
+            scope_names.push(current_scope.name.as_str());
+            current_scope = parent.as_ref();
         }
 
         // Reverse to get the correct order (root to leaf)
         scope_names.reverse();
 
-        // Build the qualifier string: "" for root, "name." for single level, "parent.child." for nested
-        if scope_names.is_empty() {
-            String::new()
-        } else {
-            format!("{}.", scope_names.join("."))
+        // Build the qualifier string
+        let mut result = String::new();
+        for name in scope_names {
+            result.push_str(&name);
+            result.push('.');
         }
+        result
     }
 }
 
--- a/src/builder/analyzer.rs
+++ b/src/builder/analyzer.rs
@@ -1059,6 +1059,7 @@ pub async fn analyze_flow(
         let target_factory = get_target_factory(&target_kind)?;
         let analyzed_target_op_group = AnalyzedExportTargetOpGroup {
             target_factory,
+            target_kind: target_kind.clone(),
             op_idx: op_ids.export_op_ids,
         };
         export_ops_futs.extend(
--- a/src/execution/evaluator.rs
+++ b/src/execution/evaluator.rs
@@ -322,7 +322,7 @@ async fn evaluate_child_op_scope(
     child_scope_entry: ScopeEntry<'_>,
     concurrency_controller: &concur_control::ConcurrencyController,
     memory: &EvaluationMemory,
-    operation_in_process_stats: Option<&crate::execution::stats::OperationInProcessStats>,
+    operation_in_process_stats: Option<&execution::stats::OperationInProcessStats>,
 ) -> Result<()> {
     let _permit = concurrency_controller
         .acquire(Some(|| {
@@ -356,7 +356,7 @@ async fn evaluate_op_scope(
     op_scope: &AnalyzedOpScope,
     scoped_entries: RefList<'_, &ScopeEntry<'_>>,
     memory: &EvaluationMemory,
-    operation_in_process_stats: Option<&crate::execution::stats::OperationInProcessStats>,
+    operation_in_process_stats: Option<&execution::stats::OperationInProcessStats>,
 ) -> Result<()> {
     let head_scope = *scoped_entries.head().unwrap();
     for reactive_op in op_scope.reactive_ops.iter() {
@@ -364,7 +364,8 @@ async fn evaluate_op_scope(
             AnalyzedReactiveOp::Transform(op) => {
                 // Track transform operation start
                 if let Some(ref op_stats) = operation_in_process_stats {
-                    op_stats.start_processing(&op.name, 1);
+                    let transform_key = format!("transform/{}", op.name);
+                    op_stats.start_processing(&transform_key, 1);
                 }
 
                 let mut input_values = Vec::with_capacity(op.inputs.len());
@@ -399,7 +400,8 @@ async fn evaluate_op_scope(
 
                 // Track transform operation completion
                 if let Some(ref op_stats) = operation_in_process_stats {
-                    op_stats.finish_processing(&op.name, 1);
+                    let transform_key = format!("transform/{}", op.name);
+                    op_stats.finish_processing(&transform_key, 1);
                 }
 
                 result?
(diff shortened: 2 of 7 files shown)


src/execution/stats.rs
```diff
@@ -64,6 +64,8 @@ pub struct UpdateStats {
     /// Number of source rows that were reprocessed because of logic change.
     pub num_reprocesses: Counter,
     pub num_errors: Counter,
+    /// Number of source rows currently being processed.
+    pub num_in_process: Counter,
```
georgeh0: There's some nuance here. `Counter` is intended to be cumulative (e.g. operations like `merge` and `delta` are designed based on this assumption, e.g. with the stats now and 1 minutes ago, we do a "delta" will get the progress within the past minute). 
But `num_in_process` is not cumulative.

One way to fix is use a pair of cumulative counters, `num_processing_starts` and `num_processing_ends`.  When we want to know num of rows in-processing, just do a `num_processing_starts - num_processing_ends` (and this also provides additional information, i.e. we'll know the overall processing count).

The same for `OperationInProcessStats` below.

I will consider create another struct wrapping these two counters like the following and reuse between two places.

```rs
struct ProcessingCounters {
  pub num_starts;
  pub num_ends;
}
```

src/execution/source_indexer.rs
```diff
@@ -317,15 +317,31 @@ impl SourceIndexingContext {
         row_input: ProcessSourceRowInput,
         mode: UpdateMode,
         update_stats: Arc<stats::UpdateStats>,
+        operation_in_process_stats: Option<Arc<stats::OperationInProcessStats>>,
         _concur_permit: concur_control::CombinedConcurrencyControllerPermit,
         ack_fn: Option<AckFn>,
         pool: PgPool,
     ) {
+        // Store operation name for tracking cleanup
+        let operation_name = {
```
georgeh0, line 326: For operation level counters, in `source_indexer` we don't have information about specific operations. We need to go deeper to collect them (it's OK to create the operation level stats in this PR and add code to collect them in a separate PR too)

These're the places running specific operations:

- for import op: https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/src/execution/source_indexer.rs#L402-L416
- for transform op: https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/src/execution/evaluator.rs#L357-L386
- for export op: https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/src/execution/row_indexer.rs#L371-L373

src/execution/stats.rs
Davda-James: okay working on this

Davda-James: @georgeh0 made change in latest commit (change only includes this review, for below one will add new commit), if any change required or if I missed out something pls tell will fix it.

src/execution/source_indexer.rs
Davda-James: Will add this changes to new commit in same PR.

src/execution/stats.rs
```diff
@@ -54,6 +54,55 @@ impl std::fmt::Debug for Counter {
...
+    }
+
+    /// End processing the specified number of items.
+    pub fn end(&self, count: i64) {
+        self.num_ends.inc(count);
+    }
+
+    /// Get the current number of items being processed (starts - ends).
+    pub fn get_in_process(&self) -> i64 {
+        self.num_starts.get() - self.num_ends.get()
```
georgeh0: One nuance is, we may read `end` first (save in a local variable), then read `start`. Otherwise, in rare cases `end` may even be larger than `start` (consider a super quick operation where `end` is increased right after `start` is increased).

src/execution/stats.rs
```diff
@@ -54,6 +54,55 @@ impl std::fmt::Debug for Counter {
...
+
+    /// Get the total number of processing operations started.
+    pub fn get_total_starts(&self) -> i64 {
+        self.num_starts.get()
+    }
+
+    /// Get the total number of processing operations ended.
+    pub fn get_total_ends(&self) -> i64 {
+        self.num_ends.get()
+    }
```
georgeh0, line 81: I'm not a big fan of having these methods.

`num_starts` is already public and users can already do `processing.num_starts.get()` conveniently without having a wrapper.

Same for `start_processing()` and `end_processing()` in `UpdateStats` below.

src/execution/source_indexer.rs
Davda-James: Added this and that unrequired wrapper functions, start- ends negative fix into latest commit .

src/execution/evaluator.rs
```diff
@@ -322,6 +322,7 @@ async fn evaluate_child_op_scope(
     child_scope_entry: ScopeEntry<'_>,
     concurrency_controller: &concur_control::ConcurrencyController,
     memory: &EvaluationMemory,
+    operation_in_process_stats: Option<&crate::execution::stats::OperationInProcessStats>,
```
georgeh0: We don't need the `crate::` prefix (`execution` already brought in from the `prelude`)

src/execution/evaluator.rs
```diff
@@ -333,32 +334,44 @@ async fn evaluate_child_op_scope(
...
+            }
+        )
+    })
 }
 
 async fn evaluate_op_scope(
     op_scope: &AnalyzedOpScope,
     scoped_entries: RefList<'_, &ScopeEntry<'_>>,
     memory: &EvaluationMemory,
+    operation_in_process_stats: Option<&crate::execution::stats::OperationInProcessStats>,
```
georgeh0: Same here, no need `crate::`

src/execution/live_updater.rs
```diff
@@ -406,6 +415,23 @@ impl FlowLiveUpdater {
...
+        self.operation_in_process_stats
+            .get_operation_in_process_count(operation_name)
+    }
+
+    /// Get a snapshot of all operation in-process counts.
+    pub fn get_all_operations_in_process(&self) -> std::collections::HashMap<String, i64> {
+        self.operation_in_process_stats
+            .get_all_operations_in_process()
+    }
+
```
georgeh0: I think we can simply make `operation_in_process_stats` public. Then these 3 helper methods are not necessary.

src/execution/row_indexer.rs
```diff
@@ -368,9 +375,36 @@ impl<'a> RowIndexer<'a> {
...
-                        export_op_group
-                            .target_factory
-                            .apply_mutation(mutations_w_ctx)
+                        // Track export operation start
+                        if let Some(ref op_stats) = self.operation_in_process_stats {
+                            for export_op_idx in &export_op_group.op_idx {
+                                let export_op = &self.src_eval_ctx.plan.export_ops[*export_op_idx];
+                                op_stats.start_processing(&export_op.name, 1);
+                            }
+                        }
```
georgeh0: This blends all targets together, and is not what we want.

For targets, since `apply_mutation` happens per "export op group". So we can also make the counters per export op group. Each group can be identified by the "target kind": we can add it as a field [here](https://github.com/cocoindex-io/cocoindex/blob/069f4a1f56613d0d66ee3245888b6c70fefd7d61/src/builder/plan.rs#L118-L121), then we can easily get the group name here.

To distinguish we're using a target kind rather than op name here, we can add a prefix for each key in the map for what it's about, e.g. `import/{op_name}`, `transform/{op_name}`, `export/{target_kind}`

src/execution/live_updater.rs
Davda-James: done

src/execution/row_indexer.rs
Davda-James: Thanks @georgeh0 learning a lot. Did this.

Davda-James: If not then will push above review and this one changes in a new commit.

src/execution/row_indexer.rs
```diff
@@ -368,9 +375,28 @@ impl<'a> RowIndexer<'a> {
                         })
                         .collect();
                     (!mutations_w_ctx.is_empty()).then(|| {
-                        export_op_group
-                            .target_factory
-                            .apply_mutation(mutations_w_ctx)
+                        // Track export operation start
+                        if let Some(ref op_stats) = self.operation_in_process_stats {
+                            let export_key = format!("export/{}", export_op_group.target_kind);
+                            op_stats.start_processing(&export_key, 1);
```
georgeh0: Consider moving this into the async closure below - more symmetric with the `finish_processing` part and code will also be simpler.

src/execution/source_indexer.rs
```diff
@@ -474,7 +492,15 @@ impl SourceIndexingContext {
             result
         };
         let process_and_ack = async {
-            process.await?;
+            let result = process.await;
+
+            // Track that we're finishing processing this row (regardless of success/failure)
+            update_stats.processing.end(1);
+            if let Some(ref op_stats) = operation_in_process_stats {
+                op_stats.finish_processing(&operation_name, 1);
```
georgeh0: This tracking range is over broad. For "import/*", we should only track on lifespan of `get_value()` call (call "start" before `get_value()` starts, and call "finish" after it returns)

https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/src/execution/source_indexer.rs#L402-L416

src/execution/evaluator.rs
```diff
@@ -333,32 +334,45 @@ async fn evaluate_child_op_scope(
...
     memory: &EvaluationMemory,
+    operation_in_process_stats: Option<&execution::stats::OperationInProcessStats>,
 ) -> Result<()> {
     let head_scope = *scoped_entries.head().unwrap();
     for reactive_op in op_scope.reactive_ops.iter() {
         match reactive_op {
             AnalyzedReactiveOp::Transform(op) => {
+                // Track transform operation start
+                if let Some(ref op_stats) = operation_in_process_stats {
+                    let transform_key = format!("transform/{}", op.name);
```
georgeh0: The `op.name` here is actually a name local to the current parent scope. We want to qualify it with name of its parent scopes.

I'm proposing adding a `scope_qualifier` field to [`AnalyzedOpScope`](https://github.com/cocoindex-io/cocoindex/blob/ac786d79598d70b04ef6c158aacde49898faeaac/src/builder/plan.rs#L129-L132) and it can be set in the analyzer [here](https://github.com/cocoindex-io/cocoindex/blob/main/src/builder/analyzer.rs#L976), e.g. its value can be `""` (for the root scope), `"file."` or `"file.chunk."`. Then we can use `format!("transform/{}{}", op_scope.full_name, op.name)` here.

src/execution/row_indexer.rs
Davda-James: Done

src/execution/evaluator.rs
Davda-James: did it, pls review if any  change required in it

src/builder/analyzer.rs
```diff
@@ -969,14 +969,41 @@ impl AnalyzerContext {
...
                 collector_len,
+                scope_qualifier,
             })
         };
         Ok(result_fut)
     }
+
+    fn build_scope_qualifier(&self, op_scope: &Arc<OpScope>) -> String {
+        let mut scope_names = Vec::new();
+        let mut current_scope = Some(op_scope.as_ref());
```
georgeh0: `current_scope` is never assigned with a `None` value. So it doesn't need to be an `Option`.

src/builder/analyzer.rs
```diff
@@ -969,14 +969,41 @@ impl AnalyzerContext {
...
         Ok(result_fut)
     }
+
+    fn build_scope_qualifier(&self, op_scope: &Arc<OpScope>) -> String {
+        let mut scope_names = Vec::new();
+        let mut current_scope = op_scope.as_ref();
+
+        // Walk up the parent chain to collect scope names
+        loop {
+            if let Some((parent, _)) = &current_scope.parent {
```
georgeh0: We can use `while let Some((parent, _)) = &current_scope.parent { ... ` here.

src/builder/analyzer.rs
```diff
@@ -969,14 +969,41 @@ impl AnalyzerContext {
...
+        }
+
+        // Reverse to get the correct order (root to leaf)
+        scope_names.reverse();
+
+        // Build the qualifier string: "" for root, "name." for single level, "parent.child." for nested
+        if scope_names.is_empty() {
+            String::new()
+        } else {
+            format!("{}.", scope_names.join("."))
```
georgeh0: nit: `scope_names.join(".") + "."` can be slightly more efficient

another way (if you like) is doing this, which have a benefit of avoiding empty as special case:

```rs
let mut result = String::new();
for name in scope_names {
  result.push_str(name);
  result.push('.');
}
result
```

src/builder/analyzer.rs
```diff
@@ -969,14 +969,41 @@ impl AnalyzerContext {
...
     }
+
+    fn build_scope_qualifier(&self, op_scope: &Arc<OpScope>) -> String {
+        let mut scope_names = Vec::new();
+        let mut current_scope = op_scope.as_ref();
+
+        // Walk up the parent chain to collect scope names
+        loop {
+            if let Some((parent, _)) = &current_scope.parent {
+                scope_names.push(current_scope.name.clone());
```
georgeh0: We don't need a `clone()` here. Use `current_scope.name.as_str()`.

src/builder/analyzer.rs
Davda-James: done

src/builder/analyzer.rs
Davda-James: done

georgeh0 (approved):
Awesome! Thanks for this PR!

georgeh0:
Fixes #340

Issue #1115, opened 2025-10-03
[FEATURE] Split `convert.py` into 2 modules


georgeh0:
The current `convert.py` has 2 types of things:

1. Encode/decode values in cocoindex (for data)
2. Dump/load objects (for configs, specs)

We want to split them into two different modules: `engine_value` and `engine_object`.

Their tests should also be split accordingly.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

Shubham15986:
I’m participating in Hacktoberfest and would love to contribute to this project.
Could you please assign this issue to me?

Davda-James:
already working on it @Shubham15986 missed to comment.

Davda-James:
Hey @georgeh0 , like I was thinking after splitting current module about test, like should we remove test_convert and test_load and put them accordingly into two test_engine_value and test_engine_object

georgeh0:
> Hey [@georgeh0](https://github.com/georgeh0) , like I was thinking after splitting current module about test, like should we remove test_convert and test_load and put them accordingly into two test_engine_value and test_engine_object

Yes, this sounds good!

badmonster0:
@Davda-James assigned it to you. Please make a comment for future issues and wait for assign :) we follow the chronological order to assign tickets, thanks! btw love your contributions, thanks!

Davda-James:
Yes generally I comment it before started working on issue, missed for this one, will take care. Thanks

PR #1120, opened 2025-10-03
feat: split convert.py into 2 modules engine_object and engine_value along with splitting of tests

Davda-James:
split convert.py into 2 modules engine_object and engine_value , also split current tests into two test files test_engine_object and test_engine_value.

- [x] Ran precommit hooks 
- [x] Ran tests after splitting
- [x] Updated old imports (from cocoindex.convert.. to match with updated modules) 
image

### DIFF
--- a/python/cocoindex/engine_object.py
+++ b/python/cocoindex/engine_object.py
@@ -1,0 +1,272 @@
+"""
+Utilities to dump/load objects (for configs, specs).
+"""
+
+from __future__ import annotations
+
+import datetime
+import dataclasses
+from enum import Enum
+from typing import Any, Mapping, TypeVar, overload, get_origin
+
+import numpy as np
+
+from .typing import (
+    AnalyzedAnyType,
+    AnalyzedBasicType,
+    AnalyzedDictType,
+    AnalyzedListType,
+    AnalyzedStructType,
+    AnalyzedTypeInfo,
+    AnalyzedUnionType,
+    EnrichedValueType,
+    FieldSchema,
+    analyze_type_info,
+    encode_enriched_type,
+    is_namedtuple_type,
+    is_pydantic_model,
+    extract_ndarray_elem_dtype,
+)
+
+
+T = TypeVar("T")
+
+try:
+    import pydantic, pydantic_core
+except ImportError:
+    pass
+
+
+def _get_auto_default_for_type(
+    type_info: AnalyzedTypeInfo,
+) -> tuple[Any, bool]:
+    """
+    Get an auto-default value for a type annotation if it's safe to do so.
+
+    Returns:
+        A tuple of (default_value, is_supported) where:
+        - default_value: The default value if auto-defaulting is supported
+        - is_supported: True if auto-defaulting is supported for this type
+    """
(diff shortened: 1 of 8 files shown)


python/cocoindex/engine_object.py
```diff
@@ -0,0 +1,272 @@
...
+    if type_info.nullable:
+        return None, True
+
+    # Case 2: Table types (KTable or LTable) - check if it's a list or dict type
+    if isinstance(type_info.variant, AnalyzedListType):
+        return [], True
+    elif isinstance(type_info.variant, AnalyzedDictType):
+        return {}, True
+
+    return None, False
```
georgeh0: Now this one is duplicated between two modules. I think we can remove the leading underscore in its name, and modify `engine_value.py` to call this one instead.

Davda-James: okay on it

Davda-James: Done mistakenly same was left in one module, now kept it in engine_value and removed leading underscore in name and called it in engine_object.py

Davda-James: Done @georgeh0 kept the get_auto_default_for_type in engine_object and imported the same in engine_value. (removed leading underscore)

python/cocoindex/tests/test_engine_value.py
```diff
@@ -1006,28 +1005,28 @@ def test_decode_error_non_nullable_or_non_list_vector() -> None:
...
+# def test_dump_vector_type_annotation_no_dim() -> None:
+#     """Test dumping a vector type annotation with no dimension."""
+#     expected_dump_no_dim = {
+#         "type": {
+#             "kind": "Vector",
+#             "element_type": {"kind": "Float64"},
+#             "dimension": None,
+#         }
+#     }
+#     assert dump_engine_object(Float64VectorTypeNoDim) == expected_dump_no_dim
```
georgeh0: These tests should be kept and carried over to the other file.

Davda-James: So we should move it to test_engine_object right ?

Davda-James: Also Float64VectorTypeNoDim, Float32VectorType, Float64VectorType are defined in test_engine_value should we import that or define them again in test_engine_object ?

georgeh0: > So we should move it to test_engine_object right ?

Yes, it tests functions defined in `engine_object.py`, so the tests belong there.

> Also Float64VectorTypeNoDim, Float32VectorType, Float64VectorType are defined in test_engine_value should we import that or define them again in test_engine_object ?

No, we don't need to. They're ad-hoc type aliases and used in tests only. We don't have to introduce cross-module dependency for them (and they only provide minimum level of convenience, so we can even removing them and use these types directly)

Davda-James: Done used inline directly without any alias and cross deps

georgeh0 (approved):
Thanks!

georgeh0:
This fixes #1115

PR #1138, opened 2025-10-04
Feat/postgres schema support

### LINKED ISSUE #722: [FEATURE] Allow users to specify target schema for Postgres targets
Postgres support [schema](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS) to cluster related tables. We want to add a `schema` field to the spec for [`Postgres`](https://cocoindex.io/docs/ops/targets#spec) target. When `schema` is specified, `table_name` must also be explicitly specified.

~CocoIndex doesn't own schemas, i.e. CocoIndex doesn't create/drop these schemas (targets from different flows can share the same schema, so it's over complex if CocoIndex needs to manage them). They need to be created/dropped by users, similar to databases.~

**Update**: CocoIndex automatically create the schema *if not exists* (`CREATE SCHEMA IF NOT EXISTS`) before doing the table creation. CocoIndex doesn't own the schema, i.e. it doesn't drop the schema: targets from different flows can share the same schema, and users are free to use the same schema for both cocoindex-managed tables and other tables, so it's not easy for CocoIndex to manage the schema drop.

---

thisisharsh7:
fixes #722 

feature implemented ->                                                                                   

- Added schema field to spec for Postgres target
- When schema is specified, table_name must also be explicitly specified
- automatically create the schema if not exists
- it doesn't drop the schema
- targets from different flows can share the same schema
- Schema is created but not owned, users can add their own tables

### DIFF
--- a/src/ops/targets/postgres.rs
+++ b/src/ops/targets/postgres.rs
@@ -18,6 +18,7 @@ use std::ops::Bound;
 pub struct Spec {
     database: Option<spec::AuthEntryReference<DatabaseConnectionSpec>>,
     table_name: Option<String>,
+    schema: Option<String>,
 }
 const BIND_LIMIT: usize = 65535;
 
@@ -143,10 +144,12 @@ impl ExportContext {
     fn new(
         db_ref: Option<spec::AuthEntryReference<DatabaseConnectionSpec>>,
         db_pool: PgPool,
-        table_name: String,
+        table_id: &TableId,
         key_fields_schema: Box<[FieldSchema]>,
         value_fields_schema: Vec<FieldSchema>,
     ) -> Result<Self> {
+        let table_name = qualified_table_name(table_id);
+
         let key_fields = key_fields_schema
             .iter()
             .map(|f| format!("\"{}\"", f.name))
@@ -255,12 +258,18 @@ pub struct Factory {}
 pub struct TableId {
     #[serde(skip_serializing_if = "Option::is_none")]
     database: Option<spec::AuthEntryReference<DatabaseConnectionSpec>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    schema: Option<String>,
     table_name: String,
 }
 
 impl std::fmt::Display for TableId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.table_name)?;
+        if let Some(schema) = &self.schema {
+            write!(f, "{}.{}", schema, self.table_name)?;
+        } else {
+            write!(f, "{}", self.table_name)?;
+        }
         if let Some(database) = &self.database {
             write!(f, " (database: {database})")?;
         }
@@ -342,6 +351,13 @@ fn to_column_type_sql(column_type: &ValueType) -> String {
             BasicValueType::Union(_) => "jsonb".into(),
         },
         _ => "jsonb".into(),
+    }
+}
+
+fn qualified_table_name(table_id: &TableId) -> String {
+    match &table_id.schema {
+        Some(schema) => format!("\"{}\".\"{}\"", schema, table_id.table_name),
+        None => format!("\"{}\"", table_id.table_name),
     }
 }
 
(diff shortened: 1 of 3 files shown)


thisisharsh7:
please let me know for any changes thanks!!

src/ops/targets/postgres.rs
```diff
@@ -345,6 +354,13 @@ fn to_column_type_sql(column_type: &ValueType) -> String {
     }
 }
 
+fn qualified_table_name(table_id: &TableId) -> String {
+    match &table_id.schema {
+        Some(schema) => format!("\"{}\".\"{}\"", schema, table_id.table_name),
+        None => format!("\"{}\"", table_id.table_name),
```
georgeh0: Let's leave table name unquoted for this case for now.  Only double-qutes schema names. 

We considered adding double-quotes for table names before, but didn't proceed because it'll break existing pipelines. We'll make this change in our next major version upgrade (e.g. v1) which is considered to be backward imcompatible.

src/ops/targets/postgres.rs
```diff
@@ -554,7 +570,15 @@ impl setup::ResourceSetupChange for SetupChange {
...
 
 impl SetupChange {
-    async fn apply_change(&self, db_pool: &PgPool, table_name: &str) -> Result<()> {
+    async fn apply_change(&self, db_pool: &PgPool, table_id: &TableId) -> Result<()> {
+        // Create schema if specified
+        if let Some(schema) = &table_id.schema {
+            let sql = format!("CREATE SCHEMA IF NOT EXISTS \"{}\"", schema);
+            sqlx::query(&sql).execute(db_pool).await?;
+        }
+
```
georgeh0: We only need to do it when we need to run `CREATE TABLE ...`. We can move it before executing that statement.

georgeh0:
Can you also describe how you test it end to end? Thanks!

src/ops/targets/postgres.rs
thisisharsh7: okay got it

thisisharsh7:
> Can you also describe how you test it end to end? Thanks!

thanks! I tested this feature with local postgres server:
- with both schema and table name -> it works
- with just schema no table name -> it works too and without schema at all

for database verification:
- created a schema using `CREATE SCHEMA IF NOT EXISTS test_schema`
- verified the schema in the database
- created a table inside the schema
- inserted and queried data successfully

thisisharsh7:
image

georgeh0:
> image

Seems these tests happen on single statements, instead of as part of cocoindex flow?
Did you test it end to end with CocoIndex? For example, pick any existing example with Postgres target, and set the schema, and run setup and update.

src/ops/targets/postgres.rs
```diff
@@ -18,6 +18,7 @@ use std::ops::Bound;
 pub struct Spec {
     database: Option<spec::AuthEntryReference<DatabaseConnectionSpec>>,
     table_name: Option<String>,
+    schema: Option<String>,
```
georgeh0, line 21: I think at least the field also needs to be added to this dataclass in Python SDK:

https://github.com/cocoindex-io/cocoindex/blob/afea19dac2364d9da31860b1f19f7f435ca4be00/python/cocoindex/targets/_engine_builtin_specs.py#L12-L16

So that users will be able to use it.

thisisharsh7: okay will do

thisisharsh7:
```
    database: AuthEntryReference[DatabaseConnectionSpec] | None = None
    table_name: str | None = None
    schema: str | None = None
```

thisisharsh7:
image

can you please check this if this is correct thanks!

thisisharsh7:
I picked text-embedding example and put the schema then ran the setup nd update cmd

georgeh0:
> Can you also help with updating this doc: https://github.com/cocoindex-io/cocoindex/blob/main/docs/docs/sources/index.md?plain=1
> 
> Thanks!

Thanks! Looks good! Can you also help update the doc in the location above? Thanks!

thisisharsh7:
yes

thisisharsh7:
what should I include also do I need to update target file? some pointers

georgeh0:
> what should I include also do I need to update target file? some pointers

Sorry I gave the wrong link. It should be for the target, not for the source: https://github.com/cocoindex-io/cocoindex/blob/main/docs/docs/targets/postgres.md?plain=1

The description for new `schema` field in the spec should be quite similar to `table_name`.

thisisharsh7:
image

please check this is this right

georgeh0:
> image
> please check this is this right

Looks good. Please push the change. Thanks!

georgeh0:
Looks good! Thanks for the PR. Merging.

PR #1153, opened 2025-10-06
feat: collector automatically merge and align multiple collect() called with different schema

### LINKED ISSUE #428: [FEATURE] collector automatically merge and align multiple `collect()` called with different schema
e.g. 

```
collector.collect(field1=x, field2=y);
collector.collect(field1=x, field3=z);
```

The schema for the collector should have `field1`, `field2` and `field3`.

During execution, we should also be able to align fields and fill null values for missing fields, and get a schema with fields `field1`, `field2` and `field3`.

skalwaghe-56:
Implemented automatic schema merging for collectors that:
- Unions fields from multiple collect() calls instead of requiring identical schemas
- Fills missing fields with null values during execution
- Maintains consistent field ordering (alphabetically sorted by field name)
- Preserves auto-generated UUID fields across merged schemas

Closes #428.


src/execution/evaluator.rs
```diff
@@ -483,26 +483,50 @@ async fn evaluate_op_scope(
...
+                for (i, field_name) in op.input_field_names.iter().enumerate() {
+                    if let Some(pos) = op
+                        .collector_schema
+                        .fields
+                        .iter()
+                        .position(|f| &f.name == field_name)
+                    {
+                        field_values[pos] = input_values[i].clone();
+                    }
+                }
```
georgeh0: For each field we iterate over fields in `collector_schema`, which is a O(N^2) string comparison - on each row of data. It has unnecessary cost.

We only need to compute the index mappings between input fields index and collector fields index only once, in the analyzer, so in the valuator we only need simple fields copy or move.

src/execution/evaluator.rs
```diff
@@ -483,26 +483,50 @@ async fn evaluate_op_scope(
...
+                        let uuid = memory.next_uuid(
+                            op.fingerprinter
+                                .clone()
+                                .with(
+                                    &field_values
+                                        .iter()
+                                        .enumerate()
+                                        .filter(|(i, _)| *i != uuid_idx)
+                                        .map(|(_, v)| v)
+                                        .collect::<Vec<_>>(),
```
georgeh0, line 545: We shouldn't need to create another `Vec` for this purpose. I think we can simply keep the original property: uuid field always at position 0 if exists. Then we don't need to modify the way to create uuid.

src/execution/evaluator.rs
skalwaghe-56: Resolved in the latest push!

src/execution/evaluator.rs
skalwaghe-56: Resolved in the latest push!

src/builder/analyzer.rs
```diff
@@ -255,14 +255,71 @@ fn try_merge_collector_schemas(
...
+    fields.sort_by(|a, b| {
+        let a_is_uuid = matches!(a.value_type.typ, ValueType::Basic(BasicValueType::Uuid));
+        let b_is_uuid = matches!(b.value_type.typ, ValueType::Basic(BasicValueType::Uuid));
+
+        match (a_is_uuid, b_is_uuid) {
+            (true, false) => std::cmp::Ordering::Less, // UUID fields first
+            (false, true) => std::cmp::Ordering::Greater, // UUID fields first
+            _ => a.name.cmp(&b.name),                  // Then alphabetical
+        }
+    });
```
georgeh0: I don't quite like sorting fields. I think we want to preserve the original order if possible. We only need to add one restriction - if the same field appears on both, they must have consistent in ordering (otherwise raise an error). Then we can merge them without changing the order.

One possible merging approach is (pseudo code, only to show the gist):

```
let mut output_fields = vec![];
let next_field_id_1 = 0;
let next_field_id_2 = 0;
for (idx, field) in schema2.iter().enumerate() {
  if Some(idx1) = field index in schema1 {
    if (idx1 < next_field_id_1) {
      api_bail!("order mismatch...");
    }
    output_fields.extend(schema1.fields[next_field_id_1..idx1]);
    output_fields.extend(schema2.fields[next_field_id_2..idx]);
    output_fields.push(merged field);
    next_field_id_1 = idx1 + 1;
    next_field_id_2 = idx + 1;
  } else if field is uuid {
    // For UUID, emit it immediately to make sure it still appears first
    ....
  }
}
next_field_id_1.extend(schema1.fields[next_field_id_1..]);
next_field_id_2.extend(schema2.fields[next_field_id_2..]);
```

skalwaghe-56: Please check now.

src/builder/analyzer.rs
```diff
@@ -255,14 +255,84 @@ fn try_merge_collector_schemas(
...
+    // Handle auto_uuid_field_idx
+    let auto_uuid_field_idx = match (schema1.auto_uuid_field_idx, schema2.auto_uuid_field_idx) {
+        (Some(idx1), Some(idx2)) => {
+            let name1 = &schema1_fields[idx1].name;
+            let name2 = &schema2_fields[idx2].name;
+            if name1 == name2 {
+                // Find the position of the auto_uuid field in the merged output
+                output_fields.iter().position(|f| &f.name == name1)
+            } else {
+                None // Different auto_uuid fields, disable
```
georgeh0: Let's raise an explicit error in this case.

src/builder/analyzer.rs
```diff
@@ -255,14 +255,84 @@ fn try_merge_collector_schemas(
...
+            }
+        }
+        _ => None, // If either doesn't have it, or both don't, disable
```
georgeh0: If one side has, but the other doesn't, let's also raise an error.

src/builder/analyzer.rs
```diff
@@ -255,14 +255,84 @@ fn try_merge_collector_schemas(
...
+    let mut next_field_id_2 = 0;
+
+    for (idx, field) in schema2_fields.iter().enumerate() {
+        if let Some(&idx1) = field_map.get(&field.name) {
+            if idx1 < next_field_id_1 {
+                api_bail!(
+                    "Order mismatch for field '{}': appears at position {} in first schema but expected after {}",
+                    field.name,
+                    idx1,
+                    next_field_id_1
```
georgeh0: The field ID is not easy to read by users. I think we should put another field name.

Once this error happens, it means `field.name` and `schema1_fields[next_field_id_1 - 1].name` appear in different orders from both schema.

src/builder/analyzer.rs
```diff
@@ -255,14 +255,84 @@ fn try_merge_collector_schemas(
...
+                name: field.name.clone(),
+                value_type: merged_type,
+                description: None,
+            });
+            next_field_id_1 = idx1 + 1;
+            next_field_id_2 = idx + 1;
+        } else if matches!(field.value_type.typ, ValueType::Basic(BasicValueType::Uuid)) {
+            // For UUID, emit it immediately to ensure it appears first
+            output_fields.push(field.clone());
+            next_field_id_2 = idx + 1;
```
georgeh0: Actually this case is "disabled" later any way (and we may actually raise an error nevertheless). So I think we don't really need the branch.

src/builder/analyzer.rs
```diff
@@ -803,16 +873,42 @@ impl AnalyzerContext {
...
+                };
+
+                // Pre-compute field index mappings for efficient evaluation
+                let field_index_mapping: Vec<usize> = input_field_names
+                    .iter()
+                    .map(|field_name| {
+                        collector_schema
+                            .fields
+                            .iter()
+                            .position(|f| &f.name == field_name)
```
georgeh0: This can have O(n^2) time complexity and can be a potential problem when there're a lot of fields. So we may want to build map first before running this.

src/builder/analyzer.rs
skalwaghe-56: Please check latest push. Thanks!

src/builder/analyzer.rs
skalwaghe-56: Please check latest push. Thanks!

src/builder/analyzer.rs
skalwaghe-56: Please check latest push. Thanks!

src/builder/analyzer.rs
skalwaghe-56: Please check latest push. Thanks!

src/builder/analyzer.rs
skalwaghe-56: Please check latest push. Thanks!

skalwaghe-56:
@georgeh0 Please check now and get this merged. Thanks!

src/builder/analyzer.rs
```diff
@@ -255,14 +255,87 @@ fn try_merge_collector_schemas(
...
+    let auto_uuid_field_idx = match (schema1.auto_uuid_field_idx, schema2.auto_uuid_field_idx) {
+        (Some(idx1), Some(idx2)) => {
+            let name1 = &schema1_fields[idx1].name;
+            let name2 = &schema2_fields[idx2].name;
+            if name1 == name2 {
+                // Find the position of the auto_uuid field in the merged output
+                output_fields.iter().position(|f| &f.name == name1)
+            } else {
+                return api_bail!(
+                    "Different auto_uuid fields are not allowed: '{}' vs '{}'",
```
georgeh0: Let's use this error message to be more clear:

```
"Generated UUID fields must have the same name across different `collect()` calls, got different names: '{}' vs '{}'"
```

src/builder/analyzer.rs
```diff
@@ -255,14 +255,87 @@ fn try_merge_collector_schemas(
...
+
+    let mut output_fields = Vec::new();
+    let mut next_field_id_1 = 0;
+    let mut next_field_id_2 = 0;
+
+    for (idx, field) in schema2_fields.iter().enumerate() {
+        if let Some(&idx1) = field_map.get(&field.name) {
+            if idx1 < next_field_id_1 {
+                api_bail!(
+                    "Order mismatch for field '{}': appears at position {} in first schema but expected after '{}'",
```
georgeh0: Let's use this error message to be more clear:

```
"Common fields are expected to have consistent order across different `collect()` calls, but got different orders between fields '{}' and '{}'"
```

src/builder/analyzer.rs
```diff
@@ -255,14 +255,87 @@ fn try_merge_collector_schemas(
...
+            } else {
+                return api_bail!(
+                    "Different auto_uuid fields are not allowed: '{}' vs '{}'",
+                    name1,
+                    name2
+                );
+            }
+        }
+        (Some(_), None) | (None, Some(_)) => {
+            return api_bail!("Inconsistent auto_uuid field presence is not allowed");
```
georgeh0: Let's use this error message to be more clear:

```
"The generated UUID field, once present for one `collect()`, must be consistently present for other `collect()` calls for the same collector"
```

src/builder/analyzer.rs
```diff
@@ -803,16 +876,47 @@ impl AnalyzerContext {
...
+                    .enumerate()
+                    .map(|(i, f)| (&f.name, i))
+                    .collect();
+                let field_index_mapping: Vec<usize> = input_field_names
+                    .iter()
+                    .map(|field_name| {
+                        field_name_to_index
+                            .get(field_name)
+                            .copied()
+                            .unwrap_or(usize::MAX)
```
georgeh0: If not found in this case, it's a bug. So instead of `unwrap_or(usize::MAX)`, we should raise an error.

use something like:

```rs
ok_or_else(|| anyhow!("field `{}` not found in merged collector schema"))?;
```

src/builder/analyzer.rs
```diff
@@ -803,16 +876,47 @@ impl AnalyzerContext {
...
-                        op.collector_name.clone(),
-                        CollectorSchema::from_fields(fields_schema, op.auto_uuid_field.clone()),
-                        op_scope,
-                    )?,
+                    input_field_names,
+                    collector_schema,
+                    collector_ref,
+                    field_index_mapping,
                     fingerprinter,
                 });
```
georgeh0: These code needs to be moved into the `async move { ... }` block below.

This is because the collector's schema will also change later - once there're more `collect(...)` introduce new fields. The `async move` below is executed after analyzation is done - only until then we have the full collector schema.

src/execution/evaluator.rs
```diff
@@ -515,6 +515,43 @@ async fn evaluate_op_scope(
...
+                // Create field_values vector for all fields in the merged schema
+                let mut field_values: Vec<value::Value> =
+                    vec![value::Value::Null; op.collector_schema.fields.len()];
+
+                // Use pre-computed field index mappings for O(1) field placement
+                for (i, &collector_field_idx) in op.field_index_mapping.iter().enumerate() {
+                    if collector_field_idx != usize::MAX {
+                        field_values[collector_field_idx] = input_values[i].clone();
+                    }
+                }
```
georgeh0, line 531: I feel the code will be slightly simpler if we build the map in the other direction - use merged schema field index as key, and use an `Option<usize>` as value which is the currently collected schema field index.

src/builder/analyzer.rs
skalwaghe-56: Resolved in the latest push!

src/builder/analyzer.rs
skalwaghe-56: Resolved in the latest push!

src/builder/analyzer.rs
skalwaghe-56: Resolved in the latest push!

src/builder/analyzer.rs
skalwaghe-56: Resolved in the latest push!

src/builder/analyzer.rs
skalwaghe-56: Resolved in the latest push!

src/execution/evaluator.rs
skalwaghe-56: Resolved in the latest push!

georgeh0 (approved):

georgeh0:
Thanks a lot for the PR! This is a huge work!

badmonster0:
thanks @skalwaghe-56 this is amazing!!

skalwaghe-56:
@georgeh0 @badmonster0 Thanks!

badmonster0:
@skalwaghe-56
https://cocoindex.io/blogs/changelog-0310#skalwaghe-56

Thanks for teaching the collector to auto-merge and align multiple collect() calls with different schemas in #1153 — simplifying complex aggregation flows. We love it ❤️!!

Issue #599, opened 2025-06-09
[FEATURE] Support Redis queue for event notification (e.g. S3)


badmonster0:
Original request from Discord:

> I think the only immediate request I would have is maybe consider adding minio event support. I use minio locally and it supports events, just not through sqs.

Currently S3 supports event notifications by SQS. Users may want a different queue system, e.g. they use minio which can publish notification to Redis queue. We want to connect to Redis queue, for S3, and also enable us to deal with notifications for other sources that go through Redis in the future.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

aryasoni98:
I’m happy to work on this! Could you please assign the issue to me?

belloibrahv:
Hi! @badmonster0 @georgeh0 I'd like to work on this issue. 

I understand that the goal is to add Redis queue support for S3 event notifications (particularly for MinIO users) as an alternative to the existing SQS implementation. Looking at the codebase, I can see the current SQS integration in `src/ops/sources/amazon_s3.rs` with the `poll_sqs()` method and `change_stream()` implementation. 

I plan to:
- Add Redis queue polling similar to the existing SQS pattern
- Support the same S3 event notification format that MinIO publishes to Redis
- Maintain the existing `SourceChangeMessage` interface for consistency
- Add configuration options for Redis connection (host, port, queue/channel name)

I'm excited to contribute this feature and would appreciate it being assigned to me. Thanks!

PR #1189, opened 2025-10-15
feat: Add Redis queue support for S3 event notifications

belloibrahv:
- Add Redis client dependency to Cargo.toml
- Extend Amazon S3 source to support Redis pub/sub for event notifications
- Add redis_url and redis_channel configuration options
- Implement RedisContext with async pub/sub support
- Update change_stream to prefer Redis over SQS when both are configured
- Add comprehensive documentation for MinIO Redis setup
- Update Python spec to include Redis configuration fields

This enables MinIO users to receive S3-compatible event notifications without requiring AWS SQS, addressing issue #599.

### DIFF
--- a/python/cocoindex/sources/_engine_builtin_specs.py
+++ b/python/cocoindex/sources/_engine_builtin_specs.py
@@ -35,6 +35,16 @@ class GoogleDrive(op.SourceSpec):
     recent_changes_poll_interval: datetime.timedelta | None = None
 
 
+@dataclass
+class RedisNotification:
+    """Redis pub/sub configuration for event notifications."""
+
+    # Redis server URL (e.g., "redis://localhost:6379")
+    redis_url: str
+    # Redis channel name for pub/sub notifications
+    redis_channel: str
+
+
 class AmazonS3(op.SourceSpec):
     """Import data from an Amazon S3 bucket. Supports optional prefix and file filtering by glob patterns."""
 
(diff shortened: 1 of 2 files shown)


belloibrahv:
@georgeh0 @badmonster0 kindly help to review the PR when you have time. Thank You

src/ops/sources/amazon_s3.rs
```diff
@@ -289,6 +331,78 @@ impl Executor {
...
+                    });
+                }
+            }
+            Ok(None) => {
+                // No message received
+                return Ok(Vec::new());
+            }
+            Err(_) => {
+                // Timeout - no message received, return empty vec
+                return Ok(Vec::new());
```
georgeh0: I think we probably don't need to set a timeout. Long-time waiting is common and expected, and we're not throwing it as error any way - so just let it wait until there's a message. Also the current handling will hide real error.

python/cocoindex/sources/_engine_builtin_specs.py
```diff
@@ -46,6 +46,8 @@ class AmazonS3(op.SourceSpec):
     included_patterns: list[str] | None = None
     excluded_patterns: list[str] | None = None
     sqs_queue_url: str | None = None
+    redis_url: str | None = None
+    redis_channel: str | None = None
```
georgeh0: I would consider wrapping these two within another dataclass.

belloibrahv:
@georgeh0 kindly help review the PR, I have implemented the required changes based on your feedbacks

badmonster0:
thanks for the contribution @belloibrahv !

PR #1210, opened 2025-10-21
docs(cli): improve docstring formatting for `--help` and markdown

### LINKED ISSUE #1108: Make CLI docstrings / metadata render well in both `--help` and generated md file
With #1084 we start to generate markdown from docstrings / metadata from [`python/cocoindex/cli.py`](https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/python/cocoindex/cli.py).
Since when we create the CLI, we didn't have the auto generated md yet, something aren't rendered very well. 

For examples:

- Something should be backquoted, like `APP_FLOW_SPECIFIER` and `an_installed.module_name` in https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/docs/docs/core/cli-commands.md?plain=1#L38-L48
- There're redundant blank lines that break up sentences into separate paragraphs: https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/docs/docs/core/cli-commands.md?plain=1#L74-L80

Goal: make both `--help` and generated md look well.

To achieve this goal, we need to:
- Update [`python/cocoindex/cli.py`](https://github.com/cocoindex-io/cocoindex/blob/624b233b108c11b0e7d62a90faba2f2baf1f38ac/python/cocoindex/cli.py)
- The script to generate the markdown [`dev/generate_cli_docs.py`](https://github.com/cocoindex-io/cocoindex/blob/main/dev/generate_cli_docs.py) may also need to be updated, e.g. avoid unnecessary blank lines.

To test it, you may
- Run `cocoindex --help`, `cocoindex <subcommand> --help`
- View the generated markdown, by
  - Run `dev/generate_cli_docs.py`
  - Also view the generated markdown in docusaurus by following [`docs/README.md`](https://github.com/cocoindex-io/cocoindex/blob/main/docs/README.md)

belloibrahv:
- Add backticks around parameters, paths, and module names
- Fix paragraph formatting in `generate_cli_docs.py`
- Eliminate redundant blank lines in generated markdown
- Ensure consistent rendering in the terminal and documentation

Fixes #1108

### DIFF
--- a/dev/generate_cli_docs.py
+++ b/dev/generate_cli_docs.py
@@ -175,7 +175,6 @@ def extract_description(help_text: str) -> str:
     # Find the description between usage and options/commands
     description_lines = []
     in_description = False
-    last_was_empty = False
 
     for line in lines:
         if line.startswith("Usage:"):
@@ -187,12 +186,8 @@ def extract_description(help_text: str) -> str:
             stripped = line.strip()
             if stripped:
                 description_lines.append(stripped)
-                last_was_empty = False
-            else:
-                # Blank line - preserve as paragraph separator, avoid duplicates
-                if description_lines and not last_was_empty:
-                    description_lines.append("")
-                    last_was_empty = True
+            elif description_lines:  # Preserve blank line only if we have content
+                description_lines.append("")
 
     # Simply join with single newline - let Markdown handle paragraph formatting naturally
     description = "\n".join(description_lines) if description_lines else ""


belloibrahv:
@georgeh0 @badmonster0, kindly help me review when you have time. Thank You

docs/docs/core/cli-commands.md
```diff
@@ -154,21 +126,11 @@ cocoindex setup [OPTIONS] APP_TARGET
...
 
-APP_FLOW_SPECIFIER: Specifies the application and optionally the target
-
-flow. Can be one of the following formats:
-
-- path/to/your_app.py
+`APP_FLOW_SPECIFIER`: Specifies the application and optionally the target flow. Can be one of the following formats:
 
-- an_installed.module_name
+- `path/to/your_app.py` - `an_installed.module_name` - `path/to/your_app.py:SpecificFlowName` - `an_installed.module_name:SpecificFlowName`
```
georgeh0: The format here is not right.

dev/generate_cli_docs.py
```diff
@@ -175,17 +175,44 @@ def extract_description(help_text: str) -> str:
...
+        else:
+            current_paragraph.append(line)
 
-    description = "\n\n".join(description_lines) if description_lines else ""
+    # Add any remaining paragraph
+    if current_paragraph:
+        result.append(" ".join(current_paragraph))
+
+    # Join paragraphs with double newline
+    description = "\n\n".join(result) if result else ""
```
georgeh0: I think it'll just work if we simply join description lines with `\n` (i.e. 1 `\n` instead of 2), without any filtering logic above.

Because Markdown already allows single line break within a paragraph.

This means as long as the docstring for `cli.py` is written in the Markdown-compatible way, it'll just work.

The logic here can be simplified a lot, and the incorrect format for bullets (as I comment below) will also be fixed.

Does this make sense?

belloibrahv: @georgeh0 - Yes, this makes perfect sense! You're absolutely right.

I was over-engineering the solution by trying to intelligently join lines and create paragraphs. Your suggestion to simply join with `\n` and let Markdown handle it naturally is much cleaner and will fix both issues:

1. **Bullet list formatting** - Preserving the original line breaks will keep bullets on separate lines
2. **Simpler logic** - Trusting Markdown's natural behavior instead of complex paragraph detection

I'll simplify the `extract_description()` function as you suggested and ensure the docstrings in `cli.py` are Markdown-compatible. This will resolve the formatting issues.

Thank you for the clear feedback!

dev/generate_cli_docs.py
```diff
@@ -175,17 +175,27 @@ def extract_description(help_text: str) -> str:
...
+        elif in_description:
+            stripped = line.strip()
+            if stripped:
+                description_lines.append(stripped)
+                last_was_empty = False
+            else:
+                # Blank line - preserve as paragraph separator, avoid duplicates
+                if description_lines and not last_was_empty:
+                    description_lines.append("")
+                    last_was_empty = True
```
georgeh0: Do we still need these logic, or simply appending `stripped` will work?

belloibrahv: @georgeh0 - Great catch! You're absolutely right - the `last_was_empty` tracking was unnecessary. I've simplified it as you suggested. The code is now much cleaner and still produces the same correct output. Thank you for the feedback!

badmonster0:
@belloibrahv
https://cocoindex.io/blogs/changelog-0310#belloibrahv

Thanks for adding Redis queue support for S3 event notifications in #1189 and improving CLI --help/Markdown formatting in #1210 — smoother event-driven updates and CLI usage. We love it ❤️!!

Issue #1211, opened 2025-10-22
[FEATURE] make `--setup` the default behavior and deprecate the flag


georgeh0:
Currently several subcommands our CLI (`update`, `server`) accepts a `--setup`. Since `--setup` needs to be confirmed by users when there's any change needed, it's actually pretty safe if we always turn on `--setup`.

From now on, we want to always automatically execute setup-related logic for these subcommands, regardless of the `--setup` flag.

We still want to keep the `--setup` flag for a while for backward compatibility (users may have some script passing this flag in), but whenever users explicitly turned on `--setup`, output a warning to tell users it's deprecated.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

GooglyBlox:
Hi! Could I be assigned this issue?

georgeh0:
> Hi! Could I be assigned this issue?

@GooglyBlox assigned. thanks for taking hits!

GooglyBlox:
Shot a PR your way, feel free to review and send any feedback!

badmonster0:
https://github.com/cocoindex-io/cocoindex/pull/1212 thanks @GooglyBlox !

PR #1212, opened 2025-10-22
feat: deprecate setup flag

GooglyBlox:
Implements [#1211](https://github.com/cocoindex-io/cocoindex/issues/1211) 
Makes `--setup` the default behavior for `update` and `server` commands and deprecates the flag for future removal.

## Changes
- The `--setup` flag now defaults to `True` (previously `False`) for both `cocoindex update` and `cocoindex server` commands
- Added a callback that displays a warning when users explicitly use `--setup`, informing them the flag is deprecated
- Flag description now shows `(DEPRECATED)` and indicates it's enabled by default


GooglyBlox:
Very simple fix, can make any changes to wording or structure you need!

badmonster0:
@GooglyBlox
https://cocoindex.io/blogs/changelog-0310#googlyblox

Thanks for deprecating the setup flag in #1212 — simplifying the CLI surface and guiding users toward the preferred setup flow. We love it ❤️!!

PR #1218, opened 2025-10-24
feat(enum): add Enum type engine+python

MrAnayDongre:
Implements Enum basic type in engine + Python SDK, adds variants attr; updates JSON Schema and docs. Tests & pre-commit pass. Related: cocoindex-io#523

### DIFF
--- a/src/execution/live_updater.rs
+++ b/src/execution/live_updater.rs
@@ -10,6 +10,8 @@ use super::stats;
 use futures::future::try_join_all;
 use indicatif::ProgressBar;
 use sqlx::PgPool;
+use std::collections::VecDeque;
+use std::time::Instant;
 use tokio::{sync::watch, task::JoinSet, time::MissedTickBehavior};
 
 pub struct FlowLiveUpdaterUpdates {
@@ -95,8 +97,9 @@ struct SourceUpdateTask {
 
     status_tx: watch::Sender<FlowLiveUpdaterStatus>,
     num_remaining_tasks_tx: watch::Sender<usize>,
+    start_time: Instant,
+    rate_history: std::sync::Mutex<VecDeque<(Instant, i64)>>,
 }
-
 impl Drop for SourceUpdateTask {
     fn drop(&mut self) {
         self.status_tx.send_modify(|update| {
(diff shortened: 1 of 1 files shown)


src/base/schema.rs
```diff
@@ -23,6 +23,9 @@ pub enum BasicValueType {
     /// String encoded in UTF-8.
     Str,
 
+    /// Enumerated symbolic value.
+    Enum,
```
georgeh0: We should keep a bunch of allowed enum values here (as strings). The information is needed to create JSON schema, etc.

MrAnayDongre: Thanks! Variants are **field-specific**, so I stored them.
JSON Schema already reads this and emits `"enum"`.
I can add a typed accessor `EnrichedValueType::enum_variants()` and a `ENUM_VARIANTS_ATTR` const to avoid magic strings.
If you’d rather model it as `BasicValueType::Enum { variants: Option<Vec<String>> }`, happy to refactor.

georgeh0: I see. Thanks for the explanation.

I think putting it into `BasicValueType` is better than using attributes. Attributes are not part of the value type - more like some additional annotations. In most type systems, enum variants is part of the type (similar to variants of union type).

 think we should go with `BasicValueType::Enum(EnumTypeSchema)`, and `EnumTypeSchema` has a `variants: Vec<Arc<str>>` member.

python/cocoindex/typing.py
```diff
@@ -64,6 +66,19 @@ def __init__(self, key: str, value: Any):
...
+    String-like enumerated type. Use `variants` to hint allowed values.
+    Example:
+        color: Enum(variants=["red", "green", "blue"])
+    At runtime this is a plain `str`; `variants` are emitted as schema attrs.
+    """
+    if variants is not None:
+        return Annotated[str, TypeKind("Enum"), TypeAttr("variants", list(variants))]
+    return Annotated[str, TypeKind("Enum")]
+
+
```
georgeh0: We probably don't need this. Users can directly use Python's native way to represent enums, similar to union type.

MrAnayDongre: Addressed all feedback!

python/cocoindex/typing.py
```diff
@@ -13,6 +13,8 @@
     Literal,
     NamedTuple,
     Protocol,
+    Optional,
+    Sequence,
```
georgeh0, line 17: They're imported but not used.

src/base/json_schema.rs
```diff
@@ -298,9 +320,39 @@ impl JsonSchemaBuilder {
...
+            let mut vals: Vec<serde_json::Value> = enum_t
+                .variants
+                .iter()
+                .map(|s| serde_json::Value::String(s.to_string()))
+                .collect();
+
+            if vals.is_empty() {
+                if let Some(a) = enriched_value_type
+                    .attrs
+                    .get("variants")
```
georgeh0, line 337: Where does this come from?

I think if it's enum type, it should always come from `enum_t.variant`, right?

src/base/json_schema.rs
```diff
@@ -458,6 +510,57 @@ mod tests {
...
+              "enum": [
+                "red",
+                "green",
+                "blue"
+              ],
+              "type": "string"
+            }"#]]
+        .assert_eq(&serde_json::to_string_pretty(&json_schema).unwrap());
+    }
+
```
georgeh0, line 563: can we add another test for enum and `nullable: true` case?

sundareswarpullela:
Any update on this issue?

badmonster0:
hi @MrAnayDongre are you still active on this? if not we can close and maybe assign to some other contributors? lmk! thanks a lot!

badmonster0:
> Any update on this issue?

thanks @sundareswarpullela i messaged the contributor, if there's no update, we can re-open it to other contributors. is it blocking you?

sundareswarpullela:
Thank you for following up with the contributor @badmonster0 its not exactly blocking me right now, I have put in a work around, but definitely nice to have to be able to restrict the data in a field using an Enum.

badmonster0:
> Thank you for following up with the contributor @badmonster0 its not exactly blocking me right now, I have put in a work around, but definitely nice to have to be able to restrict the data in a field using an Enum.

sg! thanks a lot !!

badmonster0:
closing for now @MrAnayDongre - pls feel free to reopen if you are still on it. thanks a lot for your contribution!

badmonster0:
@sundareswarpullela i'll close this for now, we are releasing cocoindex v1 0 which uese python native type hope it simplify things a lot!

PR #1224, opened 2025-10-25
fix(runtime): guard sync API inside async contexts (#1085)

### LINKED ISSUE #1085: [FEATURE] Raise an error if the sync version of our Python SDK method is called within an async context
For many API methods, we expose a sync version and an async version, and the sync version usually call the async version by `execution_context.run(...)` ([call sites](https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex%20execution_context.run(&type=code)) - i.e. use the event loop managed by cocoindex library.

If users mistakenly call the sync version within an async context, the best case is other coroutines at the context will be blocked for a while, the worst case is deadlock - if they're already in the same async context managed by cocoindex.

We want to update this [`_ExecutionContext.run()`](https://github.com/cocoindex-io/cocoindex/blob/0dc1a4871c3de50538f5dba6ab9b5e15369dbf69/python/cocoindex/runtime.py#L33-L35) method to check if it's already within an event loop:

- If it's already on the same event loop, raise an error.
- If it's on a different event loop, log a warning.

---

samojavo:
Guard sync API when called from async contexts.

- If called on the CocoIndex-managed loop: raise RuntimeError (prevents deadlock).
- If called on a different running loop: emit a warning and run safely.
- If no loop is running: run as usual.

Closes #1085.

### DIFF
--- a/python/cocoindex/runtime.py
+++ b/python/cocoindex/runtime.py
@@ -42,7 +42,7 @@ class _ExecutionContext:
         except RuntimeError:
             running_loop = None
 
-        loop = self.event_loop  
+        loop = self.event_loop
 
         if running_loop is not None:
             if running_loop is loop:


python/cocoindex/runtime.py
```diff
@@ -24,15 +25,42 @@ def event_loop(self) -> asyncio.AbstractEventLoop:
...
+
+                def _runner(l: asyncio.AbstractEventLoop) -> None:
+                    asyncio.set_event_loop(l)
+                    l.run_forever()
+
+                threading.Thread(target=_runner, args=(loop,), daemon=True).start()
             return self._event_loop
 
-    def run(self, coro: Coroutine[Any, Any, T]) -> T:
+    def run(self, coro):
```
georgeh0: A check failed. Looks like it's because this type annotation is removed. You may try to bring it back and locally run `dmypy run python` to see if the type checking passes.

georgeh0:
Format check failed. You may follow the guide [here](https://cocoindex.io/docs/contributing/guide#submit-your-code) to do `pre-commit install` and run it locally. Thanks!

samojavo:
i did run ```ruff format``` so it should be fine now

badmonster0:
@samojavo
https://cocoindex.io/blogs/changelog-0310#samojavo

Thanks for guarding sync APIs inside async contexts in #1224 — improving runtime safety. We love it ❤️!!

PR #1241, opened 2025-11-02
[FEATURE]: Add timeout to prevent long-running functions (Closes #658)

### LINKED ISSUE #658: [FEATURE] Avoid a single function stuck forever
Sometimes a single function can be stuck for a really long time, e.g. Ollama ([bug](https://github.com/ollama/ollama/issues/8200)). Our users experienced it and described in #656.

- If a function takes really long time (e.g. more than 30s), log a warning.
- Each function may have a timeout (we can start from 300s as default, don't want it to be too short as some functions indeed needs long time sometimes, e.g. some PDF parsing; and each function can override it, similar to the setting to enable cache). We abort the function and return an error if timeout.

Side note: in the future we also want to expose more metrics/stats and allow users to inspect pending functions in cocoinsight

---

AdwitaSingh1711:
Summary:
Implements a 30-second warning and 300-second abort function mechanism as default for functions in `src/execution/evaluator.rs`. It also allows per-function overriding by specifying a timeout in decorator arguments, similar to the caching method.

Fixes #658 

Examples:

1. In `examples/manuals_llm_extraction/main.py` we can add `@cocoindex.op.executor_class(timeout=15)` to override current 300second timeout as shown
Screenshot 2025-11-02 at 8 09 34 AM

2. Additionally, if any other function takes longer than 30 seconds to execute, a warning message is displayed. If the execution time exceeds 300 seconds, the function times out.

Checklist:

 - [x] I have run all tests and pre-commit checks, which have passed.
 - [x] Cache, logs and others are not accidentally added to git's tracking history.
 - [x] My commits follow the conventional commits format.

### DIFF
--- a/python/cocoindex/op.py
+++ b/python/cocoindex/op.py
@@ -396,10 +396,7 @@ def _register_op_factory(
             return op_args.behavior_version
 
         def timeout(self) -> datetime.timedelta | None:
-            if op_args.timeout is not None:
-                return op_args.timeout
-
-            return None
+            return op_args.timeout
 
         def batching_options(self) -> dict[str, Any] | None:
             if op_args.batching:
--- a/rust/cocoindex/src/ops/py_factory.rs
+++ b/rust/cocoindex/src/ops/py_factory.rs
@@ -268,9 +268,15 @@ impl interface::SimpleFunctionFactory for PyFunctionFactory {
                             .extract::<Option<u32>>(py)?;
                         let timeout = executor
                             .call_method(py, "timeout", (), None)
-                            .to_result_with_py_trace(py)?
-                            .extract::<Option<u64>>(py)?
-                            .map(std::time::Duration::from_secs);
+                            .to_result_with_py_trace(py)?;
+                        let timeout = if timeout.is_none(py) {
+                            None
+                        } else {
+                            let td = timeout.into_bound(py);
+                            let total_seconds =
+                                td.call_method0("total_seconds")?.extract::<f64>()?;
+                            Some(std::time::Duration::from_secs_f64(total_seconds))
+                        };
                         let batching_options = executor
                             .call_method(py, "batching_options", (), None)
                             .to_result_with_py_trace(py)?


src/builder/analyzer.rs
```diff
@@ -727,16 +728,22 @@ impl AnalyzerContext {
...
                 async move {
                             trace!("Start building executor for transform op `{op_name}`");
                             let executor = executor.await.with_context(|| {
                                 format!("Preparing for transform op: {op_name}")
                             })?;
                             let enable_cache = executor.enable_cache();
                             let behavior_version = executor.behavior_version();
+                            let timeout = executor.timeout()
+                                .or(execution_options_timeout)
+                                .or(Some(Duration::from_secs(300)));
```
georgeh0: Let's put the default value into a global const. And as a second thought, we may want to start from a larger value - use 1800 second for now (later after things more stable, I'll gradually reduce it).

src/execution/evaluator.rs
```diff
@@ -380,6 +381,25 @@ async fn evaluate_op_scope(
...
 
+                let timeout_duration = op
+                    .function_exec_info
+                    .timeout
+                    .unwrap_or(Duration::from_secs(300));
+                let warn_duration = Duration::from_secs(30);
+
+                let op_name_for_warning = op.name.clone();
+                let op_kind_for_warning = op.op_kind.clone();
+                let warn_handle = tokio::spawn(async move {
```
georgeh0: I think we don't need `tokio::spawn`.

`tokio::select!` is sufficient ([example](https://github.com/cocoindex-io/cocoindex/blob/d204ed90e1f6f5ed63e17ceaa93a2c936631e4a5/src/execution/live_updater.rs#L395-L412))

src/execution/evaluator.rs
```diff
@@ -380,6 +381,25 @@ async fn evaluate_op_scope(
...
+                    tokio::time::sleep(warn_duration).await;
+                    eprintln!(
+                        "WARNING: Function '{}' ({}) is taking longer than 30s",
+                        op_kind_for_warning, op_name_for_warning
+                    );
+                    warn!(
+                        "Function '{}' ({}) is taking longer than 30s",
+                        op_kind_for_warning, op_name_for_warning
+                    );
```
georgeh0: We only need one. Let's keep `warn!` and get rid of `eprintln!`

src/builder/analyzer.rs
AdwitaSingh1711: Done. Added TIMEOUT_THRESHOLD and WARNING_THRESHOLD to both `src/builder/analyzer.rs` and `src/execution/evaluator.rs`

src/execution/evaluator.rs
```diff
@@ -515,43 +576,6 @@ async fn evaluate_op_scope(
...
-                                        .filter(|(i, _)| *i != uuid_idx)
-                                        .map(|(_, v)| v)
-                                        .collect::<Vec<_>>(),
-                                )?
-                                .into_fingerprint(),
-                        )?;
-                        field_values[uuid_idx] = value::Value::Basic(value::BasicValue::Uuid(uuid));
-                    }
-                }
-
```
georgeh0, line 552: These code should be kept (merge mistake?). Please bring back.

src/utils/retryable.rs
```diff
@@ -88,6 +88,7 @@ pub fn Ok<T>(value: T) -> Result<T> {
 
 pub struct RetryOptions {
     pub retry_timeout: Option<Duration>,
+    pub per_call_timeout: Option<Duration>,
```
georgeh0: `per_call_timeout` is not used. Why do we need this?

python/cocoindex/flow.py
```diff
@@ -486,6 +486,7 @@ class _SourceRefreshOptions:
 class _ExecutionOptions:
     max_inflight_rows: int | None = None
     max_inflight_bytes: int | None = None
+    timeout: int | None = None
```
georgeh0: Let's use `datetime.timedelta | None` here: prefer stronger type

python/cocoindex/op.py
```diff
@@ -164,6 +165,7 @@ class OpArgs:
     batching: bool = False
     max_batch_size: int | None = None
     behavior_version: int | None = None
+    timeout: int | None = None
```
georgeh0: same here, let's use `datetime.timedelta | None`

src/builder/analyzer.rs
```diff
@@ -13,6 +13,9 @@ use crate::{
 };
 use futures::future::{BoxFuture, try_join3};
 use futures::{FutureExt, future::try_join_all};
+use tokio::time::Duration;
```
georgeh0: This one is an alias. Let's directly use `std::time::Duration`.

src/builder/analyzer.rs
```diff
@@ -13,6 +13,9 @@ use crate::{
...
+
+const TIMEOUT_THRESHOLD: u64 = 1800;
```
georgeh0: Let's use `Duration` type here.

src/execution/evaluator.rs
AdwitaSingh1711: Thanks a bunch. I must have accidentally messed up during the merge. It has been restored now

src/utils/retryable.rs
AdwitaSingh1711: I have removed it (was initially trying to implement a per_call_timeout parameter for the LLM calls specifically to handle the Ollama bug but since the decorator arguments work, it isn't needed anymore)

python/cocoindex/op.py
```diff
@@ -391,6 +395,12 @@ def enable_cache(self) -> bool:
         def behavior_version(self) -> int | None:
             return op_args.behavior_version
 
+        def timeout(self) -> datetime.timedelta | None:
+            if op_args.timeout is not None:
+                return op_args.timeout
+
+            return None
```
georgeh0: You can directly write `return op_args.timeout` here.

georgeh0:
Thanks for the change!

georgeh0 (approved):

AdwitaSingh1711:
I had a lot of fun. Thanks for being so patient with me!

badmonster0:
@AdwitaSingh1711 
https://cocoindex.io/blogs/changelog-0310#adwitasingh1711

Thanks for adding function timeout support in #1241 — saving pipelines from long-running blocks. We love it ❤️!!

PR #1248, opened 2025-11-03
examples: fix mypy errors in example entry points; add local type-check script (#1091)

### LINKED ISSUE #1091: Check out for mypy type errors in the examples and clean up
https://github.com/cocoindex-io/cocoindex/tree/main/examples

samojavo:
• Make all example entry points pass mypy when run individually (with --ignore-missing-imports).
• Add a small helper script (dev/mypy_check_examples.ps1) and a brief note in dev/README.md.

Closes #1091

### DIFF
--- a/examples/face_recognition/main.py
+++ b/examples/face_recognition/main.py
@@ -7,7 +7,6 @@ import os
 import face_recognition
 import numpy as np
 from PIL import Image
-from typing import cast
 
 QDRANT_URL = os.getenv("QDRANT_URL", "http://localhost:6334/")
 QDRANT_COLLECTION = "face_embeddings"
@@ -86,7 +85,7 @@ def extract_face_embedding(
         np.array(img),
         known_face_locations=[(0, img.width - 1, img.height - 1, 0)],
     )[0]
-    return cast(cocoindex.Vector[cocoindex.Float32], embedding)
+    return embedding  # type: ignore
 
 
 @cocoindex.flow_def(name="FaceRecognition")
--- a/examples/postgres_source/main.py
+++ b/examples/postgres_source/main.py
@@ -1,6 +1,5 @@
-from typing import Any, cast
+from typing import Any
 import os
-import datetime
 
 from dotenv import load_dotenv
 from psycopg_pool import ConnectionPool
@@ -134,7 +133,7 @@ def search(pool: ConnectionPool, query: str, top_k: int = 5) -> list[dict[str, A
             """,
                 (query_vector, top_k),
             )
-            return cast(list[dict[str, Any]], cur.fetchall())
+            return cur.fetchall()  # type: ignore
 
 
 def _main() -> None:


examples/custom_output_files/main.py
```diff
@@ -96,7 +97,7 @@ def mutate(
 
 @cocoindex.op.function()
 def markdown_to_html(text: str) -> str:
-    return _markdown_it.render(text)
+    return cast(str, _markdown_it.render(text))
```
georgeh0: I'm thinking about also considering code simplicity/readability. Given these examples are for users to understand how to use cocoindex, this matters.

For case like this, it's essentially the underlying library doesn't have a specific type, and the return type of the current function is clear.
I think we can just ignore it by a comment like `# type: ignore`

(IMO `cast(...)` is more useful for values within a function - from the point on, the type is clear)

examples/fastapi_server_docker/main.py
```diff
@@ -103,16 +104,19 @@ def lifespan(app: FastAPI):
...
     limit: int = Query(5, description="Number of results"),
-):
+) -> dict[str, Any]:
     pool = request.app.state.pool
     results = search(pool, q, limit)
     return {"results": results}
 
 
+# Attach route without using decorator to avoid untyped-decorator when FastAPI types are unavailable
+fastapi_app.get("/search")(search_endpoint)
```
georgeh0: This is really unfortunate if have to use this way to workaround the issue. 

What does it mean by "when FastAPI types are unavailable"? Missed a package? If add the package to dependency, will it pass the check?

examples/image_search/main.py
```diff
@@ -19,7 +19,8 @@
 QDRANT_URL = os.getenv("QDRANT_URL", "http://localhost:6334/")
 QDRANT_COLLECTION = "ImageSearch"
 CLIP_MODEL_NAME = "openai/clip-vit-large-patch14"
-CLIP_MODEL_DIMENSION = 768
+CLIP_MODEL_DIMENSION: Final[int] = 768
+CLIPVector: TypeAlias = cocoindex.Vector[cocoindex.Float32, Literal[768]]
```
georgeh0: With this change, seems `CLIP_MODEL_DIMENSION` is no longer used. I think we can remove that?

examples/pdf_elements_embedding/main.py
```diff
@@ -7,16 +7,16 @@
...
 
 
 QDRANT_GRPC_URL = "http://localhost:6334"
 QDRANT_COLLECTION_IMAGE = "PdfElementsEmbeddingImage"
 QDRANT_COLLECTION_TEXT = "PdfElementsEmbeddingText"
 
 CLIP_MODEL_NAME = "openai/clip-vit-large-patch14"
-CLIP_MODEL_DIMENSION = 768
-ClipVectorType = cocoindex.Vector[cocoindex.Float32, Literal[CLIP_MODEL_DIMENSION]]
+CLIP_MODEL_DIMENSION: Final = 768
```
georgeh0: Same here, looks like `CLIP_MODEL_DIMENSION` is never used now?

samojavo:
is it fine this time?

examples/custom_output_files/main.py
georgeh0: Please resolve this comment and avoid `cast(...)` for function return types, and use `# type: ignore` comment instead. Thanks!

georgeh0:
> is it fine this time?

Hi @samojavo, please resolve the first comment above. Thanks!

examples/custom_output_files/main.py
samojavo: done

samojavo:
Should be done

examples/image_search/main.py
```diff
@@ -169,3 +169,6 @@ def search(
             for result in search_results
         ]
     }
+
+
+# Route attached via decorator above for readability
```
georgeh0: This comment is unnecessary.

examples/image_search/main.py
```diff
@@ -37,13 +37,13 @@ def embed_query(text: str) -> list[float]:
...
         features = model.get_text_features(**inputs)
-    return features[0].tolist()
+    return cast(list[float], features[0].tolist())
 
 
 @cocoindex.op.function(cache=True, behavior_version=1, gpu=True)
 def embed_image(
     img_bytes: bytes,
-) -> cocoindex.Vector[cocoindex.Float32, Literal[CLIP_MODEL_DIMENSION]]:
+) -> list[float]:
```
georgeh0: Actually this won't work as intended without putting the specific `CLIP_MODEL_DIMENSION` here (`@cocoindex.op.function` needs the dimension information so cocoindex can infer the vector column schema out of it). We need to bring it back, together with the definition of `CLIP_MODEL_DIMENSION`. Sorry for not noticing this earlier.

examples/pdf_embedding/main.py
```diff
@@ -32,7 +33,8 @@ def __call__(self, content: bytes) -> str:
         with tempfile.NamedTemporaryFile(delete=True, suffix=".pdf") as temp_file:
             temp_file.write(content)
             temp_file.flush()
-            text, _, _ = text_from_rendered(self._converter(temp_file.name))
+            text_any, _, _ = text_from_rendered(self._converter(temp_file.name))
+            text: str = cast(str, text_any)
             return text
```
georgeh0: Use `# type: ignore` for situation like these too.

In general, for all `return`, I think `# type: ignore` is sufficient. An explicit `cast` doesn't provide extra type safety.

examples/image_search/colpali_main.py
```diff
@@ -132,3 +131,7 @@ def search(
             for result in search_results.points
         ]
     }
+
+
+# Attach route without using decorator to avoid untyped-decorator when FastAPI types are unavailable
+app.get("/search")(search)
```
georgeh0: This looks weird. Please add it back as a decorator.

samojavo:
Pushed example fixes. Ready for re-review.

examples/image_search/main.py
```diff
@@ -29,21 +33,24 @@ def get_clip_model() -> tuple[CLIPModel, CLIPProcessor]:
...
     """
     Embed the caption using CLIP model.
     """
     model, processor = get_clip_model()
     inputs = processor(text=[text], return_tensors="pt", padding=True)
     with torch.no_grad():
         features = model.get_text_features(**inputs)
-    return features[0].tolist()
+    return cast(
+        cocoindex.Vector[cocoindex.Float32, ClipDim],
```
georgeh0: Please revert this cast.

examples/image_search/main.py
```diff
@@ -29,21 +33,24 @@ def get_clip_model() -> tuple[CLIPModel, CLIPProcessor]:
     return model, processor
 
 
-def embed_query(text: str) -> list[float]:
+def embed_query(text: str) -> cocoindex.Vector[cocoindex.Float32, ClipDim]:
```
georgeh0: Since this is not a cocoindex function, it doesn't need a dimension information. Using `list[float]` is good enough.

examples/image_search/main.py
```diff
@@ -52,7 +59,10 @@ def embed_image(
     inputs = processor(images=image, return_tensors="pt")
     with torch.no_grad():
         features = model.get_image_features(**inputs)
-    return features[0].tolist()
+    return cast(
+        cocoindex.Vector[cocoindex.Float32, ClipDim],
+        features[0].tolist(),
+    )
```
georgeh0: Same here, please directly return, and leave a `# type: ignore`

examples/image_search/main.py
```diff
@@ -19,7 +19,11 @@
 QDRANT_URL = os.getenv("QDRANT_URL", "http://localhost:6334/")
 QDRANT_COLLECTION = "ImageSearch"
 CLIP_MODEL_NAME = "openai/clip-vit-large-patch14"
-CLIP_MODEL_DIMENSION = 768
+CLIP_MODEL_DIMENSION: Final[int] = 768
+if TYPE_CHECKING:
+    ClipDim = Literal[768]
+else:
+    ClipDim = int
```
georgeh0: I think we always want to use `Literal[768]`.

Also since we have `CLIP_MODEL_DIMENSION`, I think we don't need another `ClipDim`. Using `Literal[CLIP_MODEL_DIMENSION]` directly like before will be sufficient.

samojavo:
We have to use Literal[768] in type hints because mypy’s literal check rejects Literal[CLIP_MODEL_DIMENSION] (“Parameter 1 of Literal[...] is invalid”). Sorry for extra mistakes along the way. Ready for re-review.

badmonster0:
nw! thanks for patiently working with us and keep code base high quality!

examples/face_recognition/main.py
```diff
@@ -85,7 +86,7 @@ def extract_face_embedding(
         np.array(img),
         known_face_locations=[(0, img.width - 1, img.height - 1, 0)],
     )[0]
-    return embedding
+    return cast(cocoindex.Vector[cocoindex.Float32], embedding)
```
georgeh0: Please use `# type: ignore` here too, instead of a `cast`.

examples/postgres_source/main.py
```diff
@@ -134,7 +134,7 @@ def search(pool: ConnectionPool, query: str, top_k: int = 5) -> list[dict[str, A
             """,
                 (query_vector, top_k),
             )
-            return cur.fetchall()
+            return cast(list[dict[str, Any]], cur.fetchall())
```
georgeh0: Please use `# type: ignore` here too, instead of a `cast`.

georgeh0:
> We have to use Literal[768] in type hints because mypy’s literal check rejects Literal[CLIP_MODEL_DIMENSION] (“Parameter 1 of Literal[...] is invalid”). Sorry for extra mistakes along the way. Ready for re-review.

I see. Thanks for trying it out!

Two more minor comments. Almost there. Thanks for your patience!

samojavo:
Did ```# type: ignore``` in both files - should be fine.

georgeh0 (approved):
Thanks a lot for the PR!

badmonster0:
@samojavo thank you so much for contributing to cocoindex! we made a section for you in latest release note https://cocoindex.io/blogs/changelog-0311-0326#samojavo
Love your work!!

PR #1267, opened 2025-11-07
feat: make `LlmGenerationClient::generate` return json

Lzzzzzt:
This PR is to complete #400 .

To finish that, this PR completed following tasks:

- [x] change the `LlmGenerateResponse` to enum to contain the text or json
- [x] make all the LLM client that implements the `LlmGenerationClient` return json or text based on the output format.

But actually, I'm not sure about the json value that I choose is correct, I just follow the original logic to return the json value.

### DIFF
--- a/rust/cocoindex/src/llm/anthropic.rs
+++ b/rust/cocoindex/src/llm/anthropic.rs
@@ -123,7 +123,7 @@ impl LlmGenerationClient for Client {
         }
         let text = if let Some(json) = extracted_json {
             // Try strict JSON serialization first
-            serde_json::to_string(&json)?
+            return Ok(LlmGenerateResponse::Json(json));
         } else {
             // Fallback: try text if no tool output found
             match &mut resp_json["content"][0]["text"] {
@@ -155,7 +155,7 @@ impl LlmGenerationClient for Client {
             }
         };
 
-        Ok(LlmGenerateResponse { text })
+        Ok(LlmGenerateResponse::Text(text))
     }
 
     fn json_schema_options(&self) -> ToJsonSchemaOptions {
--- a/rust/cocoindex/src/llm/gemini.rs
+++ b/rust/cocoindex/src/llm/gemini.rs
@@ -147,8 +147,11 @@ impl LlmGenerationClient for AiStudioClient {
             });
         }
 
+        let mut need_json = false;
+
         // If structured output is requested, add schema and responseMimeType
         if let Some(OutputFormat::JsonSchema { schema, .. }) = &request.output_format {
+            need_json = true;
             let mut schema_json = serde_json::to_value(schema)?;
             remove_additional_properties(&mut schema_json);
             payload["generationConfig"] = serde_json::json!({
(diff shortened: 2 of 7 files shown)


rust/cocoindex/src/llm/anthropic.rs
```diff
@@ -123,7 +123,7 @@ impl LlmGenerationClient for Client {
         }
         let text = if let Some(json) = extracted_json {
             // Try strict JSON serialization first
-            serde_json::to_string(&json)?
+            return Ok(LlmGenerateResponse::Json(json));
```
georgeh0, line 126: To make the two branches more symmetric, we may make both branches return a `LlmGenerateResponse`

rust/cocoindex/src/llm/anthropic.rs
georgeh0, line 138: In this branch, since we already parsed JSON, we should directly return JSON too.

rust/cocoindex/src/llm/gemini.rs
```diff
@@ -359,6 +371,18 @@ impl LlmGenerationClient for VertexAiClient {
 
         // Call the API
         let resp = req.send().await?;
+
+        if need_json {
+            match resp.candidates.into_iter().next() {
+                Some(resp_json) => {
+                    return Ok(super::LlmGenerateResponse::Json(serde_json::json!(
+                        resp_json
```
georgeh0, line 379: This is not the JSON of the response message, e.g. things like

```json
    {
      "content": {
        "parts": [
          {
            "text": "{...}"
          }
        ],
        "role": "model"
      },
      "index": 0
    }
```

NOT the JSON as the model response.

Did you get a chance to test this?

rust/cocoindex/src/llm/ollama.rs
```diff
@@ -108,10 +108,8 @@ impl LlmGenerationClient for Client {
         })
         .await
         .context("Ollama API error")?;
-        let json: OllamaResponse = res.json().await?;
-        Ok(super::LlmGenerateResponse {
-            text: json.response,
-        })
+
+        Ok(super::LlmGenerateResponse::Json(res.json().await?))
```
georgeh0, line 112: Again, this is the larger JSON as the response message, NOT the specific JSON responded by the model.

Did you get a chance to test?

rust/cocoindex/src/llm/openai.rs
```diff
@@ -145,15 +145,29 @@ impl LlmGenerationClient for Client {
...
+        let mut response_iter = response.choices.into_iter();
 
-        Ok(super::LlmGenerateResponse { text })
+        match request.output_format {
+            Some(OutputFormat::JsonSchema { .. }) => {
+                // Extract the response json from the first choice
+                let response_json = serde_json::json!(
+                    response_iter
+                        .next()
+                        .ok_or_else(|| anyhow::anyhow!("No response from OpenAI"))?
```
georgeh0, line 156: Same problem as above. This is NOT the specific JSON respond by the model.

georgeh0 (commented):
Please make sure you test it properly with our examples using LLM. Thanks!

rust/cocoindex/src/llm/gemini.rs
Lzzzzzt: So, to be more specific, if the return value is Json, just return the whole json value that responsed by llm, is that right?

georgeh0: In this case, I think we need to parse the `text` field above as JSON and return that, because that's the field with LLM response.

Please also make sure they're tested end to end, e.g. you may use [this example](https://github.com/cocoindex-io/cocoindex/blob/9e9ae74282673d5a54d05d813e6c3de16e388000/examples/manuals_llm_extraction/main.py#L115-L117) to test. Thanks!

badmonster0:
@Lzzzzzt is this still active?

badmonster0:
Close for now due to inactive. thanks a lot @Lzzzzzt for your contribution!

Issue #1289, opened 2025-11-13
[BUG] S3 style storage needs additional configuration


dcbark01:
Great work on Cocoindex so far. 
I've been trying to integrate it with my Minio S3 storage instance.
#600 introduced support for event notifications for S3-style storage (e.g. Minio). 
However, I kept getting 'dispatch errors' when trying to connect. 
The issue seems to be the path based addressing style that Minio requires.
In theory, you can set `addressing_style` in your AWS config profile,
but I could not seem to get that to propagate through to the config in the rust client.

Adding a `force_path_style` boolean parameter on the AWS S3 config spec appears to fix the problem.
I have a PR ready that I'll submit momentarily.

Before:

<img width="1075" height="58" alt="Image" src="https://github.com/user-attachments/assets/554bda9a-746c-4ee4-b38e-e92690e36aa6" />

<img width="809" height="314" alt="Image" src="https://github.com/user-attachments/assets/8566ef63-e798-4159-892a-ef011cdec1ac" />

After:

```python
data_scope["documents"] = flow_builder.add_source(
    cocoindex.sources.AmazonS3(
        bucket_name="mybucket",
        included_patterns=["*.pdf"],
        binary=True,
        redis=RedisNotification(
            redis_url="redis://localhost:6379", redis_channel="bucketevents",
        ),
        force_path_style=True,
    ),
    refresh_interval=timedelta(seconds=60),
)
```

<img width="1063" height="55" alt="Image" src="https://github.com/user-attachments/assets/e6633948-f021-4a32-8443-e11d9a4bcfa2" />

Now the flow ingests no problem:

<img width="1499" height="30" alt="Image" src="https://github.com/user-attachments/assets/0484d7fa-89b9-4be3-83d3-8a9f3cedb088" />

georgeh0:
Hi @dcbark01,

Thanks a lot for submitting the issue and the fix! We'll make a release later today with the fix!

PR #1290, opened 2025-11-13
fix: Added force_path_style param to S3 config

dcbark01:
See #1289


badmonster0:
@dcbark01 
https://cocoindex.io/blogs/changelog-0310#dcbark01

Thanks for adding the force_path_style parameter to S3 config in #1290 — improving support for S3-compatible storage. We love it ❤️!!

Issue #364, opened 2025-04-21
[FEATURE] Configurable additionalProperties in JSON Schema


par4m:
## Background
Some LLM APIs, such as Gemini, do not support the `additionalProperties` constraint in JSON schemas and return errors if it is present. Currently, we work around this by stripping all additional fields after schema generation using a helper in `gemini.rs`.

To address this more cleanly, introduce a `supports_additional_properties: bool` field in the `ToJsonSchemaOptions` struct.

## Tasks
- Add a `supports_additional_properties: bool` field to ToJsonSchemaOptions.
- Update schema generation logic to only include additionalProperties if this field is true.
- Update all code that constructs ToJsonSchemaOptions to set this field appropriately (e.g., false for Gemini, true for OpenAI).
- Remove the remove_additional_properties function from gemini.rs and any related post-processing.
Context

This addition will make it easier to support new LLMs with different schema requirements and reduce the risk of runtime errors due to schema incompatibility.

par4m:
I'll work on this after https://github.com/cocoindex-io/cocoindex/pull/360 is merged.

aryasoni98:
I’m happy to work on this! Could you please assign the issue to me?

georgeh0:
> I’m happy to work on this! Could you please assign the issue to me?

@aryasoni98 thanks for requesting. I already assigned https://github.com/cocoindex-io/cocoindex/issues/994 to you as requested. We encourage contributors taking one issue at a time especially at the beginning. We'll assign more issues after the existing one is done. Thanks!

Haleshot:
I'd like to give this a try! Looks like a nice issue to get started & contribute (atleast on a high level) to the codebase 😃

badmonster0:
assigned @Haleshot ! excited to have you here!!

badmonster0:
merged thanks @Haleshot !

PR #1312, opened 2025-11-22
`fix`: Make `additionalProperties` configurable in JSON schemas

Haleshot:
Addresses #364

Gemini doesn't support `additionalProperties` in JSON schemas, so we were recursively stripping it out after generation. Not great.

Added a `supports_additional_properties` bool to `ToJsonSchemaOptions` instead. Now we just don't include it in the first place for Gemini (both AI Studio and Vertex AI), while other providers (OpenAI, Ollama, Anthropic, Bedrock) still get it.

Removed the whole `remove_additional_properties` workaround function from `gemini.rs`. Cleaner this way.

### DIFF
--- a/rust/cocoindex/src/llm/openai.rs
+++ b/rust/cocoindex/src/llm/openai.rs
@@ -162,6 +162,7 @@ impl LlmGenerationClient for Client {
             supports_format: false,
             extract_descriptions: false,
             top_level_must_be_object: true,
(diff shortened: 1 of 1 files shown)


badmonster0:
@Haleshot new release is out and we have made a section for you! https://cocoindex.io/blogs/changelog-0310#haleshot thanks for your contribution and we love it ❤️ !!

Issue #1187, opened 2025-10-15
[FEATURE] Support Python 3.14 free-threaded


georgeh0:
We want to support Python 3.14 free-threaded build. 

See PyO3 guide [Supporting Free-Threaded CPython](https://pyo3.rs/main/free-threading.html#supporting-free-threaded-cpython)

We need to enable version-specific wheels build (`cp314t`), before free-threaded have stable ABI support ([PEP 803](https://peps.python.org/pep-0803/))

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

ansu86d:
> We want to support Python 3.14 free-threaded build. 
> 
> See PyO3 guide [Supporting Free-Threaded CPython](https://pyo3.rs/main/free-threading.html#supporting-free-threaded-cpython)
> 
> We need to enable version-specific wheels build (`cp314t`), before free-threaded have stable ABI support ([PEP 803](https://peps.python.org/pep-0803/))
> 
> 
> ---
> ❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
> Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.
> 

This is awesome , getting the new python's concurrent and parallelism feature would boost cocoindex performance as well !!

wuxiaobai24:
Can I work on this issue?

georgeh0:
Hi @wuxiaobai24, welcome to contribute! Assigned.

wuxiaobai24:
I tried building with Python 3.14t and discovered that **lancedb does not support cp314**. Before cocoindex can support Python 3.14, we need to get lancedb to support it first.

You can reproduce the issue using the following steps:

```bash
uv venv --python 3.14t .venv
source .venv/bin/activate
pip install maturin
maturin develop -E lancedb
```

You will get this output:

```bash
🍹 Building a mixed python/rust project
[...]

PR #1357, opened 2025-12-02
feat: Support Python 3.14 free-threaded

wuxiaobai24:
This PR adds Python 3.14 free-threaded mode support to CocoIndex, resolving #1187 

  Key Changes:

  - Configure PyO3 for free-threaded mode (gil_used = false)
  - Add free-threaded feature and dependencies
  - Add Python 3.14 free-threaded test workflow

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -54,7 +54,7 @@ jobs:
         #     cache-workspace-crates: 'false'
 
         - name: Rust tests (no default features)
-          run: cargo test --no-default-features --features abi3 --verbose
+          run: cargo test --no-default-features --verbose
 
         - name: Rust tests
           run: cargo test --verbose
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ pyo3 = { version = "0.27.1", features = [
     "auto-initialize",
     "chrono",
     "uuid",
-    "extension-module",
 ] }
 pythonize = "0.27.0"
 pyo3-async-runtimes = { version = "0.27.0", features = ["tokio-runtime"] }


rust/cocoindex/Cargo.toml
```diff
@@ -11,7 +11,8 @@ crate-type = ["cdylib"]
 
 [features]
 default = ["legacy-states-v0"]
-legacy-states-v0 = []
+legacy-states-v0 = [ "pyo3/abi3-py311"]
```
georgeh0: Let's enable this feature by default (just put it in `pyo3` below). It doesn't belong to `legacy-states-v0`

rust/cocoindex/Cargo.toml
```diff
@@ -11,7 +11,8 @@ crate-type = ["cdylib"]
...
+free-threaded = []
```
georgeh0: Seems this one is not used anywhere.

rust/cocoindex/Cargo.toml
wuxiaobai24: abi3-py311 now lives in a new abi3 feature which is part of default.

rust/cocoindex/Cargo.toml
wuxiaobai24: This feature now controls whether #[pymodule] is declared with gil_used = false.

rust/cocoindex/src/py/mod.rs
```diff
@@ -603,7 +603,8 @@ fn seder_roundtrip<'py>(
 }
 
 /// A Python module implemented in Rust.
-#[pymodule]
+#[cfg_attr(feature = "free-threaded", pymodule(gil_used = false))]
+#[cfg_attr(not(feature = "free-threaded"), pymodule)]
```
georgeh0: I think we don't need two branches here. We can always declare `gil_used = false`.

Based on [the documentation](https://pyo3.rs/v0.27.1/free-threading.html?highlight=gil_used#supporting-free-threaded-python-with-pyo3), as long as our code is thread safe, we can use `gil_used = false` here. When it's declared `false`, both with-gil and thread-safe python can use it.

wuxiaobai24: I've already made this change - removed the free-threaded feature flag and set `gil_used = false` by default.

georgeh0:
Thanks! Now an existing check failed. I think you may enable the `abi3` feature to the following line and have another try:

https://github.com/cocoindex-io/cocoindex/blob/e3f47fda57ce47a84773d289b19ae9f32421f35f/.github/workflows/_test.yml#L57

Thanks!

Cargo.toml
```diff
@@ -10,10 +10,10 @@ license = "Apache-2.0"
 
 [workspace.dependencies]
 pyo3 = { version = "0.27.1", features = [
-    "abi3-py311",
     "auto-initialize",
     "chrono",
     "uuid",
+    "extension-module",
```
georgeh0: Do we need this?  This is deprecated:

https://pyo3.rs/main/features.html#extension-module

.github/workflows/_test.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         #     cache-workspace-crates: 'false'
 
         - name: Rust tests (no default features)
-          run: cargo test --no-default-features --verbose
+          run: cargo test --no-default-features --features abi3 --verbose
```
georgeh0: Actually adding `--features abi3` doesn't solve the problem. We don't need this.

Noticed another potential problem (made a comment below). You may remove that and have another try.

You can also run this specific command `cargo test --no-default-features` locally to see if it works.

Thanks!

wuxiaobai24: It seems that the extension-module features were causing the test failures. After removing them, the tests pass locally.

georgeh0:
Thanks a lot for the PR!

Issue #1352, opened 2025-12-01
[FEATURE]  support vector and FTS indexes in lancedb connector


badmonster0:
**What is the use case?**
support vector and FTS indexes in lancedb connector

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

georgeh0:
Created #1358. During test, noticed LanceDB's [LTS support](https://lancedb.com/docs/indexing/fts-index/) is for Cloud and Enterprise edition only. Nevertheless, we can checkin this feature first and waiting for users response to see if there's any problems.

badmonster0:
thanks @georgeh0 !! cc @prrao87

prrao87:
Thank you! Will try it out in some workflows and raise issues if they arise.

prrao87:
Hi @georgeh0, I think the current implementation of FTS indexes (specifically, this [statement](https://github.com/cocoindex-io/cocoindex/blob/a42f711fa484c6e3258edbf96c3a6809d50ced5a/docs/docs/core/flow_def.mdx?plain=1#L364)) is incorrect. LanceDB OSS _does_ support FTS - but this is totally our fault (it's poorly documented, and I'm fixing that part [here](https://github.com/lancedb/docs/issues/19).

Basically, when working with a LanceDB `AsyncTable` and using an async API, which is what CocoIndex defaults to, we need to avoid using `create_fts_index` and instead use the more general method `create_index`, as defined in this [async test](https://github.com/lancedb/lancedb/blob/4b5bb2d76c5804e7bc39bfc8edfd6ec0ea9b54bb/python/python/tests/test_index.py#L148) in our test suite. The documentation issue arose on our end because an older API relied on tantivy (a full-text search engine in the Rust ecosystem), before we had an async Python API, but we then migrated to a native Lance format-supported FTS index to support async tables more fully.

This minimal example should show how to interface with the FTS index in LanceDB open source (no need for enterprise edition).

```py
import asyncio
import lancedb
import polars as pl
from lancedb.index import FTS

data = pl.DataFrame(
    {
        "id": [1, 2],
[...]

PR #1358, opened 2025-12-02
feat: support FTS for LanceDB

georgeh0:
#1352


badmonster0:
That was really fast :) thanks George!!

Issue #1286, opened 2025-11-12
[FEATURE] Use the `tracing` crate for logging in Rust engine


georgeh0:
Currently all our loggings at Rust engine are at the global scope without structure, so it's not always easy to know where a specific log entry comes from. We want to adopt the [`tracing`](https://github.com/tokio-rs/tracing) crate to improve it.

Specifically, we want to:

1. Create spans at various levels. We may have the following type of spans:
    * Flow builder(with flow name)
      * For each op (with op name)
    * Flow execution (with flow name)
      * For each source (with source name)
        * For each row
          * Evaluate: For each op (with op name)
          * Commit: For each target (with target kind)
    * Setup
      * For each flow (with specific flow name)
    * Each [service handler](https://github.com/cocoindex-io/cocoindex/blob/main/rust/cocoindex/src/service/flows.rs) has its own span
    * Each [API call](https://cocoindex.io/docs/core/flow_methods) has its own span
2. Replace various logging calls to the macros exposed by the `tracing` package ([more information](https://docs.rs/tracing/latest/tracing/#for-log-users)).
3. Make sure trace entries are also emitted to logs ([more information](https://docs.rs/tracing/latest/tracing/#emitting-log-records)).

Haleshot:
Would love to give this a try!

badmonster0:
@Haleshot it's yours :) thanks a lot for the contribution, excited for your work

PR #1363, opened 2025-12-03
Shift to `tracing` crate for logging (replacing `log`)

Haleshot:
This brings in [`tracing`](https://github.com/tokio-rs/tracing) to give the Rust logs _structure_ -- basically addressing the "where did this log come from?" problem mentioned in #1286. I've added spans to the [flow builder](https://github.com/cocoindex-io/cocoindex/blob/main/rust/cocoindex/src/builder/flow_builder.rs), [execution paths](https://github.com/cocoindex-io/cocoindex/tree/main/rust/cocoindex/src/execution), [service handlers](https://github.com/cocoindex-io/cocoindex/blob/main/rust/cocoindex/src/service/flows.rs), and setup operations. Replaced `env_logger` with [`tracing-subscriber`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/) for the subscriber, using `EnvFilter` so `RUST_LOG` still works as before. 

Had to keep `log` around since `sqlx` depends on it internally?


rust/cocoindex/src/execution/source_indexer.rs
```diff
@@ -335,6 +336,13 @@ impl SourceIndexingContext {
         _concur_permit: concur_control::CombinedConcurrencyControllerPermit,
         ack_fn: Option<AckFn>,
     ) {
+        // Note: Using trace! instead of span since this is spawned into a task
+        // and EnteredSpan is not Send-safe across await points
```
Haleshot: Went with a `trace!` event here rather than wrapping in `info_span!`. Since `EnteredSpan` isn't `Send`-safe and we have multiple `.await` points throughout `process_source_row`, a span would break the future's `Send` bound.

rust/cocoindex/src/lib_context.rs
```diff
@@ -283,7 +284,14 @@ impl LibContext {
 static LIB_INIT: OnceLock<()> = OnceLock::new();
 pub async fn create_lib_context(settings: settings::Settings) -> Result<LibContext> {
     LIB_INIT.get_or_init(|| {
-        let _ = env_logger::try_init();
+        // Initialize tracing subscriber with env filter for log level control
+        // Default to "info" level if RUST_LOG is not set
+        let env_filter = EnvFilter::try_from_default_env()
+            .unwrap_or_else(|_| EnvFilter::new("info"));
```
Haleshot: Defaulting to `"info"` here if `RUST_LOG` isn't set – seems reasonable since that's what `env_logger` was effectively doing before. Happy to change if there's a preferred default.

rust/cocoindex/src/execution/source_indexer.rs
georgeh0: How about using a `#[instrument(...)]` on the async method level instead?

rust/cocoindex/src/execution/live_updater.rs
```diff
@@ -122,6 +122,12 @@ impl SourceUpdateTask {
     }
 
     async fn run(self) -> Result<()> {
+        // Note: Using info! instead of span since EnteredSpan is not Send-safe across await points
+        info!(
+            flow_name = %self.flow.flow_instance.name,
+            source_name = %self.import_op().name,
+            "starting source update task"
+        );
```
georgeh0: How about using a `#[instrument(...)]` on the async method level instead?

georgeh0 (commented):
Really nice! Thanks a lot for the PR!

rust/cocoindex/src/execution/live_updater.rs
Haleshot: Nice catch: ~I mistakenly thought `EnteredSpan` not being Send applied to #[instrument] too, but it handles this internally.~ 
Updated!

georgeh0 (approved):
awesome! thanks for the PR!

badmonster0:
Great PR, thanks @Haleshot !

Issue #1377, opened 2025-12-09
[FEATURE] Refactor Error Handling: Replace `anyhow` with structured `thiserror` to support Host Exception Tunneling


georgeh0:
# Context & Motivation

Currently, our Core library relies on `anyhow` for error handling. While convenient, this causes a major issue when interfacing with the Python runtime (PyO3).

**The Problem:**
When a user-defined Python callback raises an exception (e.g., `ValueError`), it passes through our Rust Core. Because `anyhow` type-erases errors into a generic dynamic error object, the original Python exception is effectively lost. When the error surfaces back to the top-level Python code, the user receives a generic `RuntimeError` containing a stringified message, losing the original exception type and traceback.

**The Goal:**
We need to refactor the error handling to use `thiserror` with a type-erased "Host Error" variant. This will allow us to:

1.  **"Tunnel" Host Exceptions:** Preserve the exact `PyErr` (or future `JsValue`) object through the Rust Core as a trait object and re-raise it at the boundary.
2.  **Avoid Generic Pollution:** Use dynamic dispatch (`Box<dyn HostError>`) so core functions do not require generic type parameters.
3.  **Structured Errors:** Distinguish between "Client/Validation" errors (which should look like API errors) and "Internal" errors.
4.  **Support Backtraces & Context:** Automatically capture stack traces and support `anyhow`-style context wrapping.

-----

# The Proposed Design

We will replace `anyhow::Result<T>` with `cocoindex_utils::Result<T>`.

## 1\. The Host Error Trait

In `cocoindex_utils`, we define a trait that any host language exception (Python `PyErr`, JS `Error`) must implement to pass through our core.

```rust
use std::any::Any;
use std::fmt::{Debug, Display};

// This trait allows us to store the error, print it, and downcast it later.
pub trait HostError: Debug + Display + Send + Sync + 'static {
    // Required for downcasting back to the concrete type (e.g. PyErr)
    fn as_any(&self) -> &dyn Any;
}
```

## 2\. The Core Error Enum

The `Error` enum uses the `HostError` trait object for the tunnel.

```rust
use thiserror::Error;
use std::backtrace::Backtrace;

#[derive(Error, Debug)]
pub enum Error {
    // 1. Context Wrapper (replaces anyhow::Context)
    // Allows us to wrap errors with strings but still "drill down" to the cause later.
    #[error("{msg}")]
    Context {
        msg: String,
        #[source]
        source: Box<Error>,
    },

[...]

PR #1380, opened 2025-12-13
`refactor`: introduce structured error system w/ host exception tunneling

Haleshot:
Addresses #1377 by introducing a structured error system that supports host exception tunneling. The core `CError` enum has four variants: `Context` for wrapping errors with messages (replacing [`anyhow::Context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html)), `HostLang` for tunneling host exceptions via `Box<dyn HostError>`, `Client` for validation/4xx errors, and `Internal` for 5xx errors; both of which I believe, capture the backtraces via/ [`std::backtrace::Backtrace`](https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html).

On the Python side, `PyErrWrapper` implements `HostError` and `cerror_to_pyerr()` handles the reverse conversion, traversing the context chain and downcasting via [`Any`](https://doc.rust-lang.org/std/any/trait.Any.html) to extract the original `PyErr` if present.

Following your [suggestion](https://github.com/cocoindex-io/cocoindex/issues/1377#issuecomment-3649578946) about incremental/iteratively working within the same PR, I'm pushing this first for your review. If this looks good, the next step is migrating the a lot of `anyhow::Result` usages across the codebase to use `CResult`. The existing `anyhow`-based types should/will be preserved for backwards compatibility during that migration.

_Edit 1_: Tended the changes requested [here](https://github.com/cocoindex-io/cocoindex/pull/1380#discussion_r2616504393) and [here](https://github.com/cocoindex-io/cocoindex/pull/1380#discussion_r2616505175); can proceed with the migration of `anyhow::Result` usages if it looks good!

_Edit 2_: Completed the migration of error macros from `anyhow::bail!`/`anyhow::anyhow!` to the new structured `client_bail!`/`client_error!` and `internal_bail!`/`internal_error!` macros across ~35 files. 

The new macros create `CError` with `Client`/`Internal` variants that capture backtraces, then convert to `anyhow::Error` for compatibility with existing return types. This preserves all the structured error semantics (including host exception tunneling for Python) while avoiding a massive signature refactor. The `anyhow::Result` type remains as-is...

**_PS_:** I took, what I believe is a pragmatic approach here - rather than replacing every `anyhow::Result<T>` with `CResult<T>` (which would've required touching hundreds of serde impls and function signatures), I wrapped `CError` inside `anyhow::Error`. Is this acceptable, or would you prefer a full `CResult<T>` migration in a _follow-up_?

_Full disclosure: Claude (via Cursor) helped me grep through the (relevant files in the) codebase to identify all the error macro usages. Turns out clicking through search results in an IDE beats wrestling w/ git grepping 🙃_

_Edit 3_: Per [this comment](https://github.com/cocoindex-io/cocoindex/pull/1380/#issuecomment-3666393199), I did what was stated, but also ended up fixing/addding a bunch of other stuff along the way:

**What I also ended up doing** _(because the build kept yelling at me...)_:
- Fixed Python interop in `py_factory.rs` - all `PyErr` now tunneled via `Error::host()`, `pythonize`/`depythonize` errors converted with `.internal()`, `from_py_future` results properly handled
- Migrated `schema.rs`, `analyzer.rs`, `live_updater.rs`, `fingerprint.rs` to use (existing ) structured `Error` type - swapped `TryInto<Error = anyhow::Error>` bounds, removed `anyhow::bail!` usages, etc.
- Fixed `SharedError` to wrap our structured `Error` instead of `anyhow::Error`
- Fixed some `retryable::run` closures in `neo4j.rs` that were trying to use a non-existent error variant... 🤔 

<details>
<summary>What's intentionally left as-is (for future) follow-up PRs)</summary>

Some files still use `anyhow` and that's... intentional? Why I feel:

**`retryable.rs`** - The `retryable::Error` struct seems to wrap around `anyhow::Error` by design. The loop gathers (appends) context on each attempt and the trait bounds alos expect it. Reworking this would mean redesigning how the whole retry mechanism handles errors.

**`http.rs`** - Built around `anyhow` again. Could migrate but it touches retry logic too? // TODO: Follow-up?

**`py_utils/src/error.rs` (`ToResultWithPyTrace`)** -  Call sites immediately convert via `.map_err(Error::from)` so it works, just... not pretty. Could eventually replace with something that returns our `Result<T>` and wraps formatted traces in `Error::host()`. Gradual phase-out?

**`utils/src/error.rs` (the bridge stuff)** - these are conversion bridges that let the migration happen incrementally. `From<anyhow::Error>` is actually _required_ right now since retryable/http still produce anyhow errors. Can remove these once those deps migrate?

</details>

<details>
<summary> TLDR  (for edit 3 mainly)</summary>

- w/ host tunneling (`Error::host()` for `PyErr`)
- Migrated relevant files from `anyhow::bail!` to structured macros
- `Error`/`Result` as primary types (renamed from `CError`/`CResult`)
- `retryable.rs` and `http.rs` left as-is - designed around anyhow, needs separate refactor
- Conversion bridges kept for gradual migration

</details>

### DIFF
--- a/rust/utils/src/error.rs
+++ b/rust/utils/src/error.rs
@@ -28,8 +28,9 @@ pub enum Error {
     },
     Internal {
         source: Box<dyn StdError + Send + Sync>,
-        bt: Option<Backtrace>,
+        bt: Backtrace,
     },
+    InternalAnyhow(anyhow::Error),
 }
 
 impl Display for Error {
@@ -39,6 +40,7 @@ impl Display for Error {
             Error::HostLang(e) => write!(f, "{}", e),
             Error::Client { msg, .. } => write!(f, "Invalid Request: {}", msg),
             Error::Internal { source, .. } => write!(f, "{}", source),
+            Error::InternalAnyhow(e) => write!(f, "{}", e),
         }
     }
 }
@@ -49,6 +51,7 @@ impl StdError for Error {
             Error::Context { source, .. } => Some(source.as_ref()),
             Error::HostLang(e) => Some(e.as_ref()),
             Error::Internal { source, .. } => Some(source.as_ref()),
+            Error::InternalAnyhow(e) => e.source(),
             Error::Client { .. } => None,
         }
     }
@@ -75,21 +78,22 @@ impl Error {
     pub fn internal(e: impl StdError + Send + Sync + 'static) -> Self {
         Self::Internal {
             source: Box::new(e),
-            bt: Some(Backtrace::capture()),
+            bt: Backtrace::capture(),
         }
     }
 
     pub fn internal_msg(msg: impl Into<String>) -> Self {
         Self::Internal {
             source: Box::new(StringError(msg.into())),
-            bt: Some(Backtrace::capture()),
+            bt: Backtrace::capture(),
         }
     }
 
     pub fn backtrace(&self) -> Option<&Backtrace> {
         match self {
             Error::Client { bt, .. } => Some(bt),
-            Error::Internal { bt, .. } => bt.as_ref(),
+            Error::Internal { bt, .. } => Some(bt),
+            Error::InternalAnyhow(e) => Some(e.backtrace()),
             Error::Context { source, .. } => source.backtrace(),
             Error::HostLang(_) => None,
         }
@@ -121,28 +125,9 @@ impl From<serde_json::Error> for Error {
     }
 }
 
-// Wrapper for anyhow::Error to implement StdError
-#[derive(Debug)]
-pub struct AnyhowWrapper(anyhow::Error);
-
-impl Display for AnyhowWrapper {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        Display::fmt(&self.0, f)
-    }
-}
-
-impl StdError for AnyhowWrapper {
-    fn source(&self) -> Option<&(dyn StdError + 'static)> {
-        self.0.source()
-    }
-}
-
 impl From<anyhow::Error> for Error {
     fn from(e: anyhow::Error) -> Self {
-        Error::Internal {
-            source: Box::new(AnyhowWrapper(e)),
-            bt: None,
-        }
+        Error::InternalAnyhow(e)
     }
 }
 
@@ -154,20 +139,14 @@ impl From<ApiError> for Error {
                 bt: Backtrace::capture(),
             }
         } else {
-            Error::Internal {
-                source: Box::new(AnyhowWrapper(e.err)),
-                bt: None,
-            }
+            Error::InternalAnyhow(e.err)
         }
     }
 }
 
 impl From<crate::retryable::Error> for Error {
     fn from(e: crate::retryable::Error) -> Self {
-        Error::Internal {
-            source: Box::new(AnyhowWrapper(e.error)),
-            bt: None,
-        }
+        Error::InternalAnyhow(e.error)
     }
 }
 
@@ -210,7 +189,7 @@ impl<T, E: StdError + Send + Sync + 'static> IntoInternal<T> for std::result::Re
     fn internal(self) -> Result<T> {
         self.map_err(|e| Error::Internal {
             source: Box::new(e),
-            bt: Some(Backtrace::capture()),
+            bt: Backtrace::capture(),
         })
     }
 }
@@ -248,7 +227,7 @@ impl<T, E: StdError + Send + Sync + 'static> ResultExt<T, E> for std::result::Re
             msg: context.into(),
             source: Box::new(Error::Internal {
                 source: Box::new(e),
-                bt: Some(Backtrace::capture()),
+                bt: Backtrace::capture(),
             }),
         })
     }
@@ -258,7 +237,7 @@ impl<T, E: StdError + Send + Sync + 'static> ResultExt<T, E> for std::result::Re
             msg: f().into(),
             source: Box::new(Error::Internal {
                 source: Box::new(e),
-                bt: Some(Backtrace::capture()),
+                bt: Backtrace::capture(),
             }),
         })
     }
@@ -281,7 +260,7 @@ impl IntoResponse for Error {
         let (status_code, error_msg) = match &self {
             Error::Client { msg, .. } => (StatusCode::BAD_REQUEST, msg.clone()),
             Error::HostLang(e) => (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()),
-            Error::Context { .. } | Error::Internal { .. } => {
+            Error::Context { .. } | Error::Internal { .. } | Error::InternalAnyhow(_) => {
                 (StatusCode::INTERNAL_SERVER_ERROR, format!("{:?}", self))
             }
         };
@@ -374,7 +353,7 @@ impl SharedError {
                 // Already extracted; return a generic internal error with the residual message.
                 return Error::Internal {
                     source: Box::new(err.clone()),
-                    bt: Some(Backtrace::capture()),
+                    bt: Backtrace::capture(),
                 };
             }
             SharedErrorState::Error(err) => ResidualError::new(err),
@@ -422,7 +401,7 @@ impl SharedError {
         Self(Arc::new(Mutex::new(SharedErrorState::Error(
             Error::Internal {
                 source: Box::new(err),
-                bt: Some(Backtrace::capture()),
+                bt: Backtrace::capture(),
             },
         ))))
     }


rust/utils/src/error.rs
```diff
@@ -6,12 +6,234 @@ use axum::{
...
+    backtrace::Backtrace,
+    error::Error as StdError,
     fmt::{Debug, Display},
     sync::{Arc, Mutex},
 };
 
-pub struct ResidualErrorData {
+pub trait HostError: Debug + Display + Send + Sync + 'static {
+    fn as_any(&self) -> &dyn Any;
+}
```
georgeh0: I just realized we actually don't need `as_any()` method here. Instead, we can make `Any` a base trait:

```rust
// 1. The Supertraits: Any (for downcasting) + Send/Sync (for thread safety)
pub trait HostError: Any + Debug + Display + Send + Sync + 'static {}

// 2. Blanket Implementation
// Any type that meets the requirements automatically implements HostError.
// This means PyErr implements this AUTOMATICALLY. No wrapper needed.
impl<T> HostError for T 
where 
    T: Any + Debug + Display + Send + Sync + 'static 
{}
```

 (it was needed before Rust v1.86.0, which introduced [trait upcasting](https://blog.rust-lang.org/2025/04/03/Rust-1.86.0/#trait-upcasting))

rust/utils/src/error.rs
```diff
@@ -6,12 +6,234 @@ use axum::{
...
+
+#[derive(Debug)]
+pub enum CError {
```
georgeh0: I think we can use `thiserror` when defining this error type. It saves boilerplates in error messages, and wire source/backtraces/etc. in standard way.

georgeh0:
@Haleshot Really fast! Looks very reasonable! Thanks!

rust/utils/src/error.rs
Haleshot: done - blanket impl w/ `Any + 'static` as suggested. Thanks for the catch (did not nkow about the later version which intro'ed this).

rust/utils/src/error.rs
Haleshot: Tried this but thiserror version (1.x) has issues: `#[error(transparent)]` needs `HostError: StdError` (breaks blanket impl), and the `Backtrace` fields require unstable `error_generic_member_access`. Using it for `StringError` only?

rust/utils/src/error.rs
```diff
@@ -1,17 +1,237 @@
...
+        }
+    }
+
+    pub fn is_host_error(&self) -> bool {
+        self.find_host_error().is_some()
+    }
+
+    pub fn is_client_error(&self) -> bool {
+        matches!(self, CError::Client { .. })
+    }
```
georgeh0: I think this should behave similarly as `is_host_error`, i.e. peeling off the `Context` layers. So maybe consider having a more generic helper method like `without_contexts()` and share by the few methods above.

rust/utils/src/error.rs
```diff
@@ -1,17 +1,237 @@
...
+    any::Any,
+    backtrace::Backtrace,
+    error::Error as StdError,
     fmt::{Debug, Display},
     sync::{Arc, Mutex},
 };
+use thiserror::Error;
 
-pub struct ResidualErrorData {
+pub trait HostError: Any + Debug + Display + Send + Sync + 'static {}
```
georgeh0: actually we can replace `Debug + Display` with `StdError` here. Then for `HostLang` branch, we can return it for `source()`.

rust/utils/src/error.rs
georgeh0: Thanks for trying out.

FYI, after using `HostError: StdError`, the `#[error(transparent)]` will work. And the unstable `error_generic_member_access` means the `thiserror` won't provide a `backtrace()` method for `Error`, but will still capture the `Backtrace` field when there's a `#[from]`. But since we don't use `#[from]` now, the value-add of `thiserror` is very limited for our case. (it's probably more useful for libraries using fine-grain error types, with specific-typed errors, and `#[from]` will be used a lot)

Now I think we're perfectly fine with manually implement the error. We can remove dependency on `thiserror`.

rust/utils/src/error.rs
```diff
@@ -1,17 +1,237 @@
...
+
+impl<T, E: StdError + Send + Sync + 'static> IntoInternal<T> for std::result::Result<T, E> {
+    fn internal(self) -> CResult<T> {
+        self.map_err(|e| CError::Internal {
+            source: Box::new(e),
+            bt: Backtrace::capture(),
+        })
+    }
+}
+
```
georgeh0, line 196: (Not needed immediately, but may need later during your migration, if any of our dependency uses `anyhow::Error`)

`anyhow::Error` doesn't implement `StdError`, so it cannot fit into our `Internal`.

The solution will be:
- Define a wrapper type for `anyhow::Error` that implements `StdError`.
- Implement `IntoInternal<T>` for `anyhow::Error`, constructing the wrapper type above.

rust/py_utils/src/error.rs
```diff
@@ -11,6 +15,54 @@ impl PythonExecutionContext {
...
+            return Python::attach(|py| wrapper.0.clone_ref(py));
+        }
+    }
+
+    match &err {
+        CError::Client { msg, .. } => PyValueError::new_err(msg.clone()),
+        CError::HostLang(e) => PyRuntimeError::new_err(e.to_string()),
+        CError::Context { .. } | CError::Internal { .. } => {
+            PyRuntimeError::new_err(format!("{:?}", err))
+        }
```
georgeh0: here we want to format a context chain, and also attach the back trace if present for internal error, e.g. 

```
{error_message}
Caused by: {context1}
Caused by: {context2}
......

{backtrace}
```

georgeh0 (commented):
The API looks good! Left a few comments for implementations. Feel free to start migrating existing code to the new error type.

Haleshot:
Updated the PR description above with "Edits" to reflect the changes made incrementally; awaiting your review :)

rust/py_utils/src/error.rs
```diff
@@ -11,6 +15,66 @@ impl PythonExecutionContext {
...
+    }
+}
+
+impl Display for PyErrWrapper {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl std::error::Error for PyErrWrapper {}
```
georgeh0: seems this `PyErrWrapper` doesn't provide additional values. We may not need this layer.

rust/py_utils/src/error.rs
```diff
@@ -11,6 +15,66 @@ impl PythonExecutionContext {
...
+pub fn cerror_to_pyerr(err: CError) -> PyErr {
+    if let Some(host_err) = err.find_host_error() {
+        let any: &dyn Any = host_err;
+        if let Some(wrapper) = any.downcast_ref::<PyErrWrapper>() {
+            return Python::attach(|py| wrapper.0.clone_ref(py));
+        }
+    }
+
+    match err.without_contexts() {
+        CError::Client { msg, .. } => PyValueError::new_err(msg.clone()),
```
georgeh0: I think for client error, context is also valuable. Probably the only difference is backtrace is not necessary.

rust/utils/src/error.rs
```diff
@@ -1,17 +1,339 @@
...
+        match self {
+            CError::HostLang(host_err) => Some(host_err.as_ref()),
+            CError::Context { source, .. } => source.find_host_error(),
+            _ => None,
+        }
+    }
+
+    pub fn is_host_error(&self) -> bool {
+        self.find_host_error().is_some()
+    }
```
georgeh0: Actually both `is_host_error()` and `is_client_error()` are redundant (they're not used by any code any way, other than code to test themselves). Using `without_contexts()` can easily achieve the same goal. We don't need to create a tester for each error type.

Similar to `find_host_error()`. Callers can use `without_contexts()` + matcher directly too (and they need a mater any way for the `Option`).

georgeh0:
About

> **_PS_:** I took, what I believe is a pragmatic approach here - rather than replacing every `anyhow::Result<T>` with `CResult<T>` (which would've required touching hundreds of serde impls and function signatures), I wrapped `CError` inside `anyhow::Error`. Is this acceptable, or would you prefer a full `CResult<T>` migration in a _follow-up_?

There're two parts:
1. Error creations. e.g. serde related stuffs. Calling on the existing `bail!` and `api_bail!` macros also belong to this.
2. Error propagation. e.g. popping up by `?`, with additional contexts, etc.

For 1, I think it's it's OK to handle different parts separately. The final goal is not to use `anyhow`, but it can happen in multiple steps.

For 2, the problem of using `anyhow::Error` as wrapper is: the structured information will be lost during propagation (it allows `downcast` to restore the original structured information, but after calling `context()`, `downcast` can no longer restore to the original structured error).

To make the migration easier, I have another idea. A lot of existing crates just use `Error` as name of their error type, and they also define a specialized `Result` type. For example, for `sqlx` ([`Error`](https://docs.rs/sqlx_wasi/latest/sqlx/enum.Error.html), [`Result`](https://docs.rs/sqlx_wasi/latest/sqlx/error/type.Result.html)). We can do the similar thing:

- Rename `CError` to`Error`, and add a specialized `Result` type.
- Update [this line](https://github.com/cocoindex-io/cocoindex/blob/f8d63cc3dddd230ca6cc97302c940cd15081350f/rust/cocoindex/src/prelude.rs#L3) in prelude to use our  new types instead.
- One minor thing to keep in mind is that we use `anyhow::Ok(...)` sometimes (which is a syntax sugar, equivalent to `Ok::<_, anyhow::Error>(...)`. They can be avoided by explicitly annotate return type values ([example](https://github.com/cocoindex-io/cocoindex/blob/f8d63cc3dddd230ca6cc97302c940cd15081350f/rust/cocoindex/src/setup/driver.rs#L69)).

In this way, most modules won't need to be touched, if they're just propagating errors rather than creating errors.

What do you think?

rust/py_utils/src/error.rs
Haleshot: I see it asa thin wrapper, but PyErr doesn't impl `std::error::Error` which `HostLang(Box<dyn HostError>)` needs (some bridging is needed in this case). Were you thinking of storing PyErr directly in a dedicated variant (`CError::PyErr(PyErr)` variant instead of using trait object?), or relaxing the trait bounds? Happy to go either direction.

Haleshot:
On the approach suggested above: I'll go about doing the following:

- Rename `CError` → `Error`, `CResult` → `Result` in `cocoindex_utils`
- Update `prelude.rs` to use these instead of `anyhow`

Also, hope it's fine for putting this (& you) through multiple review rounds 😅 - definitely not intentional (hope you don't mind)! You're also more than welcome to push directly to this branch if you feel I'm not getting it (still making my way about Rust and the codebase.

rust/py_utils/src/error.rs
georgeh0: We still want to keep the generic `HostLang(Box<dyn HostError>)` for future proof.

Looks like `PyErr` actually implemented `std::error::Error`?
https://docs.rs/pyo3/latest/pyo3/struct.PyErr.html#impl-Error-for-PyErr

georgeh0:
> On the approach suggested above: I'll go about doing the following:
> 
> * Rename `CError` → `Error`, `CResult` → `Result` in `cocoindex_utils`
> * Update `prelude.rs` to use these instead of `anyhow`
> 
> Also, hope it's fine for putting this (& you) through multiple review rounds 😅 - definitely not intentional (hope you don't mind)! You're also more than welcome to push directly to this branch if you feel I'm not getting it (still making my way about Rust and the codebase.

Looks good! No worries, this is a large refactor and it's well expected to have multiple rounds of review :)  Thanks a lot for pushing it through!

rust/py_utils/src/error.rs
Haleshot: Ahh, yes 🙈- `PyErr` does impl `Error`, thanks for the relevant docs webpage. Will remove `PyErrWrapper` and use `PyErr` directly with `CError::host(err)`.

rust/utils/src/error.rs
```diff
@@ -1,17 +1,327 @@
...
+            Error::Client { .. } => None,
+        }
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+// Backwards compatibility aliases
+pub type CError = Error;
+pub type CResult<T> = Result<T>;
```
Haleshot, line 64: These aliases preserve backwards compat during migration - can be dropped once all are switched to `Error`/`Result` directly. Makes sense?

rust/cocoindex/src/prelude.rs
```diff
@@ -27,13 +26,17 @@ pub(crate) use crate::setup;
 pub(crate) use crate::setup::AuthRegistry;
 pub(crate) use cocoindex_utils as utils;
 pub(crate) use cocoindex_utils::error::{ApiError, invariance_violation};
+pub(crate) use cocoindex_utils::error::{ContextExt, Error, IntoInternal, Result, ResultExt};
```
Haleshot, line 29: Now exporting our structured `Error`/`Result` as primary types. `anyhow::Result` no longer re-exported here.

Haleshot:
I will be tending to the CI fails soon. 😓 (even w/ all the local building done)...
_Edit_: Let me know if I went outside the scope of the issue at hand (or if the changes made are not relevant). I made these changes incrementally, locally (after having read the issue outline a few times and the build errors that followed on making the changes as I deemed appropriate to "The Proposed Design"). 

Would appreciate any help in fixing changes that aren't relevant. I have also made the edit to the PR description here (with Edit 3).

rust/utils/src/error.rs
```diff
@@ -1,17 +1,327 @@
...
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        self.0.source()
+    }
+}
+
+impl From<anyhow::Error> for Error {
+    fn from(e: anyhow::Error) -> Self {
+        Error::Internal {
+            source: Box::new(AnyhowWrapper(e)),
+            bt: Backtrace::capture(),
```
georgeh0: `anyhow::Error` has its own backtrace. We can use it: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.backtrace

rust/utils/src/error.rs
```diff
@@ -1,17 +1,327 @@
...
+        match self {
+            Error::Context { source, .. } => Some(source.as_ref()),
+            Error::HostLang(e) => Some(e.as_ref()),
+            Error::Internal { source, .. } => Some(source.as_ref()),
+            Error::Client { .. } => None,
+        }
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
```
georgeh0: Actually we do this:

```rust
pub type Result<T, E = Error> = std::result::Result<T, E>;
```

(similar to [`anyhow::Result`](https://github.com/dtolnay/anyhow/blob/cce0579d85fd1f6352a5955a9c134fc8655c853d/src/lib.rs#L468))

So users can use both `Result<T>` (default) and `Result<T, OtherError>`.

rust/py_utils/src/error.rs
```diff
@@ -75,3 +128,13 @@ impl<T> AnyhowIntoPyResult<T> for anyhow::Result<T> {
         }
     }
 }
+
+pub trait CResultIntoPyResult<T> {
+    fn into_py_result(self) -> PyResult<T>;
+}
```
georgeh0: I think a different trait is not needed here. We can reuse the existing `IntoPyResult` trait, and implement it for `CResult<T>`.

This will make the migrations easier.

rust/cocoindex/src/py/mod.rs
```diff
@@ -47,22 +47,20 @@ fn init_pyo3_runtime() {
 
 #[pyfunction]
 fn init(py: Python<'_>, settings: Pythonized<Option<Settings>>) -> PyResult<()> {
-    py.detach(|| -> anyhow::Result<()> {
+    py_utils::CResultIntoPyResult::into_py_result(py.detach(|| -> Result<()> {
```
georgeh0: This is not the idiomatic way to call this trait method.

The main reason we put `into_py_result()` into a trait (rather than a standalone function) is to be able to call it as a method of `Result` (i.e. `xx.into_py_result()`), which is more concise and can be easily chained.

As long as you `use` the trait, you can directly call it as a member method.

See [this article](https://blog.masteringbackend.com/rust-extension-traits-in-5-minutes) for more explanation about this "extension trait pattern".

rust/utils/src/error.rs
Haleshot: Ahh good catch! Not sure how I missed that - was probably in the mindset w/ custom error layering in this PR. Fixed by making `bt: Option<Backtrace>`.

_(I kept it optional to preserve anyhow’s original bt without double‑capturing, while still capturing for locally created errors). Makes sense?_

rust/cocoindex/src/py/mod.rs
Haleshot: This was a quick workaround (that I failed to tend to before pushing finally) to get past the build errors locally, during dev. Thanks for the catch again; `.into_py_result()` throughout now. Thanks for the relevant article (good read).

rust/cocoindex/src/ops/targets/neo4j.rs
```diff
@@ -1070,13 +1070,17 @@ impl TargetFactoryBase for Factory {
                     for mut_with_ctx in muts.iter() {
                         let export_ctx = &mut_with_ctx.export_context;
                         for upsert in mut_with_ctx.mutation.upserts.iter() {
-                            export_ctx.add_upsert_queries(upsert, &mut queries)?;
+                            export_ctx
+                                .add_upsert_queries(upsert, &mut queries)
+                                .map_err(retryable::Error::not_retryable)?;
```
georgeh0: I think we want to implement `From` for the new error type, similar to this `From<anyhow::Error>` here:

https://github.com/cocoindex-io/cocoindex/blob/352295c69e598363f50afc029ca65436a5885c2e/rust/utils/src/retryable.rs#L82-L89

Then we don't need a manual `map_err` in the migration.

rust/utils/src/error.rs
georgeh0: Good question!

I think our purpose of keeping the backtrace fields is to capture these information when converting to Python exceptions, so people can see that for debugging. But the backtrace of `anyhow::Error` are not part of the error message (don't show up in `to_string()`), so we'll lose them.

However, I just notice `Backtrace` doesn't support `Clone`, so we have no way to use it to set the `bt` field any way.

Because of this, one pragmatic solution in my mind is to add a specific enum variation for `anyhow`. e.g. `InternalAnyhow(anyhow::Error)`: we treat it the same as internal errors, except it's represented by `anyhow::Error`. We won't need anther wrapper. In `backtrace()` method we can directly get the backtrace from it.

rust/cocoindex/src/ops/targets/neo4j.rs
Haleshot: Tthat's cleaner, yes, thank you. Makes sense so we can just use `?` directly thanthe manual `map_err`.

rust/cocoindex/src/ops/factory_bases.rs
```diff
@@ -397,8 +399,13 @@ impl<E: BatchedFunctionExecutor> batching::Runner for BatchedFunctionExecutorRun
     async fn run(
         &self,
         inputs: Vec<Self::Input>,
-    ) -> Result<impl ExactSizeIterator<Item = Self::Output>> {
-        Ok(self.0.evaluate_batch(inputs).await?.into_iter())
+    ) -> anyhow::Result<impl ExactSizeIterator<Item = Self::Output>> {
```
georgeh0, line 402: This should use the new error too, to preserve structured error information across the stack. I think `batching.rs` needs to adopt the new error type. But it's OK to deal with in a follow-up PR.

georgeh0 (approved):
Super awesome! This is a huge effort! Thanks for pushing it through!

One leftover comment: which can be resolved separately.

Approving and merging this PR.

badmonster0:
Thanks @Haleshot  love your PR!

PR #1383, opened 2025-12-14
include mypy type checking for the examples directory 

### LINKED ISSUE #1092: Update existing workflow and precommit hooks to also check the examples mypy type error
Update existing workflow and precommit hooks to also check the examples mypy type error

https://github.com/cocoindex-io/cocoindex/blob/main/.github/workflows/_test.yml

precommit hook
https://github.com/cocoindex-io/cocoindex/blob/main/.pre-commit-config.yaml

thisisharsh7:
fixes #1092

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,8 +56,8 @@ repos:
 
         - id: mypy-check
           name: mypy type check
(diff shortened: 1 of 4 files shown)


georgeh0 (approved):
Thanks a lot for adding this!

PR #1385, opened 2025-12-14
[FEATURE] Provide an option to force reprocessing during update #323

### LINKED ISSUE #323: [FEATURE] Provide an option to force reprocessing during update
When doing an update, not everything is reprocessed, due to:

* Reprocessing can be skipped if no source + logic version change.
* Intermediate function output can be cached and reused.

Sometimes users may need to force a reprocess on everything, e.g. something changed but we don't capture it. So we need to provide an option to force reprocessing during update.

prithvi-moonshot:
Implements new cli flag `--full-reprocess` in order to invalidate cache during reexport. 

closes #323

### DIFF
--- a/examples/image_search/main.py
+++ b/examples/image_search/main.py
@@ -140,7 +140,7 @@ app.mount("/img", StaticFiles(directory="img"), name="img")
 
 
 # --- Search API ---
-@app.get("/search")  # type: ignore[misc]
+@app.get("/search")  # type: ignore
 def search(
     q: str = Query(..., description="Search query"),
     limit: int = Query(5, description="Number of results"),


docs/yarn.lock
```diff
@@ -2,256 +2,269 @@
...
-  resolved "https://registry.yarnpkg.com/@algolia/client-analytics/-/client-analytics-5.45.0.tgz#4f58d6c1c8d43afeedfd8ef7b84b49e3e7cdff14"
-  integrity sha512-I3g7VtvG/QJOH3tQO7E7zWTwBfK/nIQXShFLR8RvPgWburZ626JNj332M3wHCYcaAMivN9WJG66S2JNXhm6+Xg==
-  dependencies:
-    "@algolia/client-common" "5.45.0"
-    "@algolia/requester-browser-xhr" "5.45.0"
-    "@algolia/requester-fetch" "5.45.0"
-    "@algolia/requester-node-http" "5.45.0"
+"@algolia/abtesting@1.3.0":
+  version "1.3.0"
+  resolved "https://registry.npmjs.org/@algolia/abtesting/-/abtesting-1.3.0.tgz"
```
georgeh0: These lock file changes are unexpected. We're using `yarn`, so by default use  `yarnpkg.com` instead of `npmjs.org`. Please revert change of this file.

rust/cocoindex/src/execution/source_indexer.rs
```diff
@@ -237,6 +237,7 @@ pub enum UpdateMode {
     #[default]
     Normal,
     ReexportTargets,
+    FullReprocess,
```
georgeh0, line 240: What `FullReprocess` achieves is roughly a superset of `ReexportTargets`, so for things that need to happen on `ReexportTargets` mode, they need to happen for `FullReprocess` too

https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex%20ReexportTargets&type=code

I think we can consider add a method `UpdateMode::needs_full_export()` (which returns `true` on both modes) and use this one in many places above.

prithvi-moonshot: Just added this in the latest commit. Let me know what you think!

prithvi-moonshot:
@georgeh0 so I was having problems pushing this commit due to a reason unrelated to this particular issue. I did a `git pull` a short while ago and one of new changes it brought was this example called `image-search` that has a type error. In order to push this commit to my remote branch I added a `# type: ignore[misc]` as that was the error code the type check was showing. Now a different type error is causing the GitHub pipeline test to fail. How should I proceed as this is an unrelated issue?

prithvi-moonshot:
@georgeh0 Hey, so I decided to fix the commit issue by just expanding the scope of `# type: ignore`  as I have seen it being used in other examples too. This commit is ready for review

badmonster0:
thanks for your contribution @prithvi-moonshot !

PR #1395, opened 2025-12-20
feat: Add GeneratedOutput enum for direct JSON returns 

### LINKED ISSUE #400: Allow `LlmGenerationClient::geneate()` directly return a JSON value

thisisharsh7:
fixes #400

### DIFF
--- a/rust/cocoindex/src/llm/anthropic.rs
+++ b/rust/cocoindex/src/llm/anthropic.rs
@@ -2,7 +2,7 @@ use crate::prelude::*;
 use base64::prelude::*;
 
 use crate::llm::{
-    LlmGenerateRequest, LlmGenerateResponse, LlmGenerationClient, OutputFormat,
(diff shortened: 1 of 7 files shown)


badmonster0:
appreciate the quick work @thisisharsh7 !

badmonster0:
thanks for the great PR @thisisharsh7 !

thisisharsh7:
thanks !!

Issue #1396, opened 2025-12-21
[BUG] cargo test crashes with ModuleNotFoundError: encodings when developing with uv-managed Python


skprasadu:
**Describe the bug**

When running cargo test (via pre-commit run --all-files) inside a repo checkout that’s using a uv‑managed Python virtualenv, the Rust test binary crashes during Python initialization:
	•	sys.prefix / sys.base_prefix resolve to /install
	•	Python cannot locate the stdlib and fails with ModuleNotFoundError: No module named 'encodings'
	•	This aborts cargo test and therefore fails the pre-commit hook cargo-test

This looks like a Python interpreter discovery / stdlib-path issue when Python is embedded (the executable is the Rust test binary, not the Python executable), and the uv-managed Python distribution doesn’t resolve its stdlib automatically in that scenario.

Note: maturin develop, pytest, ruff, mypy pass; only the Rust test stage crashes.

**To Reproduce**
	1.	Clone the repo
	
```bash
git clone https://github.com/cocoindex-io/cocoindex.git
cd cocoindex
```
	2.	Create and activate a uv venv (example uses Python 3.12):

```bash
uv venv --python 3.12
source .venv/bin/activate
```

	3.	Install dev tools + build the extension:

```bash
uv pip install maturin pre-commit
maturin develop -E all,dev
pre-commit install
```

	4.	Run pre-commit (or run cargo tests directly):

```bash
pre-commit run --all-files
# or
cargo test -p cocoindex --lib -- --nocapture
```

	5.	Observe crash from the Rust test binary during Python initialization with:
	•	“Could not find platform independent libraries ”
	•	stdlib dir = '/install/lib/python3.12'
	•	ModuleNotFoundError: No module named 'encodings'

**Expected behavior**

cargo test should run the Rust unit tests successfully (or fail with normal test assertions), not crash during embedded Python initialization due to missing stdlib (encodings).

**Screenshots**

Not applicable. (I can paste full logs; the key crash excerpt is below.)

<details>
<summary>Crash excerpt</summary>

Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python3'
  ...
  stdlib dir = '/install/lib/python3.12'
  sys.executable = '.../target/debug/deps/cocoindex_engine-...'
  sys.prefix = '/install'
  ...
[...]

PR #1398, opened 2025-12-21
dev: wrap cargo test to avoid embedded Python stdlib discovery crash

skprasadu:
•	What: Adds dev/run_cargo_test.sh, updates pre-commit cargo-test hook to use it, adds docs workaround.
	•	Why: cargo test can crash with ModuleNotFoundError: encodings when using uv-managed Python (embedded Python reports sys.prefix=/install and can’t locate stdlib).
	•	Scope: Dev-only. No runtime/engine behavior change.
	•	Refs: Refs [1396](https://github.com/cocoindex-io/cocoindex/issues/1396)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
 
         - id: cargo-test
           name: cargo test
-          entry: cargo test
+          entry: dev/run_cargo_test.sh
           language: system
           files: ^(rust/|Cargo\.toml)
           pass_filenames: false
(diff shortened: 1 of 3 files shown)


skprasadu:
Thanks @georgeh0 for such quick response

badmonster0:
@skprasadu thank you so much for contributing to cocoindex! we made a section for you in latest release note https://cocoindex.io/blogs/changelog-0311-0326#skprasadu
Love your work!!

PR #1408, opened 2025-12-23
Warn when live update pass exceeds refresh_interval

### LINKED ISSUE #267: [FEATURE] stats for #runs and runtime for a update for live mode setting update frequency
**What is the use case?**
for example i'm setting it to 10 sec but it takes 1 min, and it didn't ran up to my expectations

Would be nice to expose the stats to users

**Additional context**

skprasadu:
fixes #267

## What
Adds a warning when an interval live update pass takes longer than refresh_interval,
so interval updates will lag behind.

## Why
When refresh_interval is small, users may not realize the updater is falling behind.
This warning makes lag visible.

## How
- Added warn_if_overrun(fut, threshold, warn_fn) helper
- Wired threshold into interval update passes only (not initial pass)
- Keeps MissedTickBehavior::Delay

## Testing
- cargo test (unit tests in live_updater.rs)
- dev/run_cargo_test.sh warn_if_overrun
- E2E: examples/live_updates with a large file triggers WARN:
  "Live update pass exceeded refresh_interval; interval updates will lag behind"

## Notes
- `cargo clippy -p cocoindex --all-targets -- -D warnings` fails on main due to
  `clippy::needless_return` in `rust/utils/src/batching.rs:218` (unrelated).

### DIFF
--- a/rust/cocoindex/src/execution/live_updater.rs
+++ b/rust/cocoindex/src/execution/live_updater.rs
@@ -292,8 +292,6 @@ impl SourceUpdateTask {
                 )
                 .await;
 
-                // IMPORTANT:
-                // If refresh_interval is None, we're done after the initial pass.
                 let Some(refresh_interval) = refresh_interval else {
                     return Ok(());
                 };
@@ -455,7 +453,6 @@ impl SourceUpdateTask {
         let start_time = std::time::Instant::now();
         let update_stats = Arc::new(stats::UpdateStats::default());
 
-        // No overrun warning here anymore (founder wants it at the interval loop layer).
         let update_fut = source_indexing_context.update(&update_stats, update_options);
 
         self.run_with_progress_report(


rust/cocoindex/src/execution/live_updater.rs
```diff
@@ -424,13 +447,40 @@ impl SourceUpdateTask {
...
+
+            if let Some(threshold) = warn_overrun_after {
+                warn_if_overrun(fut, threshold, move || {
+            let elapsed = started_at.elapsed();
+            warn!(
+                flow_name = %flow_name,
+                source_name = %source_name,
+                update_title = %update_title_owned,
+                refresh_interval_secs = threshold.as_secs_f64(),
+                elapsed_secs = elapsed.as_secs_f64(),
```
georgeh0: these two are expected to be the same. We only need to log one.

rust/cocoindex/src/execution/live_updater.rs
```diff
@@ -424,13 +447,40 @@ impl SourceUpdateTask {
...
+        warn_overrun_after: Option<std::time::Duration>,
     ) -> Result<()> {
         let start_time = std::time::Instant::now();
         let update_stats = Arc::new(stats::UpdateStats::default());
 
         // Check update result
+        let flow_name = self.flow.flow_instance.name.clone();
+        let source_name = self.import_op().name.clone();
+        let update_title_owned = update_title.to_string();
+        let started_at = start_time; // Instant is Copy
```
georgeh0: Seems these `clone()` are not necessary. The futures below are local, and `&self` outlives them. We can directly access them below.

rust/cocoindex/src/execution/live_updater.rs
```diff
@@ -424,13 +447,40 @@ impl SourceUpdateTask {
...
+
+        let update_fut = async {
+            let fut = source_indexing_context.update(&update_stats, update_options);
+
+            if let Some(threshold) = warn_overrun_after {
+                warn_if_overrun(fut, threshold, move || {
```
georgeh0: I feel the code will be simpler without this helper function.

I'm even thinking doing it in the layer of managing interval ticks is cleaner and simpler. e.g. the code [here](https://github.com/cocoindex-io/cocoindex/blob/c02e81699119559d946460d08787c89f05e3ce57/rust/cocoindex/src/execution/live_updater.rs#L310-L319) can be changed to something like this:

```rust
loop {
  .... 

  let mut update_fut = task.update_one_pass_with_error_logging(...);
  tokio::select! {
   _ = &mut update_fut { }
   _ = tokio::time::sleep(refresh_interval) {
      // Overrun, warn and finish  
      warn!(...);
      update_fut.await;
    }
  }
}
```

What do you think?

skprasadu:
@georgeh0 , updated to move the overrun warning into the interval tick loop (tokio::select! between update and sleep(refresh_interval)), removed the helper, and log only refresh_interval_secs (no elapsed_secs). Confirmed end-to-end by appending to a large file and verifying Postgres row length/tail updates; warning appears when refresh interval is too small. I test it.

skprasadu:
@georgeh0 , can you check the PR now and is it ok?

georgeh0 (approved):
Awesome! Thanks for the PR!

PR #1425, opened 2025-12-27
`ci`: add lychee link-checker workflow file

Haleshot:
Closes #1415

Adds a GitHub Actions workflow that uses [`lychee`](https://github.com/lycheeverse/lychee) to catch broken links in the docs before they go live. It runs on PRs that touch any `.md` or `.mdx` files under [`/docs`](https://github.com/cocoindex-io/cocoindex/tree/main/docs), plus there's a daily scheduled run at midnight UTC to catch ext-links that might've gone stale/outdated over time.

When lychee finds broken links, it'll automatically open an issue with an appropriate report for the same.

Kept it scoped to just [`/docs`](https://github.com/cocoindex-io/cocoindex/tree/main/docs) since `examples/` got moved to their [own repo](https://github.com/cocoindex-io/examples) in #1412. If we want to add this over there too at some point, it's pretty much copy-paste.

_PS: My first time trying out claude-code (which is why you see the `co-authored` tag); given the inclusion of `CLAUDE.md` in [`6804ea6`](https://github.com/cocoindex-io/cocoindex/pull/1417/commits/6804ea698703371bcf8481063110e88e94efd0db) and the (optional) installation recommendation of [claude-code-skills in docs](https://cocoindex.io/docs/getting_started/installation#-install-claude-code-skill-optional), I thought it would be alright to experiment with this (since this is a small PR). Hope that's alright! I anyway put up disclaimers for when AI-assisted code is used in my PRs._

PR #1433, opened 2025-12-28
`ci`: add `dependabot` and improve link-checker config

Haleshot:
Follow-up to https://github.com/cocoindex-io/cocoindex/pull/1425 to reduce false positives (FPs) in the daily link checker reports. I got notified of FPs in my fork's actions [here](https://github.com/Haleshot/cocoindex/actions/runs/20547227693) flagging a bunch of auth/console URLs and localhost links that are perfectly valid (resp. to their context in the webpages) in our docs but fail in CI.                                                                                                                                                                                      

Added a `.lycheeignore` file to exclude such (problematic) patterns so the checker only fails on actual broken links. Also included dependabot as per https://github.com/cocoindex-io/cocoindex/pull/1425#issuecomment-3694455227 that you replied positively to. 

Also added a links badge (as shown in the [workflow action description](https://github.com/marketplace/actions/lychee-broken-link-checker#fancy-badge)) in the README for quick broken links glancing.

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Run link checker
         id: lychee
-        uses: lycheeverse/lychee-action@v2
+        uses: lycheeverse/lychee-action@a8c4c7cb88f0c7386610c35eb25108e448569cb0 # v2
         with:
           args: --cache --max-cache-age 1d --verbose --no-progress 'docs/**/*.md' 'docs/**/*.mdx'
           fail: true


.github/workflows/links.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - name: Run link checker
         id: lychee
-        uses: lycheeverse/lychee-action@v2
+        uses: lycheeverse/lychee-action@a8c4c7cb88f0c7386610c35eb25108e448569cb0 # v2
```
Haleshot, line 35: Am I overthinking this 😅 

I figured it would be important since it was documented [here](https://github.com/marketplace/actions/lychee-broken-link-checker#security-and-updates) which references this [report](https://francoisbest.com/posts/2020/the-security-of-github-actions).

.lycheeignore
Haleshot, line 1: Usually don't hardcode paths/links, etc. but felt it necessary to do so, as these are pretty specific to the text present on the docs webpages; and to prevent them from appearing on the reports such as [this](https://github.com/Haleshot/cocoindex/actions/runs/20547227693#summary-59019437207).

PR #1444, opened 2025-12-29
Feat/specific db schema for internal tables

Geoff-Robin:
### Database Schema Improvements
* **Added schema as a field** – Added optional `schema` to `DatabaseConnectionSpec` (Python + Rust) and updated settings/tests to support per-connection schema.
* **Creating schema in case it does not exist** – Add logic to `CREATE SCHEMA IF NOT EXISTS` for the configured schema and wire schema-aware setup checks.
* **Adding public to search path** – Configure PostgreSQL `search_path` to `"schema,public"` when a custom schema is used.

### Configuration & Environment
* **Setting schema_name from .env** – Load `COCOINDEX_DATABASE_SCHEMA` into Python settings so the schema can be set via environment.
* **Fixed white character space in .env.lib_debug** – Remove stray space so `COCOINDEX_PYTHON` is assigned correctly.

### Branch Maintenance
* **Merge branch 'cocoindex-io:main' (Dec 29)** – Sync/resolve changes from upstream main into the feature branch.
* **Merge branch 'cocoindex-io:main' (Dec 28)** – Sync feature branch with upstream main.

### DIFF
--- a/rust/utils/src/db.rs
+++ b/rust/utils/src/db.rs
@@ -1,6 +1,30 @@
+use crate::prelude::*;
+use sqlx::PgPool;
 pub enum WriteAction {
     Insert,
     Update,
+}
+
+pub async fn ensure_schema_from_search_path(pool: &PgPool) -> Result<()> {
+    let search_path: String = sqlx::query_scalar("SHOW search_path")
+        .fetch_one(pool)
+        .await?;
+    // Parse the first schema that is not pg_catalog or information_schema
+    let schema_opt = search_path
+        .split(',')
+        .map(|s| s.trim().trim_matches('"'))
+        .find(|s| *s != "pg_catalog" && *s != "information_schema");
+    if let Some(schema) = schema_opt {
+        let query = format!(
+            "CREATE SCHEMA IF NOT EXISTS \"{}\"",
+            schema.replace('"', "\"\"")
+        );
+        sqlx::query(&query)
+            .execute(pool)
+            .await
+            .with_context(|| format!("Failed to create schema `{}`", schema))?;
+    }
+    Ok(())
 }
 
 pub fn sanitize_identifier(s: &str) -> String {
--- a/rust/cocoindex/src/execution/db_tracking_setup.rs
+++ b/rust/cocoindex/src/execution/db_tracking_setup.rs
@@ -3,32 +3,7 @@ use crate::prelude::*;
 use crate::setup::{CombinedState, ResourceSetupChange, ResourceSetupInfo, SetupChangeType};
 use serde::{Deserialize, Serialize};
 use sqlx::PgPool;
-
-pub async fn create_schema(pool: &PgPool) -> Result<()> {
-    let search_path: String = sqlx::query_scalar("SHOW search_path")
-        .fetch_one(pool)
-        .await?;
-
-    // Parse the first schema that is not pg_catalog or information_schema
-    let schema_opt = search_path
-        .split(',')
-        .map(|s| s.trim().trim_matches('"'))
-        .find(|s| *s != "pg_catalog" && *s != "information_schema");
-
-    if let Some(schema) = schema_opt {
-        let query = format!(
-            "CREATE SCHEMA IF NOT EXISTS \"{}\"",
-            schema.replace('"', "\"\"")
-        );
-
-        sqlx::query(&query)
-            .execute(pool)
-            .await
-            .with_context(|| format!("Failed to create schema `{}`", schema))?;
-    }
-
-    Ok(())
-}
+use utils::db;
 
 pub fn default_tracking_table_name(flow_name: &str) -> String {
     format!(
@@ -57,7 +32,7 @@ async fn upgrade_tracking_table(
             .has_fast_fingerprint_column
             .then(|| "processed_source_fp BYTEA,")
             .unwrap_or("");
-        create_schema(&pool).await?;
+        db::ensure_schema_from_search_path(&pool).await?;
         let query =  format!(
             "CREATE TABLE IF NOT EXISTS {table_name} (
                 source_id INTEGER NOT NULL,
--- a/rust/cocoindex/src/setup/db_metadata.rs
+++ b/rust/cocoindex/src/setup/db_metadata.rs
@@ -3,7 +3,7 @@ use crate::prelude::*;
 use super::{ResourceSetupChange, ResourceSetupInfo, SetupChangeType, StateChange};
 use axum::http::StatusCode;
 use sqlx::PgPool;
-use utils::db::WriteAction;
+use utils::db::{WriteAction, ensure_schema_from_search_path};
 
 const SETUP_METADATA_TABLE_NAME: &str = "cocoindex_setup_metadata";
 pub const FLOW_VERSION_RESOURCE_TYPE: &str = "__FlowVersion";
@@ -362,6 +362,7 @@ impl MetadataTableSetup {
         }
         let lib_context = get_lib_context().await?;
         let pool = lib_context.require_builtin_db_pool()?;
+        ensure_schema_from_search_path(&pool).await?;
         let query_str = format!(
             "CREATE TABLE IF NOT EXISTS {SETUP_METADATA_TABLE_NAME} (
                 flow_name TEXT NOT NULL,


python/cocoindex/setting.py
```diff
@@ -35,6 +35,7 @@ class DatabaseConnectionSpec:
     url: str
     user: str | None = None
     password: str | None = None
+    schema: str | None = None
```
georgeh0: `schema` is not part of db connection (we don't need to create different db connection because of different db schema). Also `DatabaseConnectionSpec` is also shared by Postgres target, which already supports `schema` in other fields. So I think we want to make it a new field within `Settings` for this (e.g. `db_schema_name`)

rust/cocoindex/src/lib_context.rs
```diff
@@ -171,18 +171,35 @@ pub fn get_auth_registry() -> &'static Arc<AuthRegistry> {
...
-type PoolKey = (String, Option<String>);
+type PoolKey = (String, Option<String>, Option<String>);
 type PoolValue = Arc<tokio::sync::OnceCell<PgPool>>;
 
 #[derive(Default)]
 pub struct DbPools {
     pub pools: Mutex<HashMap<PoolKey, PoolValue>>,
 }
 
+pub async fn create_schema(schema_name: &str, pool: &PgPool) -> Result<()> {
```
georgeh0: I think `execution/db_tracking_setup.rs` is a better place for this function.

rust/cocoindex/src/lib_context.rs
```diff
@@ -300,7 +320,9 @@ pub async fn create_lib_context(settings: settings::Settings) -> Result<LibConte
     let db_pools = DbPools::default();
     let persistence_ctx = if let Some(database_spec) = &settings.database {
         let pool = db_pools.get_pool(database_spec).await?;
-        let all_setup_states = setup::get_existing_setup_state(&pool).await?;
+        let schema_name = database_spec.schema.as_deref().unwrap_or("public");
+        create_schema(schema_name, &pool).await?;
```
georgeh0: - We don't want to run this `CREATE SCHEMA` statement each time when establishing the connection. I think we only need to call it before we need to create the tracking table (around [here](https://github.com/cocoindex-io/cocoindex/blob/9fe9d7cf5d44b701860ddd3bc2b41b41ed6320a4/rust/cocoindex/src/execution/db_tracking_setup.rs#L271))
 
- We don't want to hardcode to `public` when schema is not specified. The schema doesn't have to be `public` in this case (e.g. users may configure another default schema based on their db user). Let's just skip creating the schema if not specified.

.env.lib_debug
```diff
@@ -1,5 +1,6 @@
 export RUST_LOG=warn,cocoindex_engine=trace,tower_http=trace
 export RUST_BACKTRACE=1
+export COCOINDEX_PYTHON="python"
```
georgeh0: Thanks for adding this fix. But I don't quite understand the specific problem it solves.

Currently `COCOINDEX_PYTHON` is only used in [`run_cargo_test.sh`](https://github.com/cocoindex-io/cocoindex/blob/9fe9d7cf5d44b701860ddd3bc2b41b41ed6320a4/dev/run_cargo_test.sh#L8-L11) and the purpose is to allow users to override the python interpreter.  `python` is not always the right interpreter to use. e.g. when `COCOINDEX_PYTHON` is not set, currently we default to `.venv/bin/python` which is a more reasonable default.

You mention "white character space". Where is the "white character space" specifically, and why is this change related to it?

Geoff-Robin: My mistake! I got mixed up between my local debug settings and the repo defaults. The 'whitespace error' was actually just a typo I made on my end while setting up that export. You're right that we should stick with the .venv defaul, I'll revert this change now.

rust/cocoindex/src/lib_context.rs
Geoff-Robin: > We don't want to run this CREATE SCHEMA statement each time when establishing the connection. I think we only need to call it before we need to create the tracking table (around [here](https://github.com/cocoindex-io/cocoindex/blob/9fe9d7cf5d44b701860ddd3bc2b41b41ed6320a4/rust/cocoindex/src/execution/db_tracking_setup.rs#L271))

I agree, i think i'll move it around there.

> We don't want to hardcode to public when schema is not specified. The schema doesn't have to be public in this case (e.g. users may configure another default schema based on their db user). Let's just skip creating the schema if not specified.

if it’s fine, I can set the search path to `schema_name, $user, public`, considering we still need a default schema for resolving pgvector types when the extension is installed in the default schema.

Geoff-Robin: I can't remove the schema as it's needed [here](https://github.com/cocoindex-io/cocoindex/pull/1444/changes#r2651337022) as well as [this place](https://github.com/cocoindex-io/cocoindex/pull/1444/changes#diff-d1a40d96e3868552a4062bab8971fd9d6a8af5b6a6df36bd51df40915b367cf6R41)
Although, above changes will address the concern

python/cocoindex/setting.py
Geoff-Robin: @georgeh0 I've just made changes hope they are to your liking

georgeh0: Thanks for updating! After the latest change, it's still part of the `DatabaseConnectionSpec`. As I explained above, it's not expected. Please move it out to `Settings`. 

Besides, for the name, I prefer `db_schema_name` instead of `internal_schema`, and use `COCOINDEX_DATABASE_SCHEMA_NAME` for the environment variable name. Because the term "schema" is too generic (in most situations it means structure of the database / tables, e.g. what columns a table have). `db_schema_name` will be more clear.

Thanks!

Geoff-Robin: @georgeh0 Hey I just removed it and renamed it to db_schema_name. Hope it's to your liking!

georgeh0: My comment was not only bout the name. I also suggested moving it out of `DatabaseConnectionSpec`. Please read my comments carefully and make the change. Thanks!

rust/cocoindex/src/lib_context.rs
Geoff-Robin: I'm opening this conversation again, as i have just noticed calling this 
- Apply changes to tracking tables which would be created if it did not exist in the 'specified' schema but if that schema hasn't been created, search path will automatically cause it to create in user's default schema. 

I will be adding a call for that in db_metadata.rs as well to ensure schema is created.

Geoff-Robin:
@georgeh0 Hey, changes have been made. Can you check if there's anything missing?

rust/utils/src/db.rs
```diff
@@ -1,8 +1,32 @@
...
 
+pub async fn ensure_schema_from_search_path(pool: &PgPool) -> Result<()> {
+    let search_path: String = sqlx::query_scalar("SHOW search_path")
+        .fetch_one(pool)
+        .await?;
+    // Parse the first schema that is not pg_catalog or information_schema
+    let schema_opt = search_path
+        .split(',')
+        .map(|s| s.trim().trim_matches('"'))
+        .find(|s| *s != "pg_catalog" && *s != "information_schema");
```
georgeh0, line 16: Why do we need to create schema based on search path? I don't think we need to do so. We only need to create it based on db schema name specified in `Settings`.

rust/cocoindex/src/lib_context.rs
```diff
@@ -195,7 +196,10 @@ impl DbPools {
                 if let Some(password) = &conn_spec.password {
                     pg_options = pg_options.password(password);
                 }
-
+                if let Ok(schema) = env::var("COCOINDEX_DATABASE_SCHEMA_NAME") {
+                    let path = format!("\"{}\",\"$user\",public", schema);
+                    pg_options = pg_options.options([("search_path", path.as_str())]);
+                }
```
georgeh0, line 202: This is unexpected. As I said in the other comment, we don't want to treat db_schema as connection-level property. e.g. the same connection can be shared to access both the internal table and the target table,  and it's perfectly fine for them to use different schema.

Also we don't directly read env variable from Rust core.  As I said in the other comment, they should be wired through `Settings` ([Python type](https://github.com/cocoindex-io/cocoindex/blob/b16dccf1cc9adc87f15d50208848dc6d9b3f670d/python/cocoindex/setting.py#L75), [Rust type](https://github.com/cocoindex-io/cocoindex/blob/e476fa2fe722724b8eaf2fa1340ff5aaaf2b7d15/rust/cocoindex/src/settings.rs#L19)).

rust/cocoindex/src/lib_context.rs
```diff
@@ -300,7 +304,8 @@ pub async fn create_lib_context(settings: settings::Settings) -> Result<LibConte
     let db_pools = DbPools::default();
     let persistence_ctx = if let Some(database_spec) = &settings.database {
         let pool = db_pools.get_pool(database_spec).await?;
-        let all_setup_states = setup::get_existing_setup_state(&pool).await?;
+        let schema_name = env::var("COCOINDEX_DATABASE_SCHEMA_NAME")?;
+        let all_setup_states = setup::get_existing_setup_state(&schema_name, &pool).await?;
```
georgeh0, line 308: Same here, this is unexpected. Please don't treat schema as a connection level property, and please don't directly read env variable from Rust side.

Geoff-Robin:
I'll be closing this PR as I'm trying a different approach,
New PR will be out soon.

Issue #1415, opened 2025-12-26
Add `lychee` link-check (GitHub Action workflow) for docs


Haleshot:
**What is the use case?**
Right now there’s no automated way to catch broken links across the CocoIndex docs and example projects before they ship. This makes it easy for stale links to creep into the site and/or READMEs, especially as external docs and blog posts are added (or maybe moved around).
_Edit: Just saw in this PR: https://github.com/cocoindex-io/cocoindex/pull/1412 that examples got ported over to the [examples repo](https://github.com/cocoindex-io/examples). Might make sense to add it in the there too then?_

**Describe the solution you'd like**
Set up a GitHub Actions workflow using [`lycheeverse/lychee-action`](https://github.com/lycheeverse/lychee-action) to scan relevant files in this repo on a schedule and on relevant pushes/PRs.  The workflow can run `lychee` against the docs and examples (and directories explicitly specified), write an appropriate `.md` report, and optionally [open an automated issue](https://github.com/marketplace/actions/lychee-broken-link-checker) when it finds broken links. 
_(Also, pinning the action to a tagged version and wiring it into Dependabot updates would keep it secure?)_

**Additional context**
Check out [`lychee`](https://github.com/lycheeverse/lychee) for more context; the gist is: it’s fast, async, understands Markdown/HTML out of the box, etc. 

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

Haleshot:
Would love to work on a PR for the same based on feedback on this issue proposal.

georgeh0:
That's a great idea, @Haleshot. Assigned.

About the trigering, how about:
- Run on PRs touching any MD/MDX files under `/docs`
- Also run on a daily schedule

Haleshot:
Makes sense, will add and configure the workflow files accordingly.

georgeh0:
I noticed there're still a lot of false positives in [the latest report](https://github.com/cocoindex-io/cocoindex/actions/runs/20566376385/job/59065278951).

Many are "local files", referenced by document mdx. Such as:

```
[...]

PR #1445, opened 2025-12-29
`ci`: reduce link checker false positives

Haleshot:
_Follow-up_ to #1425 to reduce false positives in the daily link checker. The [`latest report`](https://github.com/cocoindex-io/cocoindex/actions/runs/20566376385/job/59065278951) showed a bunch of (internal) doc errors that should not be coming up (all valid URLs) and docusaurus already checks for their validity during build and docs-ci.

Change:
Add to `.lycheeignore` to exclude internal `file://` links (relevant comment: https://github.com/cocoindex-io/cocoindex/issues/1415#issuecomment-3695581996) which should resolve the "errors" that show up in the reports.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -20,4 +20,3 @@
 ^https://aws\.amazon\.com/console/
 
 ^https://platform\.openai\.com/api/ # may require auth
-


georgeh0 (approved):
Thanks for the quick fix!

BTW, the check is not running for this PR. We may want to update the workflow to make updates on [`.lycheeignore‎`](https://github.com/cocoindex-io/cocoindex/pull/1445/changes#diff-fafe8d160920b6bd97b3c5235607969496ac0550eb4eeb9833393711fc210f18) also trigger the workflow.

(it's not blocking this PR. I'll merge right now to fix the broken badge on main)

PR #1452, opened 2025-12-29
`ci`: add additional params to `links.yml` for appropriate checking of links

Haleshot:
_Follow-up_ to #1445 to concerning #1415

- Added lance explicitly to the `.lycheeignore` file (not sure why a valid link gets flagged as 404; might have to report this upstream in the action workflow's [GitHub](https://github.com/lycheeverse)).
- Added `--accept` params to the `links.yml` workflow to keep link-checking valid and relevant.

Hope you can manually run the workflow after this and it passes. Thanks for your patience thus far :)

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -6,6 +6,15 @@ on:
     paths:
       - "docs/**/*.md"
       - "docs/**/*.mdx"
+      - ".lycheeignore"
+      - ".github/workflows/links.yml"
+  push:
+    branches: [main]
+    paths:
+      - "docs/**/*.md"
+      - "docs/**/*.mdx"
+      - ".lycheeignore"
+      - ".github/workflows/links.yml"
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:


.github/workflows/links.yml
georgeh0, line 8: As I commented [here](https://github.com/cocoindex-io/cocoindex/pull/1445#pullrequestreview-3616200520), can we also trigger the workflow on changes of `.lycheeignore` (and also `links.yml`)? so we can see the report immediately rather than having multiple arounds or PR merge.

Besides, I think we actually also want to enable it for `push` to `main`. This triggers another run when each related PR is merged into `main`, and actually update the status of the badge.

You can follow the convention for our most other workflows ([example](https://github.com/cocoindex-io/cocoindex/blob/9fe9d7cf5d44b701860ddd3bc2b41b41ed6320a4/.github/workflows/CI.yml#L9-L23))

.lycheeignore
```diff
@@ -18,5 +18,6 @@
 ^https://us-east-1\.console\.aws\.amazon\.com/
 ^https://cloud\.google\.com/
 ^https://aws\.amazon\.com/console/
+^https://github\.com/lancedb/lance/issues/
```
georgeh0: I looked at the link again. The original link actually redirects to `https://github.com/lance-format/lance/issues/4034`. so we want to indeed update the URL itself (somehow when I use `curl` I get 404 rather than a redirect). So we don't need add it here.

georgeh0 (commented):
Thanks for the quick fix!

.github/workflows/links.yml
Haleshot: Sorry, I missed the first comment in the merged PR. I'll tend to these changes suggested! Thanks for listing the `CI.yml` file.

.lycheeignore
Haleshot: Yup, same for me. I suppose the repo was moved to the `lance-format` org; wonder why curl (and I suppose the action workflow is built like this too to test URLs) fails to redirect.

TIL: Apparently, a common issue with GitHub - they sometimes return 404/403 responses to automated tools for [rate limiting ](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#rate-limit-errors)or bot detection, even though the links work fine in browsers.

georgeh0 (approved):
It passes now. Thanks!

PR #1454, opened 2026-01-01
Feat: Make target deletion (drop) tolerate failures

### LINKED ISSUE #1379: [FEATURE] make target deletion (drop) tolerate failures
When we delete (drop) a target backend, sometimes it fails because the connection is no longer available. Currently this may block flow's setup process.

We want to make target deletion (drop) tolerate failures. Specifically:

- Add a new field `ignore_target_drop_failures` to [`Settings`](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/setting.py#L75-L80). Use env variable `COCOINDEX_IGNORE_TARGET_DROP_FAILURES` as the default value.
- When `ignore_target_drop_failures` is set to `true`, when there's any error in dropping a target backend, log an error instead of popping up the error which terminates the process.
- When `ignore_target_drop_failures` is set to `false`, in the error message created during dropping a target backend, put a hint to let users know about the `COCOINDEX_IGNORE_TARGET_DROP_FAILURES` option.

---

prithvi-moonshot:
# Add option to ignore target backend drop failures

## Problem
Target backend drops can fail when connections are unavailable, blocking the flow setup/cleanup process.

## Solution
Added `ignore_target_drop_failures` setting (env: `COCOINDEX_IGNORE_TARGET_DROP_FAILURES`):
- When `true`: Drop failures are logged but don't block execution
- When `false` (default): Drop failures terminate with a helpful hint about the option

## Changes
- Added setting field to Python and Rust `Settings`
- Implemented error tolerance in `driver.rs` during target deletion
- Error messages include actionable hints when flag is disabled

Closes #1379

### DIFF
--- a/rust/cocoindex/src/setup/driver.rs
+++ b/rust/cocoindex/src/setup/driver.rs
@@ -648,10 +648,10 @@ async fn apply_changes_for_flow(
             write,
             resources.into_iter(),
             |targets_change| async move {
-                    let factory = get_export_target_factory(target_kind).ok_or_else(|| {
+                let factory = get_export_target_factory(target_kind).ok_or_else(|| {
                     internal_error!("No factory found for target kind: {}", target_kind)
-                        })?;
-                    for target_change in targets_change.iter() {
+                })?;
+                for target_change in targets_change.iter() {
                     for delete in target_change.setup_change.attachments_change.deletes.iter() {
                         delete.apply_change().await?;
                         }


rust/cocoindex/src/setup/driver.rs
```diff
@@ -514,7 +514,7 @@ async fn maybe_update_resource_setup<
     K: 'a,
     S: 'a,
     C: ResourceSetupChange,
-    ChangeApplierResultFut: Future<Output = Result<()>>,
+    ChangeApplierResultFut: Future<Output = Result<(), Error>>,
```
prithvi-moonshot: Hi @georgeh0, as you review this PR, could you take a look at the line above? I made the error type explicit in the ChangeApplierResultFut bound to resolve a type inference issue. This doesn’t change behavior, just clarifies the constraint that was already intended. Happy to adjust if you see any concerns.

rust/cocoindex/src/setup/driver.rs
```diff
@@ -645,33 +647,62 @@ async fn apply_changes_for_flow(
...
-                let factory = get_export_target_factory(target_kind).ok_or_else(|| {
-                    internal_error!("No factory found for target kind: {}", target_kind)
-                })?;
-                for target_change in targets_change.iter() {
-                    for delete in target_change.setup_change.attachments_change.deletes.iter() {
-                        delete.apply_change().await?;
+            |targets_change| {
+                let target_kind = target_kind.to_string();
+                let flow_instance_ctx = flow_ctx.flow.flow_instance_ctx.clone();
+                let flow_name = flow_ctx.flow_name().to_string();
```
georgeh0: These lines of changes seem unrelated and can be reverted? We don't need one more layer of block outside the `async move {}`, and don't need these additional clones.

rust/cocoindex/src/setup/driver.rs
```diff
@@ -645,33 +647,62 @@ async fn apply_changes_for_flow(
...
+                            tracing::error!("Ignoring target drop failure for kind '{}' in flow '{}': {:#}",
+                                &target_kind, &flow_name, e);
+                            return Ok::<(), Error>(());
+                        }
+                        if is_deletion {
+                            tracing::error!(
+                                "{}\n\nHint: set COCOINDEX_IGNORE_TARGET_DROP_FAILURES=true to ignore target drop failures.",
+                                e
+                            );
+                            return Err(e);
```
georgeh0: nit: this line of return is redundant

rust/cocoindex/src/setup/driver.rs
georgeh0: Thanks for heads up. This change itself looks like a no-op to me, given the default `E` type for `Result<T, E>` is already `Error` (as defined [here](https://github.com/cocoindex-io/cocoindex/blob/b16dccf1cc9adc87f15d50208848dc6d9b3f670d/rust/utils/src/error.rs#L48)). I'm just curious which type inference issue you encountered without this change.

rust/cocoindex/src/setup/driver.rs
prithvi-moonshot: Removed!

rust/cocoindex/src/setup/driver.rs
prithvi-moonshot: Reverted

rust/cocoindex/src/setup/driver.rs
prithvi-moonshot: @georgeh0 Hey, you're right it was a no-op, so I reverted this back to ``Result<()>`` , and I no longer get the type inference error. The error I had received was basically something like expected  ``Result<(), Error> `` but received  ``Result<(), ()> ``, which I think was probably caused by ambiguous closures. This line ``Ok::<(), Error>(())`` must have fixed it, but I thought the fix was ``Result<(), Error>`` . Should be good now!

georgeh0 (approved):
thanks for the PR!

Issue #1411, opened 2025-12-24
[FEATURE] Allow users to specify db schema for cocoindex internal table


georgeh0:
Postgres supports [schema](https://www.postgresql.org/docs/current/ddl-schemas.html) to isolate different tables. Currently we create internal tables in default schema. We want to allow users to control db schema for internal tables.

BTW, this is already supported by the builtin postgres target ([document](https://cocoindex.io/docs/targets/postgres#spec)). #1138 is the PR to support it.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

georgeh0:
Before we have this feature, there's a workaround by using `search_string` option in [Postgres connection string](https://www.postgresql.org/docs/current/libpq-connect.html).

e.g. if we want to use `coco` as the schema, the connection string can be set to:

```
COCOINDEX_DATABASE_URL=postgres://cocoindex:cocoindex@localhost/cocoindex?options=-csearch_path%3Dcoco,public
```

Geoff-Robin:
@georgeh0 Hey, Can this be assigned to me?
Would love to work on this.

badmonster0:
welcome to the community @Geoff-Robin , looking forward to work together! issue is yours :)

badmonster0:
thanks @Geoff-Robin for supporting this!

PR #1459, opened 2026-01-03
feat/specific_db_schema_for_internal_tables

Geoff-Robin:
- Explicit Table Qualification: Refactored SQL queries to use schema.table syntax to ensure robust isolation and avoid connection-level side effects like modifying the search_path.
- Automated Schema Lifecycle: Implemented automatic schema creation and ALTER TABLE ... SET SCHEMA migration logic to provide a seamless upgrade path for existing tables currently in the public schema.

Re-implementation of [this PR](https://github.com/cocoindex-io/cocoindex/pull/1444) and addresses [this issue](https://github.com/cocoindex-io/cocoindex/issues/1411)

### DIFF
--- a/rust/cocoindex/src/execution/db_tracking.rs
+++ b/rust/cocoindex/src/execution/db_tracking.rs
@@ -436,8 +436,9 @@ impl ListTrackedSourceKeyMetadataState {
         source_id: i32,
         db_setup: &'a TrackingTableSetupState,
         pool: &'a PgPool,
-    ) -> impl Stream<Item = std::result::Result<TrackedSourceKeyMetadata, sqlx::Error>> + 'a {
-        let table_name = qualify_table_name_with_schema(&db_setup.table_name).unwrap();
+    ) -> Result<impl Stream<Item = std::result::Result<TrackedSourceKeyMetadata, sqlx::Error>> + 'a>
+    {
+        let table_name = qualify_table_name_with_schema(&db_setup.table_name)?;
         self.query_str = format!(
             "SELECT \
             source_key, processed_source_ordinal, {}, process_logic_fingerprint, max_process_ordinal, process_ordinal \
@@ -449,7 +450,7 @@ impl ListTrackedSourceKeyMetadataState {
             },
             table_name
         );
-        sqlx::query_as(&self.query_str).bind(source_id).fetch(pool)
+        Ok(sqlx::query_as(&self.query_str).bind(source_id).fetch(pool))
     }
 }
 
--- a/rust/cocoindex/src/setup/db_metadata.rs
+++ b/rust/cocoindex/src/setup/db_metadata.rs
@@ -19,15 +19,15 @@ pub struct SetupMetadataRecord {
     pub staging_changes: sqlx::types::Json<Vec<StateChange<serde_json::Value>>>,
 }
 
-pub fn get_setup_metadata_table_name() -> String {
-    let schema = get_settings().unwrap().db_schema_name;
+pub fn get_setup_metadata_table_name() -> Result<String> {
+    let schema = get_settings()?.db_schema_name;
     if let Some(schema) = schema {
-        return format!(
+        Ok(format!(
             "{}.{SETUP_METADATA_TABLE_NAME_UNQUALIFIED}",
             utils::db::sanitize_identifier(&schema)
-        );
+        ))
     } else {
-        return SETUP_METADATA_TABLE_NAME_UNQUALIFIED.to_string();
+        Ok(SETUP_METADATA_TABLE_NAME_UNQUALIFIED.to_string())
     }
 }
 
@@ -41,7 +41,7 @@ pub fn parse_flow_version(state: &Option<serde_json::Value>) -> Option<u64> {
 /// Returns None if metadata table doesn't exist.
 pub async fn read_setup_metadata(pool: &PgPool) -> Result<Option<Vec<SetupMetadataRecord>>> {
     let mut db_conn = pool.acquire().await?;
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str =
         format!("SELECT flow_name, resource_type, key, state, staging_changes FROM {table_name}",);
     let metadata = sqlx::query_as(&query_str).fetch_all(&mut *db_conn).await;
@@ -84,7 +84,7 @@ async fn read_metadata_records_for_flow(
     flow_name: &str,
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
 ) -> Result<HashMap<ResourceTypeKey, SetupMetadataRecord>> {
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str = format!(
         "SELECT flow_name, resource_type, key, state, staging_changes FROM {table_name} WHERE flow_name = $1",
     );
@@ -112,7 +112,7 @@ async fn read_state(
     type_id: &ResourceTypeKey,
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
 ) -> Result<Option<serde_json::Value>> {
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str = format!(
         "SELECT state FROM {table_name} WHERE flow_name = $1 AND resource_type = $2 AND key = $3",
     );
@@ -132,7 +132,7 @@ async fn upsert_staging_changes(
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
     action: WriteAction,
 ) -> Result<()> {
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str = match action {
         WriteAction::Insert => format!(
             "INSERT INTO {table_name} (flow_name, resource_type, key, staging_changes) VALUES ($1, $2, $3, $4)",
@@ -158,7 +158,7 @@ async fn upsert_state(
     action: WriteAction,
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
 ) -> Result<()> {
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str = match action {
         WriteAction::Insert => format!(
             "INSERT INTO {table_name} (flow_name, resource_type, key, state, staging_changes) VALUES ($1, $2, $3, $4, $5)",
@@ -183,7 +183,7 @@ async fn delete_state(
     type_id: &ResourceTypeKey,
     db_executor: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
 ) -> Result<()> {
-    let table_name = get_setup_metadata_table_name();
+    let table_name = get_setup_metadata_table_name()?;
     let query_str = format!(
         "DELETE FROM {table_name} WHERE flow_name = $1 AND resource_type = $2 AND key = $3",
     );
@@ -351,9 +351,8 @@ impl MetadataTableSetup {
 impl ResourceSetupChange for MetadataTableSetup {
     fn describe_changes(&self) -> Vec<setup::ChangeDescription> {
         if self.metadata_table_missing {
-            let table_name = get_setup_metadata_table_name();
             vec![setup::ChangeDescription::Action(format!(
-                "Create the cocoindex metadata table {table_name}"
+                "Create the cocoindex metadata table {SETUP_METADATA_TABLE_NAME_UNQUALIFIED}"
             ))]
         } else {
             vec![]
@@ -376,7 +375,7 @@ impl MetadataTableSetup {
         }
         let lib_context = get_lib_context().await?;
         let pool = lib_context.require_builtin_db_pool()?;
-        let db_schema = get_settings().unwrap().db_schema_name;
+        let db_schema = get_settings()?.db_schema_name;
 
         if let Some(schema) = &db_schema {
             let query_str = format!(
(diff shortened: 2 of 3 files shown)


georgeh0:
Thanks for giving another try!

I think we don't need to logic to migrate schema for the tracking table, because it cannot be detected in the current way: Our metadata table and tracking table in our internal database should be always within the same schema.
- If we use the metadata table to keep the schema name, it's always the same as its own schema. It's essentially repeating it's own schema name.
- If later users set a different schema name, we won't be able to access their metadata table under the previous schema any way. So we won't be able to automatically migrate db schema for users.

So I think
- We won't support migrating schema for the tracking table automatically.
- We also don't need to keep the schema name in the metadata table.

In this sense the db schema name is similar to the settings in db connection: they're *addresses* to cocoindex internal data, not the data itself. So we can always use the current settings, and don't deal with the previous settings.

Any thoughts?

Geoff-Robin:
Yeah no point in migrating, when we lost the 'pointer' to the old schema. I'll be removing that piece of logic then.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -254,26 +268,35 @@ impl ResourceSetupChange for TrackingTableSetupChange {
...
         let pool = lib_context.require_builtin_db_pool()?;
         if let Some(desired) = &self.desired_state {
-            for lagacy_name in self.legacy_tracking_table_names.iter() {
-                let query = format!(
-                    "ALTER TABLE IF EXISTS {} RENAME TO {}",
-                    lagacy_name, desired.table_name
-                );
-                sqlx::query(&query).execute(pool).await?;
-            }
-
```
georgeh0, line 293: Any reason why we remove this logic?

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -254,26 +268,35 @@ impl ResourceSetupChange for TrackingTableSetupChange {
...
+
+        // Remove legacy tracking tables not matching desired state
+        for legacy_name in self.legacy_tracking_table_names.iter() {
+            // If desired state exists, we only drop if it's strictly a legacy name different from current
+            let keep = self
+                .desired_state
+                .as_ref()
+                .map(|s| &s.table_name == legacy_name)
+                .unwrap_or(false);
+            if !keep {
```
georgeh0: The check here is duplicated. 

Let's keep the code as close as before. Only add code related to the new feature. It's easier to verify the correctness.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -282,14 +305,18 @@ impl TrackingTableSetupChange {
...
             .as_ref()
             .and_then(|v| v.source_state_table_name.as_ref());
+
         if let Some(source_state_table_name) = source_state_table_name {
-            for lagacy_name in self.legacy_source_state_table_names.iter() {
-                let query = format!(
-                    "ALTER TABLE IF EXISTS {lagacy_name} RENAME TO {source_state_table_name}"
-                );
-                sqlx::query(&query).execute(pool).await?;
-            }
```
georgeh0, line 321: Same here. Why do we remove it?

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -304,9 +331,15 @@ impl TrackingTableSetupChange {
...
-                let query = format!("DROP TABLE IF EXISTS {lagacy_name}");
+        }
+
+        // Remove legacy source state tables
+        for legacy_name in self.legacy_source_state_table_names.iter() {
+            let keep = source_state_table_name
+                .map(|n| n == legacy_name)
+                .unwrap_or(false);
+            if !keep {
+                let query = format!("DROP TABLE IF EXISTS {legacy_name}");
```
georgeh0: Same here. The new check here is not needed.

rust/cocoindex/src/execution/db_tracking_setup.rs
Geoff-Robin: @georgeh0 The original RENAME TO logic was removed because it was designed for single-schema environments and would fail if we attempted to rename tables across different schemas (which we now support).

To solve the data loss issue for flow renames, I'll be introducing logic that uses qualified legacy table names. This will safely perform the rename if the legacy table already exists within the current configured schema. If the user has changed schemas, those old tables will be ignored for now to avoid the complexity and risks of cross-schema data migration.

I'll add the changes for this, once you give a go.

georgeh0: > would fail if we attempted to rename tables across different schemas

Since metadata table and tracking/state tables are always within the same schema, once schema changed, we won't be able to see legacy table names any way, so won't hit this (in this case, it's user's responsibility to rename the entire schema; for us it's similar to a database address change)

> To solve the data loss issue for flow renames, I'll be introducing logic that uses qualified legacy table names. This will safely perform the rename if the legacy table already exists within the current configured schema.

Sounds good. This will be sufficient. Thanks!

Geoff-Robin: I've made the change now, you can review them.

rust/cocoindex/src/builder/exec_ctx.rs
```diff
@@ -320,7 +321,9 @@ pub fn build_flow_setup_execution_context(
                         .as_ref()
                         .map(|v| v.table_name.clone())
                 })
-                .unwrap_or_else(|| db_tracking_setup::default_tracking_table_name(&flow_inst.name)),
+                .unwrap_or_else(|| {
+                    db_tracking_setup::default_tracking_table_name(db_schema_name, &flow_inst.name)
+                }),
```
georgeh0: Actually I think we don't need to put schema name here. The value of this field will go to the metadata table, but we don't need to, because we expect the tracking table and metadata table always reside in the same db schema. 

A simpler and cleaner way is to only add the schema name at the last stop, i.e. when building the SQL statements. We should never persist schema name, just like we don't persist db connection string.

This will create the following benefits:

- When users want to migrate the db schema of their tables, they can simply rename the schema name, or move all their cocoindex tables to the target schema, and things will just work.
- We don't need reverse parsing schema name out of qualified table name (like `split_qualified_name()` below).
- Less change to the existing code base, so easier to verify  things not broken.

Thanks!

Geoff-Robin: @georgeh0 I agree propagating the schema through the call stack seems to be a roundabout way, i'll fetch the db_schema_name directly.

georgeh0:
Sorry for the late review.

The latest change still has broader scope than necessary. e.g. both `default_tracking_table_name()` and `default_source_state_table_name()` qualifies the table name. It'll unnecessarily adding qualified table names to `FlowSetupState`.

Can we strictly qualify the table name with schema name only before creating SQL statements? This will minimize the impact to the finial SQL statements and also sufficient. Thanks!

Geoff-Robin:
@georgeh0 Made the change just now, Lmk if there's anything else.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -1,9 +1,36 @@
...
+fn split_qualified_name(name: &str) -> (Option<&str>, &str) {
+    if let Some((schema, table)) = name.split_once('.') {
+        (Some(schema), table)
+    } else {
+        (None, name)
+    }
+}
+
+pub fn qualify_table_name_with_schema(table: &str) -> String {
+    let (schema_in_name, table_part) = split_qualified_name(table);
```
georgeh0: I think we don't need to call `split_qualified_name()`. The input `table` name should always be an unqualified table name, right?

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -278,28 +307,54 @@ impl ResourceSetupChange for TrackingTableSetupChange {
...
-            for lagacy_name in self.legacy_tracking_table_names.iter() {
-                let query = format!(
-                    "ALTER TABLE IF EXISTS {} RENAME TO {}",
-                    lagacy_name, desired.table_name
-                );
-                sqlx::query(&query).execute(pool).await?;
+            // Attempt to rename legacy tables within the same schema
+            for legacy_name in self.legacy_tracking_table_names.iter() {
+                let (legacy_schema, _) = split_qualified_name(legacy_name);
+                let (_, desired_table) = split_qualified_name(&desired.table_name);
```
georgeh0: Since we only qualify it before we build SQL and never store it, both legacy name and desired name should never be qualified. We don't need it. All change we need is to use qualify the name when constructing the `ALTER TABLE` statement below.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -278,28 +307,54 @@ impl ResourceSetupChange for TrackingTableSetupChange {
...
-        } else {
-            for lagacy_name in self.legacy_tracking_table_names.iter() {
-                let query = format!("DROP TABLE IF EXISTS {lagacy_name}");
-                sqlx::query(&query).execute(pool).await?;
-            }
+        }
+
+        // Remove legacy tracking tables
+        for legacy_name in self.legacy_tracking_table_names.iter() {
+            let query = format!("DROP TABLE IF EXISTS {legacy_name}");
```
georgeh0: `legacy_name` should be qualified when we construct this SQL statement.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -312,14 +367,22 @@ impl TrackingTableSetupChange {
...
+                    let (_, desired_table) = split_qualified_name(source_state_table_name);
+                    if legacy_schema == desired_schema && table_exists(pool, legacy_name).await? {
+                        let query = format!(
+                            "ALTER TABLE {legacy_name} RENAME TO {}",
+                            utils::db::sanitize_identifier(desired_table)
+                        );
+                        sqlx::query(&query).execute(pool).await?;
+                        break;
+                    }
+                }
```
georgeh0: This shouldn't be needed any more.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -336,11 +399,12 @@ impl TrackingTableSetupChange {
...
-        } else {
-            for lagacy_name in self.legacy_source_state_table_names.iter() {
-                let query = format!("DROP TABLE IF EXISTS {lagacy_name}");
-                sqlx::query(&query).execute(pool).await?;
-            }
+        }
+
+        // Remove legacy source state tables
+        for legacy_name in self.legacy_source_state_table_names.iter() {
+            let query = format!("DROP TABLE IF EXISTS {legacy_name}");
```
georgeh0: `legacy_name` should be qualified when we construct this SQL statement.

Geoff-Robin:
@georgeh0 I think all uneccessary changes should have been tracked backed. I've qualified all table names before sql queries.
Lmk if there's anything wrong.

rust/cocoindex/src/setup/db_metadata.rs
```diff
@@ -320,11 +337,17 @@ pub struct MetadataTableSetup {
 
 impl MetadataTableSetup {
     pub fn into_setup_info(self) -> ResourceSetupInfo<(), (), MetadataTableSetup> {
+        let db_schema = get_settings().unwrap().db_schema_name;
+        let description = if let Some(schema) = &db_schema {
+            format!("CocoIndex Metadata Table ({schema})")
+        } else {
+            "CocoIndex Metadata Table (public)".to_string()
+        };
```
georgeh0: Note that when there's no db_schema specified, the db_schema doesn't have to be `public` (there can be a default one in the connection).

I think we don't need this change. Just use the old description without the schema name.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -1,9 +1,25 @@
...
 
 use crate::execution::db_tracking;
+use crate::lib_context::get_settings;
 use crate::setup::{CombinedState, ResourceSetupChange, ResourceSetupInfo, SetupChangeType};
 use serde::{Deserialize, Serialize};
 use sqlx::PgPool;
+use std::collections::{BTreeMap, BTreeSet};
+
+pub fn qualify_table_name_with_schema(table: &str) -> String {
+    let schema: Option<String> = get_settings().unwrap().db_schema_name;
```
georgeh0: Let's avoid `unwrap()` here. Use `?`. Please change the return type to `Result<String>`, and also use `?` in callers.

georgeh0:
> @georgeh0 I think all uneccessary changes should have been tracked backed. I've qualified all table names before sql queries. Lmk if there's anything wrong.

Thanks! Most part looks good. Two last comments.

rust/cocoindex/src/execution/db_tracking.rs
```diff
@@ -427,6 +437,7 @@ impl ListTrackedSourceKeyMetadataState {
         db_setup: &'a TrackingTableSetupState,
         pool: &'a PgPool,
     ) -> impl Stream<Item = std::result::Result<TrackedSourceKeyMetadata, sqlx::Error>> + 'a {
+        let table_name = qualify_table_name_with_schema(&db_setup.table_name).unwrap();
```
georgeh0: Please use `?` here too. Change the return type of the function please.

rust/cocoindex/src/setup/db_metadata.rs
```diff
@@ -18,6 +19,18 @@ pub struct SetupMetadataRecord {
     pub staging_changes: sqlx::types::Json<Vec<StateChange<serde_json::Value>>>,
 }
 
+pub fn get_setup_metadata_table_name() -> String {
+    let schema = get_settings().unwrap().db_schema_name;
```
georgeh0: Please use `?` here too. Change the return type of the function please.

rust/cocoindex/src/setup/db_metadata.rs
```diff
@@ -358,8 +376,19 @@ impl MetadataTableSetup {
         }
         let lib_context = get_lib_context().await?;
         let pool = lib_context.require_builtin_db_pool()?;
+        let db_schema = get_settings().unwrap().db_schema_name;
```
georgeh0: Please use `?` here too.

Geoff-Robin:
@georgeh0 I think error propagation has been added everywhere now.  If there are any changes let me know

georgeh0 (approved):
Looks good! Thanks for making it happen!

Issue #1429, opened 2025-12-27
[FEATURE] Optimize for LanceDB commit


georgeh0:
## Backgrounds

When many small appends are performed, Lance tables require [compaction](https://lance.org/guide/read_and_write/?h=compaction#compact-data-files) and index optimization at regular intervals (say every several thousand rows) to ensure that the table fragments are merged so they aren't too small. This reduces query latency over time.

Additionally, in LanceDB OSS, index updates aren't automatic - when new data is added in small batches (as we do in incremental processing with CocoIndex), we need to call table.optimize() periodically to ensure that all indexes are updated with the new data. In the interim (when new data is coming in, but index isn't updated), LanceDB will still serve results from the index + brute force retrieval (for both vector/FTS indexes), but as the volume of data that's not in the index increases, this can also impact query latency.

## Proposed solution

We can improve our LanceDB target. Do two things:

1. We have [batching](https://cocoindex.io/blogs/batching) in cocoindex but it's only enabled for cocoindex functions now. We can further integrate with the lancedb targets, so we can commit to lancedb in batch. This will combine smaller commits.
2. Within the lancedb target, we can add a counter to call `optimize()` after a certain number of batched commits. We can set a reasonable default value for the threshold (e.g. 100) and make it configurable.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

prrao87:
I think this can be closed @georgeh0

PR #1466, opened 2026-01-09
feat: call `optimize()` for lancedb tables

georgeh0:
#1429


badmonster0:
thanks a lot for getting it out quickly!! cc @prrao87

PR #1468, opened 2026-01-09
feat: clean up tracking metadata and target data for stale sources

### LINKED ISSUE #22: [FEATURE] Clean up tracking metadata when flow definition changed
**What is the use case?**
When the flow is updated, currently we don't clean up stale tracking metadata in the database yet.

**Additional context**

thisisharsh7:
This PR adds automatic cleanup during flow setup. When CocoIndex detects that a source has been renamed or removed, it now:

- Streams through tracking metadata for the stale source
 - Uses the target factory to properly delete rows from all target types (Postgres, DuckDB, etc.)
- Removes tracking entries and source state records
- Processes cleanups with limited parallelism to avoid overwhelming the database

fixes #22

### DIFF
--- a/rust/cocoindex/src/execution/db_tracking_setup.rs
+++ b/rust/cocoindex/src/execution/db_tracking_setup.rs
@@ -480,10 +480,12 @@ impl TrackingTableSetupChange {
             // Parse tracked keys using pre-computed field_schemas
             let mut parsed_keys = Vec::new();
             for tracked_key_info in tracked_keys {
-                match Self::parse_key_from_json(
-                    &tracked_key_info.key,
+                match value::KeyValue::from_json(
+                    tracked_key_info.key.clone(),
                     &target_info.key_field_schemas,
-                ) {
+                )
+                .with_context(|| format!("Failed to parse key: {}", tracked_key_info.key))
+                {
                     Ok(key) => parsed_keys.push(key),
                     Err(e) => {
                         tracing::warn!(
@@ -531,14 +533,5 @@ impl TrackingTableSetupChange {
         }
 
         Ok(total_deleted)
-    }
-
-    /// Parse a key from JSON using pre-computed field schemas
-    fn parse_key_from_json(
-        key_json: &serde_json::Value,
-        key_field_schemas: &[schema::FieldSchema],
-    ) -> Result<value::KeyValue> {
-        value::KeyValue::from_json(key_json.clone(), key_field_schemas)
-            .with_context(|| format!("Failed to parse key: {}", key_json))
     }
 }


badmonster0:
thanks a lot @thisisharsh7 we will get on it shortly!

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -312,4 +334,220 @@ impl TrackingTableSetupChange {
...
+        );
+
+        Ok(())
+    }
+
+    /// Delete target data for Postgres targets using direct SQL
+    /// This implements the "for each tracked source row, do a deletion" requirement
+    ///
+    /// Note: Currently only supports targets using the builtin database.
+    /// Targets with custom databases will require manual cleanup.
```
georgeh0: Treating builtin postgres db target as special thing is unexpected. We should use the target factory to do the cleanup.

There're two cases:
- The specific target is no longer there (e.g. it's removed, or the key schema changed so cocoindex assigns a new target ID for it). We don't need to do anything special, given the target setup cleanup logic will drop the target.
- The target is still there, i.e. the same target ID still exists in the current running flow. In this case, we have the specific target factory and export context, so we can call it to delete rows from it.

rust/cocoindex/src/execution/db_tracking.rs
```diff
@@ -312,6 +312,64 @@ pub async fn delete_source_tracking_info(
     Ok(())
 }
 
+/// Tracking data for cleanup - includes target keys for each source row
+#[derive(Debug)]
+#[allow(dead_code)]
```
georgeh0: `#[allow(dead_code)]` is not needed here as long as the code is in use?

thisisharsh7: Yes it needs to be remove, thanks for flagging it I will remove it

rust/cocoindex/src/execution/db_tracking_setup.rs
thisisharsh7: got it thanks for the review.  few questions I have 

1. When a flow is being completely dropped (`desired_state = None`), should the stale source cleanup logic run at all?

Currently, my implementation has a fallback that includes all targets when `desired_state` is None. But based on your feedback about Case 1 vs Case 2, I'm thinking that during a flow drop:
  - All targets are Case 1 (being dropped anyway)
  - Target setup cleanup will handle dropping them
  - So stale source cleanup should return early and do nothing

 Is this understanding correct? Should I change it to skip cleanup entirely when `desired_state` is `None`?
  
2. If some tracked keys parse successfully but others fail (due to schema mismatch or corruption), should I:

  - Delete only the keys that parsed successfully (partial cleanup)
  - Fail the entire operation if any key fails to parse
  - Log warnings and continue, accepting orphaned data

My current implementation does first one with warning. What do you think?

3. If a `target_id` exists in the tracking entries but is not in the existing_targets map (maybe it was already dropped or there's a race condition), should I:
  - Silently skip it
  - Log an error but continue
  - Fail the entire cleanup operation

  What would be safest here?

georgeh0: Hi @thisisharsh7, thanks for the detailed thoughts!

1. This is correct!
2. Ideally this shouldn't happen, because when the target key schema change, we always assign a new `target_id` and the original target will be dropped (i.e. fall into case 1). In case it happens, deleting only the keys that parsed successfully + log warnings make sense.
3. In this case, skip and logging warning also look good.

thisisharsh7: got it thanks!

thisisharsh7:
kindly review it

rust/cocoindex/src/setup/driver.rs
```diff
@@ -917,6 +945,47 @@ pub(crate) async fn apply_changes_for_flow_ctx(
     db_pool: &PgPool,
     write: &mut (dyn std::io::Write + Send),
 ) -> Result<()> {
+    // Attach export contexts to tracking table setup change BEFORE getting the setup_change
+    // This is needed for stale source cleanup which uses target factories
+    if let FlowSetupChangeAction::Setup = action {
+        if let Some(tracking_table) = &mut flow_exec_ctx.setup_change.tracking_table {
+            if let Some(setup_change) = &mut tracking_table.setup_change {
```
georgeh0: nit: with the new let-chain syntax can, the 3 layers can merge into one:

https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -312,4 +338,235 @@ impl TrackingTableSetupChange {
...
+
+        tracing::info!(
+            "Cleaning up tracking metadata and target data for {} stale source(s): {:?}",
+            source_ids.len(),
+            source_ids
+        );
+
+        // Step 1: Read all tracking entries for stale sources
+        let tracking_entries =
+            db_tracking::read_tracking_entries_for_sources(&source_ids, desired, pool).await?;
```
georgeh0: This can be huge (note that the number of target entries can be much larger than source entries, considering the fanouts).

I think we want to read it in streaming way, and do cleanup for each row one by one rather than in a huge transaction (similar to how we process each row: there can be limited number of rows being processed in parallel).

georgeh0:
Please also merge with latest `main`. The sccache issue that makes workflow should be gone.

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -312,4 +334,214 @@ impl TrackingTableSetupChange {
...
+
+        // Stream tracking entries instead of loading all at once
+        let entries_stream = db_tracking::read_tracking_entries_for_sources_stream(
+            source_ids.clone(),
+            desired.clone(),
+            pool.clone(),
+        );
+
+        // Process with limited parallelism
+        const MAX_CONCURRENT_ROW_CLEANUPS: usize = 10;
```
georgeh0: Let's use [`GlobalExecutionOptions::source_max_inflight_rows`](https://github.com/cocoindex-io/cocoindex/blob/6cae8fe65b00c4f77f279e7c128faff9a6a02077/rust/cocoindex/src/settings.rs#L14) for this.

You can get the lib context by [this](https://github.com/cocoindex-io/cocoindex/blob/6cae8fe65b00c4f77f279e7c128faff9a6a02077/rust/cocoindex/src/lib_context.rs#L365).

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -312,4 +334,214 @@ impl TrackingTableSetupChange {
...
+            .map(|(i, value_type)| schema::FieldSchema {
+                name: format!("_key_{}", i),
+                value_type: schema::EnrichedValueType {
+                    typ: value_type.clone(),
+                    nullable: false,
+                    attrs: Arc::new(BTreeMap::new()),
+                },
+                description: None,
+            })
+            .collect();
```
georgeh0: We're constructing this vector repeatedly for all rows.

If a `schema::FieldSchema` is all we need, we can consider create it upfront before putting it into `TrackingTableSetupChange::desired_target`

georgeh0:
hi @thisisharsh7 can you also briefly mention how is this tested? Thanks!

thisisharsh7:
> hi @thisisharsh7 can you also briefly mention how is this tested? Thanks!

hi @georgeh0, I tested this by creating two test scripts:
- test_v1.py: Creates a flow with a LocalFile source named "docs"
- test_v2.py: Same flow but with the source renamed to "documents"

Test steps:
1. Reset database and run v1 to create initial state
2. Run v2 which triggers the cleanup

output -:
```
2026-01-24T10:14:39.541887Z  INFO setup.apply_changes_for_flow_ctx{flow_name=TestFlow}:setup.apply_changes_for_flow{flow_name=TestFlow}: cocoindex_engine::execution::db_tracking_setup: Cleaning up tracking metadata and target data for 1 stale source(s): [3]
2026-01-24T10:14:39.561177Z  INFO setup.apply_changes_for_flow_ctx{flow_name=TestFlow}:setup.apply_changes_for_flow{flow_name=TestFlow}: cocoindex_engine::execution::db_tracking_setup: Processed 2 tracking entries, deleted 2 target rows
2026-01-24T10:14:39.563644Z  INFO setup.apply_changes_for_flow_ctx{flow_name=TestFlow}:setup.apply_changes_for_flow{flow_name=TestFlow}: cocoindex_engine::execution::db_tracking_setup: Deleted 2 tracking entries
```
Also I discovered that the cleanup code was nested inside a conditional block that only ran when source state tables existed. This caused cleanup to skip for sources like LocalFile. Fixed by moving the cleanup outside the conditional.

let me know if you'd like me to test anything else!!

rust/cocoindex/src/execution/db_tracking_setup.rs
```diff
@@ -522,26 +533,12 @@ impl TrackingTableSetupChange {
...
-                    attrs: Arc::new(BTreeMap::new()),
-                },
-                description: None,
-            })
-            .collect();
-
-        value::KeyValue::from_json(key_json.clone(), &field_schemas)
+        value::KeyValue::from_json(key_json.clone(), key_field_schemas)
             .with_context(|| format!("Failed to parse key: {}", key_json))
     }
```
georgeh0: nit: it doesn't need to be a function now

georgeh0:
Thanks a lot for making this happen! Merged.

PR #1470, opened 2026-01-10
`docs`: add env file sourcing tip to `meeting_notes_graph` example

Haleshot:
Added an alternative (better?) approach for setting environment variables in the meeting notes graph example.

Simply export .env variables with `set -a && source .env && set +a` instead of manually exporting each variable (might get tedious for non-CLI folks).

### DIFF
--- a/examples/meeting_notes_graph/README.md
+++ b/examples/meeting_notes_graph/README.md
@@ -50,6 +50,12 @@ export GOOGLE_SERVICE_ACCOUNT_CREDENTIAL=/absolute/path/to/service_account.json
 export GOOGLE_DRIVE_ROOT_FOLDER_IDS=folderId1,folderId2
 ```
 
+Alternatively, fill in your values in `.env.example` and source it:
+
+```sh
+set -a && source .env.example && set +a
+```
+
 Notes:
 
 - `GOOGLE_DRIVE_ROOT_FOLDER_IDS` accepts a comma-separated list of folder IDs


badmonster0:
oh actually we ignored env file to avoid user accident check in keys. 

@cocoindex-dev thoughts?

Haleshot:
For this example you mean? I found the [`.env.example`](https://github.com/cocoindex-io/cocoindex/blob/main/examples/meeting_notes_graph/.env.example) for the meetings_notes example and searches of [`.env`](https://github.com/search?q=repo%3Acocoindex-io%2Fcocoindex+.env+language%3ADotenv&type=code&l=Dotenv) across the codebase.

badmonster0:
great question !! ;) i can see this inconsistency brings in confusion

so the heuristics we take is if it contains XYZ_API_KEY then we only provide a .env.example and we expect users copy it to .env

if it doesnt contain XYZ_API_KEY we use .env

i think this heuristics is not that consistent across multiple examples. 

maybe time to think about a consistent practice here to avoid confusion going on / and would benefit v1 rollout, coming soon!

Haleshot:
> so the heuristics we take is if it contains XYZ_API_KEY then we only provide a .env.example and we expect users copy it to .env

My first draft of the note (added in this PR) was to tell folks to move it to `.env` first from the `.env.example` and then source it from the `.env`. 

_Re:_ the heuristics and potential ambiguity owing to inconsistencies b/w both, I see comments in some [`.env.example`](https://github.com/cocoindex-io/cocoindex/blob/0d360cb36e267043c38192c0e74dca108aa0f57b/examples/custom_source_hn/.env.example#L2) files where you've documented the content to be copied to an `.env` file.

Would love to know your and @georgeh0's thoughts on how I can go about documenting this (similar comments across such `.env.example` files or...

Haleshot:
Pinging on this thread again (personal reminder to wire something up to clear this ambiguity).

badmonster0:
make sense, thanks a lot i totally forget about it, would you like to create a new issue - standardize .env in all examples or something like that?

Haleshot:
That makes more sense (than to continue on a merged PR thread), will do that, thanks.

badmonster0:
Thank you!

PR #1482, opened 2026-01-14
workflow: reuse precommit workflow for CI

georgeh0:


Haleshot:
Would love to know your thoughts on using [`prek | Better pre-commit, re-engineered in Rust`](https://github.com/j178/prek). Can create a detailed issue later on based on your thoughts (on whether this is something worth looking into). There also exists a [GitHub action](https://github.com/j178/prek-action) for the same.
The readme states projects such as FastAPI, Apache airflow using it :)

My thoughts on proposing/stating this was with the intentions of using something built in Rust! 🦀

georgeh0:
> Would love to know your thoughts on using [`prek | Better pre-commit, re-engineered in Rust`](https://github.com/j178/prek). Can create a detailed issue later on based on your thoughts (on whether this is something worth looking into). There also exists a [GitHub action](https://github.com/j178/prek-action) for the same. The readme states projects such as FastAPI, Apache airflow using it :)
> 
> My thoughts on proposing/stating this was with the intentions of using something built in Rust! 🦀

@Haleshot Thanks for proposing this! Without evaluating in detail yet, if it's stable enough and supports all features we already used for the current pre-commit, we can switch.  Thanks!

PR #1487, opened 2026-01-15
Add Ladybug (successor to Kuzu) as a target

prrao87:
## Motivation and context

Kuzu was an open-source (MIT) licensed property graph database, but the project was archived in October 2025. This PR introduces [Ladybug](https://github.com/LadybugDB/ladybug), which is a fork of Kuzu (and an identical API) as a target for users who may have wanted to use a locally running embedded graph database built for performance.

Ladybug is maintained by @adsharma and there is a lot more functionality planned for greater interoperability with the lakehouse ecosystem, so CocoIndex users can also benefit.

## Breaking change? 

No.

## Description of the changes

This PR adds a new Ladybug target that is a drop‑in Kuzu replacement by reusing the [same Cypher API server](https://github.com/LadybugDB/api-server) integration, plus Python spec/declaration definitions for users to access in CocoIndex.

### High-level summary of changes
* Wires Ladybug into the expected target registration for CocoIndex
* Adds Ladybug target docs (new page + index/sidebar)
* Adds a small serialization test to ensure the new specs emit the expected engine shape. A regression in serialization (missing kind, wrong mapping fields, missing connection key) would surface only at runtime. The new tests added are a minimal guard that the new specs serialize into the exact shape the engine expects.
* Adds an opt‑in `COCOINDEX_SKIP_UV=1` env variable switch in `run_cargo_test.sh` to keep Python env vars intact (if the local uv-managed Python doesn’t align with the version expected by the tests) when running cargo test.

### Detailed list
* Added Ladybug Rust target (Identical to the Kuzu API server-based solution that existed previously) in `ladybug.rs`.
* Registered Ladybug target module in `mod.rs`.
* Registered Ladybug factory in `registration.rs`.
    * `reqwest_client` is moved into `targets::kuzu::register`, so we clone it to reuse the same client for Ladybug registration.
    * We should considering de-registering Kuzu and letting users know to use Ladybug
* Added Python target/declaration specs in `_engine_builtin_specs.py`.
* Added Ladybug doc page in `ladybug.md` (includes pip install real_ladybug note).
* Added Ladybug to targets index in `index.md`.
* Added Ladybug to docs sidebar in `sidebars.ts`.
* Added a note in Kuzu docs pointing to Ladybug in `kuzu.md`.
    * Ideally, we should remove these docs the codebase as Kuzu is no longer maintained - Ladybug is its true successor.
* Added Ladybug spec serialization tests in `test_targets_specs.py`.
* Added opt‑in skip of `uv run` for Rust tests via `COCOINDEX_SKIP_UV=1` in `run_cargo_test.sh`.

## To do
* Decide on a strategy to remove Kuzu from the codebase and communicate to the user base that Ladybug is the “new Kuzu” going forward.
* Recommendation: Announce to users on discord that Ladybug is like-for-like with Kuzu, and suggest removing Kuzu entirely from this code base.


badmonster0:
thanks @prrao87 !! will get back on this shortly!

rust/cocoindex/src/ops/targets/ladybug.rs
georgeh0, line 1: The implementation seems exactly the same as Kuzu's. We only need to keep one of them.

One way is to keep this new `ladybug.rs`, but add a new field `target_name: str` to `Factory`.  We can pass in the `name` to the `register()` function with `"Ladybug"` or `"Kuzu"` separately. Then we get the same code working for both targets.

What do you think?

docs/docs/targets/ladybug.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Ladybug
+description: CocoIndex Ladybug Target
+toc_max_heading_level: 4
+---
+import { ExampleButton } from '../../src/components/GitHubButton';
+
+# Ladybug
+
+Exports data to a [Ladybug](https://github.com/LadybugDB/ladybug) graph database. Ladybug is a maintained fork of Kuzu that carries forth the original vision of Kuzu, with added functionality for the lakehouse ecosystem. Just like Kuzu, Ladybug follows the structurd property graph model and functions as a fast, embedded database with a permissive (MIT) license.
```
adsharma: structured

rust/cocoindex/src/ops/targets/ladybug.rs
adsharma: Given its read-only status, in the next few months it would be dangerous to run Kuzu as security problems are discovered in its dependencies. I expect the usage to shift to one of its forks such as ladybug.

prrao87: I agree, this is why i strongly recommend removing Kuzu from the list and only using Ladybug in place of it (as Ladybug is identical to Kuzu but is maintained and being improved).

georgeh0: Makes sense! I think we can cleanup Kuzu (but leave its doc for now, with a single line to point to Ladybug; and also leave type aliases in code (e.g. `KuzuConnection = LadybugConnection`). Do you want to do this cleanup in this PR? (the history will look nicer as it'll clearly show ladybug modules coming from kuzu)

prrao87: @georgeh0 just so I'm clear, are you suggesting to *update* the Kuzu file and remove the new file `ladybug.rs`, so that the git history shows that Kuzu was updated to Ladybug? I can do that, just want to be sure I understood.

georgeh0: Hi @prrao87, I'm suggesting:

- Keep all the new Ladybug files, and remove all Rust code for kuzu (this will make Git detect it as a "file move", which leaves the Git history clean)
- In `targets/_engine_builtin_specs.py‎`, replace current Kuzu types with type aliases to Ladybug types
- For the doc `kuzu.md`, leave a single line to say it's replaced by Ladybug and point to there

(there's also Kuzu examples, should be migrated to ladybug too, but with the alias they won't be broken, so we can migrate it later)

badmonster0:
friendly follow up on this PR, @prrao87 pls keep us posted when it is ready for another look!

prrao87:
Hi @badmonster0, I'll test this out locally and make the fixes as requested soon, thanks!

badmonster0:
thanks a lot for getting back to us! please take your time :)

prrao87:
Hi @georgeh0 and @badmonster0, this follows up on the requested cleanup to finish the Kuzu → Ladybug transition.

- Kuzu support has been removed from Rust, while keeping Ladybug as the maintained successor, and the history should hopefully now reflect Ladybug coming from the former Kuzu module (clean move/rename).
- Kept Python `Kuzu*` names as aliases to `Ladybug` to ease migration pains. `KuzuConnection`, `Kuzu`, and `KuzuDeclaration` are now aliases to the Ladybug types in Python.
- The Kuzu docs page is reduced to a single-line pointer directing users to Ladybug.

I also checked the following by running tests locally:
```bash
uv run pytest -q python/cocoindex/tests/test_targets_specs.py
cargo check --manifest-path rust/cocoindex/Cargo.toml
```

They check out fine. Am running some tests on a toy dataset locally!

prrao87:
There are some blocking issues with ladybug's api-server, need to resolve those before this can be used. 😅 
Will keep you posted.

adsharma:
Hopefully resolved now. Let me know if you're still seeing problems @prrao87

prrao87:
Hi @georgeh0, I just tested with the latest ladybug and the latest api-server via the below code example in v0 of CocoIndex. It works great! 🎉

Once it's merged, I'd love to take this for a spin and have @badmonster0 and @adsharma chat on how we can make this integration tighter with the v1 API, cheers!

### Raw inputs

The ladybug graph inputs come from these two CSV files:

#### Nodes
```csv
name,age
Alice,30
Bob,25
```

#### Relationships
```csv
src,dst
Alice,Bob
```

### CocoIndex Flow (v0)
```py
import csv
import datetime
import io
import os
from pathlib import Path

import cocoindex

PERSON_CSV_PATH = Path("/Users/prrao/code/test_lbug/person.csv")
KNOWS_CSV_PATH = Path("/Users/prrao/code/test_lbug/knows.csv")
LADYBUG_API_SERVER_URL = "http://localhost:8000"
REFRESH_INTERVAL_SECONDS = 1


ladybug_connection = cocoindex.add_auth_entry(
    "ladybug_connection",
    cocoindex.targets.LadybugConnection(api_server_url=LADYBUG_API_SERVER_URL),
)


@dataclass
class PersonRow:
    name: str
    age: int


@dataclass
class KnowsRow:
    src: str
    dst: str


@cocoindex.op.function()
def parse_person_csv(content: str) -> list[PersonRow]:
    rows: list[PersonRow] = []
    reader = csv.DictReader(io.StringIO(content))
    for row in reader:
        name = (row.get("name") or "").strip()
        age_raw = (row.get("age") or "").strip()
        if not name:
            continue
        if not age_raw:
            raise ValueError(f"Missing age for person row: {row}")
        rows.append(PersonRow(name=name, age=int(age_raw)))
    return rows


@cocoindex.op.function()
def parse_knows_csv(content: str) -> list[KnowsRow]:
    rows: list[KnowsRow] = []
    reader = csv.DictReader(io.StringIO(content))
    for row in reader:
        src = (row.get("src") or "").strip()
        dst = (row.get("dst") or "").strip()
        if not src or not dst:
            continue
        rows.append(KnowsRow(src=src, dst=dst))
    return rows


@cocoindex.op.function()
def make_edge_id(src: str, dst: str) -> str:
    return f"{src}->{dst}"


@cocoindex.flow_def(name="LadybugCsvSync")
def ladybug_csv_sync_flow(
    flow_builder: cocoindex.FlowBuilder, data_scope: cocoindex.DataScope
) -> None:
    csv_dir = str(PERSON_CSV_PATH.parent)
    refresh_interval = datetime.timedelta(seconds=REFRESH_INTERVAL_SECONDS)

    data_scope["person_csv"] = flow_builder.add_source(
        cocoindex.sources.LocalFile(path=csv_dir, included_patterns=[PERSON_CSV_PATH.name]),
        refresh_interval=refresh_interval,
    )
    data_scope["knows_csv"] = flow_builder.add_source(
        cocoindex.sources.LocalFile(path=csv_dir, included_patterns=[KNOWS_CSV_PATH.name]),
        refresh_interval=refresh_interval,
    )

    person_nodes = data_scope.add_collector()
    knows_rels = data_scope.add_collector()

    with data_scope["person_csv"].row() as person_file:
        person_file["rows"] = person_file["content"].transform(parse_person_csv)
        with person_file["rows"].row() as person:
            person_nodes.collect(
                name=person["name"],
                age=person["age"],
            )

    with data_scope["knows_csv"].row() as knows_file:
        knows_file["rows"] = knows_file["content"].transform(parse_knows_csv)
        with knows_file["rows"].row() as knows:
            knows["edge_id"] = flow_builder.transform(
                make_edge_id, src=knows["src"], dst=knows["dst"]
            )
            knows_rels.collect(
                edge_id=knows["edge_id"],
                src=knows["src"],
                dst=knows["dst"],
            )

    person_nodes.export(
        "person_nodes",
        cocoindex.targets.Ladybug(
            connection=ladybug_connection,
            mapping=cocoindex.targets.Nodes(label="Person"),
        ),
        primary_key_fields=["name"],
    )

    knows_rels.export(
        "knows_relationships",
        cocoindex.targets.Ladybug(
            connection=ladybug_connection,
            mapping=cocoindex.targets.Relationships(
                rel_type="KNOWS",
                source=cocoindex.targets.NodeFromFields(
                    label="Person",
                    fields=[
                        cocoindex.targets.TargetFieldMapping(source="src", target="name"),
                    ],
                ),
                target=cocoindex.targets.NodeFromFields(
                    label="Person",
                    fields=[
                        cocoindex.targets.TargetFieldMapping(source="dst", target="name"),
                    ],
                ),
            ),
        ),
        primary_key_fields=["edge_id"],
    )


def main() -> None:
    if not os.getenv("COCOINDEX_DATABASE_URL"):
        raise RuntimeError(
            "COCOINDEX_DATABASE_URL is not set. "
            "CocoIndex requires internal DB storage to run flows."
        )

    for path in (PERSON_CSV_PATH, KNOWS_CSV_PATH):
        if not path.exists():
            raise FileNotFoundError(f"Expected CSV not found: {path}")

    cocoindex.init()
    ladybug_csv_sync_flow.setup(report_to_stdout=True)

    print(f"Watching: {PERSON_CSV_PATH}")
    print(f"Watching: {KNOWS_CSV_PATH}")
    print(f"Sync target: Ladybug API server at {LADYBUG_API_SERVER_URL}")
    print("Press Ctrl+C to stop.")

    with cocoindex.FlowLiveUpdater(
        ladybug_csv_sync_flow, cocoindex.FlowLiveUpdaterOptions(print_stats=True)
    ) as updater:
        try:
            updater.wait()
        except KeyboardInterrupt:
            print("Stopping live updater...")


if __name__ == "__main__":
    main()
```

rust/cocoindex/src/ops/targets/ladybug.rs
prrao87: Thanks George, this should be complete now. @badmonster0 I'm not sure if it makes sense to migrate the old Kuzu example over to Ladybug, it would make more sense to just directly add Ladybug as a target in v1, so that we can start promoting the v1 API in blog post. At least, that's my plan once we have a clear plan on how to use this target in v1 😄.

badmonster0:
amazing love your work @prrao87 !! excited for this thanks

georgeh0 (approved):
Thanks a lot!

PR #1498, opened 2026-01-17
docs: revise the core concepts

georgeh0:


docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Core Concepts
+description: Briefly introduce the core concepts of CocoIndex, covering state-driven sync, Effects, Components, Apps, and incremental execution across data and code changes.
+---
+
+# Core Concepts
+
+CocoIndex is a **state-driven** computing framework that **transforms** your data and keeps **persistent external state** in **sync**.
```
prrao87, line 8: Perhaps it's worth clarifying what "external" means here to a new reader? External relative to _what_? In my head, it's external to the engine (CocoIndex itself, but to a newcomer that may not be obvious).

docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
...
+You describe the state you want in your external systems; CocoIndex computes what changed since the last run and applies only the necessary updates — **incrementally** for both data and code changes.
+
+## State-driven from sources
+
+CocoIndex evaluates against the **current state of your sources** (files, APIs, databases, etc.).
```
prrao87, line 13: In "evaluates against ..." _what_? It can be ambiguous to an newcomer. Perhaps something like "CocoIndex evaluates data and code against the current state of your sources" would make more sense, just to be more explicit?

docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Components: the sync boundaries
+
+A ***Component*** is a long-lived instance (identified by a stable path) that **owns** the Effects declared within it.
+After each run, CocoIndex compares that Component’s current Effects with its prior run at the same path and applies the resulting changes **as a unit**.
+This boundary provides clear ownership and predictable scoping of updates.
+
+## Apps: the runnable unit
+
+An ***App*** is the top-level thing you run.
```
prrao87, line 66: "An App is the top-level entity of a CocoIndex pipeline" sounds better imo.

docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
...
+# Core Concepts
+
+CocoIndex is a **state-driven** computing framework that **transforms** your data and keeps **persistent external state** in **sync**.
+You describe the state you want in your external systems; CocoIndex computes what changed since the last run and applies only the necessary updates — **incrementally** for both data and code changes.
+
+## State-driven from sources
+
+CocoIndex evaluates against the **current state of your sources** (files, APIs, databases, etc.).
+You don’t wire event handlers.
+Reruns simply re-evaluate your logic on up-to-date inputs, keeping programs deterministic and easier to reason about.
```
prrao87, line 15: The program itself may not be deterministic, because LLMs may be involved - would it make sense to be explicit and state it as "keeping program execution deterministic" instead?

docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
...
+      <td>publish an “insert” event</td>
+      <td>publish an “update” event</td>
+      <td>publish a “delete/tombstone” event</td>
+    </tr>
+  </tbody>
+</table>
+
+## Components: the sync boundaries
+
+A ***Component*** is a long-lived instance (identified by a stable path) that **owns** the Effects declared within it.
```
prrao87, line 60: Finding this a little hard to understand. Maybe the diagram will illustrate this better, but in words, it makes sense to give an example of what  Component could be? Like we do for sources (databases, APIs, etc.), it would drive home the point a lot more clearly to newcomers if you replace "(identified by a stable path)" with something more approachable.

docs/docs/programming_guide/core_concepts.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Apps: the runnable unit
+
+An ***App*** is the top-level thing you run.
+It names your pipeline, binds a top-level function and its parameters, which establishes the root Component, and all work happens within the component tree rooted there.
+Given the same code and inputs, runs are repeatable; when data or code changes, only the necessary parts re-execute.
+
+## Incrementality: data + code
+
+CocoIndex minimizes work through **function-level memoization** and **change tracking**:
```
prrao87, line 72: Is memoization the default approach? Are _all_ functions memoizable? Just wondering if there are any special cases where memoization is not possible or not desirable (say we want to generate random text from an LLM as part of an effect).

PR #1499, opened 2026-01-17
docs: add the SDK overview section for the doc

georgeh0:


docs/docs/programming_guide/sdk_overview.md
```diff
@@ -4,4 +4,122 @@ title: SDK Overview
...
+|---------|-------------|
+| `cocoindex` | Main package with synchronous core APIs |
+| `cocoindex.asyncio` | Asynchronous core APIs for `async`/`await` workflows |
+
+### Sub-packages
+
+| Package | Description |
+|---------|-------------|
+| `cocoindex.connectors` | Connectors for data sources and targets (e.g., `localfs`, `postgres`, `lancedb`, `qdrant`, `google_drive`) |
+| `cocoindex.extras` | Utility modules for common tasks (e.g., text splitting, embedding with SentenceTransformers) |
```
prrao87, line 25: The boundary between extras vs. resources is a little unclear - could you explain a bit more about what qualifies as "extras" and what qualifies as "resources"? From reading this, it looks like resources are entities (e.g., Chunk) and extras are implementations that create those resources (e.g., splitter). Does that make sense?

PR #1500, opened 2026-01-17
docs: add document about component

georgeh0:


docs/docs/programming_guide/component.md
```diff
@@ -1,7 +1,80 @@
 ---
 title: Component
+description: Understanding components as the sync boundaries for effects, including Scope and mounting APIs.
 ---
 
 # Component
 
-TODO: Add content.
+A **Component** is a long-lived instance that defines the boundary where CocoIndex syncs effects to external systems. Components are identified by a stable path and own all effects declared within them.
```
prrao87, line 8: It would really help to have a diagram/visual here, or at least an example to help what we mean when we say a component is "a long-lived instance". Where does a component sit in relation to an effect? A diagram would help follow the larger structure of the pipeline and place all these visually for newcomers to more quickly grasp.

It really helps a newcomer understand better when they reason-by-analogy, so they can understand this new concept based on something else they may already be familiar with. Is "Component" like a React component? Or some other component from elsewhere?

docs/docs/programming_guide/component.md
```diff
@@ -1,7 +1,80 @@
...
+
+## Scope
+
+Every component function receives a `Scope` as its first argument. The scope carries:
+
+- A **stable path** that uniquely identifies the component across runs
```
prrao87, line 14: It might make sense to add a sentence here explaining what a "stable path" is, and why it's needed.

PR #1538, opened 2026-01-23
v1 docs - minor updates, wording,illustrations

badmonster0:


docs/static/img/concept/add.svg
georgeh0, line 1: this is not the right image for "add.svg"

docs/static/img/concept/component2.svg
georgeh0, line 1: This is not the right image

docs/static/img/concept/delete.svg
georgeh0, line 1: This is not the right image

docs/static/img/concept/update.svg
georgeh0, line 1: This is not the right image

docs/static/img/concept/add.svg
badmonster0: yep yep sorry i messed up!

PR #1541, opened 2026-01-24
Documentation v1 clarity fixes

prrao87:
- This PR contains clarity fixes that help beginners understand core concepts in v1 of CocoIndex.
- Also addresses a styling nit: For documentation, headers should ideally be "Sentence case" (only first letter capitalized) - this reads more smoothly and naturally than "Title Case" (everything capitalized). It's recommended to keep this header style consistent across all pages

### DIFF
--- a/docs/docs/programming_guide/app.md
+++ b/docs/docs/programming_guide/app.md
@@ -5,7 +5,7 @@ description: Understanding Apps as the top-level runnable unit, including creati
 
 # App
 
-An **App** is the top-level runnable unit in CocoIndex. 
+An **App** is the top-level runnable unit in CocoIndex.
 It names your pipeline and binds a main function with its parameters. When you call `app.update()`, CocoIndex runs that main function as the root [processing component](./processing_component.md) which can mount child processing components to do work and declare target states.
 
 ## Creating an app


docs/docs/programming_guide/app.md
```diff
@@ -68,17 +69,39 @@ The `report_to_stdout` option prints periodic progress updates during execution.
...
+    ├── "hello.md"             ← process_file component
+    └── "world.md"             ← process_file component
+```
+
+See [Processing Component](./processing_component.md) for how mounting and scopes define these boundaries.
+
 ## Lifespan
 
-You can define a **lifespan function** to configure settings and initialize resources before your pipeline updates. The lifespan is shared across all apps in your process by default.
+Apps are typically updated many times. A **lifespan function** defines the CocoIndex runtime lifecycle: its setup runs when the runtime starts (automatically before the first `app.update()`), and its cleanup runs when the runtime stops. Use it to configure CocoIndex and initialize shared resources that processing components can reuse.
```
prrao87, line 102: I added this mainly because the concept of lifespans was introduced very abruptly. This paragraph attempts to explain what a lifespan is, and why it matters.

docs/docs/getting_started/quickstart.md
```diff
@@ -7,6 +7,8 @@ import { GitHubButton, DocumentationButton } from '@site/src/components/ActionBu
 
 # Quickstart
 
+TODO: Replace this quickstart example with the same one that's described in the core concepts (PDF -> markdown), as that's useful to a wider pool of developers. Also, the word "flow" is used from v0 - this would need to be renamed per the v1 state-driven nomenclature, no?
```
prrao87, line 10: The word "flow" appears in some parts of this page, so that's out of date.

docs/docs/programming_guide/target_state.md
```diff
@@ -93,13 +99,13 @@ Under the hood, CocoIndex compares your declared target states with the previous
   </thead>
   <tbody>
     <tr>
-      <td>A SQL table</td>
+      <td>A database table</td>
```
prrao87, line 102: "SQL table" doesn't sound right (SQL is a language, table is an abstraction, but tables can be in other databases where SQL is not the main language). So I think the word "database" makes more sense.

georgeh0: make perfect sense!

georgeh0:
Thanks a lot for the revision! I'll merge and make further changes to resolve the TODOs!

PR #1548, opened 2026-01-25
`feat`: Add (initial) `chromadb` target integration

### LINKED ISSUE #1214: [FEATURE] support ChromaDB as a builtin target
Similar to [LanceDB](https://cocoindex.io/docs/targets/lancedb) support. Please make ChromaDB an optional dependency.

---

Haleshot:
Closes #1214                                                                                                                           
                                                                                                                                  
Preliminary ChromaDB target connector, following the LanceDB pattern from #1031. Supports both `PersistentClient`, `CloudClient` and `HttpClient`, w/ HNSW config options.

Examples and docs to follow in separate PRs. 

@jeffchuber would love to loop you in on this - also happy to collaborate on a _use-case driven_ blog post (for cocoindex x chromadb) if you're interested (as I keep adding more follow-up PRs)!

_Edit_: Addressed review comments -- removed `asyncio.to_thread` wrapping; the framework's `to_async_call` handles sync→async. Also replaced the _ad-hoc_ ("content", "text", "document") heuristic with an explicit `document_field` spec option. Removed `asyncio` import & asyncio.Lock entirely (they're no longer needed).

Also added an early guard in `get_setup_state` that raises if no vector field is present (fails fast) and wired api_key through `_State` so `apply_setup_change` can authenticate `CloudClient` for collection creation/deletion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,15 +132,6 @@ disable_error_code = ["unused-ignore"]
 [[tool.mypy.overrides]]
 # Ignore missing imports for optional dependencies from cocoindex library
 module = ["sentence_transformers", "torch", "colpali_engine", "PIL", "aiohttp", "aiomysql", "pymysql", "chromadb"]
-module = [
-    "sentence_transformers",
-    "torch",
-    "colpali_engine",
-    "PIL",
-    "aiohttp",
-    "aiomysql",
-    "pymysql",
-]
 ignore_missing_imports = true
 
 [[tool.mypy.overrides]]


python/cocoindex/targets/chromadb.py
```diff
@@ -0,0 +1,358 @@
...
+            raise ValueError("ChromaDB only supports a single key field")
+
+        # Find vector fields
+        vector_fields = [f for f in value_fields_schema if _is_vector_field(f)]
+        if len(vector_fields) > 1:
+            raise ValueError(
+                f"ChromaDB only supports a single vector field per collection, "
+                f"but found {len(vector_fields)}: {[f.name for f in vector_fields]}. "
+                f"Consider using LanceDB or Qdrant for multiple vector fields."
+            )
```
Haleshot, line 184: ChromaDB's _core_ limitation - one embedding per document. Chose to error _explicitly_ rather than silently drop fields (just in case users know to reach for `LanceDB`/`Qdrant` if they need multi-vector?)

python/cocoindex/targets/chromadb.py
```diff
@@ -0,0 +1,358 @@
...
+            metadata=metadata,
+        )
+
+    @staticmethod
+    async def prepare(
+        spec: ChromaDB,
+        setup_state: _State,
+    ) -> _MutateContext:
+        client = _get_client(spec)
+        collection = await asyncio.to_thread(client.get_collection, spec.collection_name)
```
Haleshot: `PersistentClient` is _sync-only_, so wrapped w/ `to_thread` to avoid blocking the event loop.

python/cocoindex/targets/chromadb.py
```diff
@@ -0,0 +1,358 @@
...
+                    document: str | None = None
+
+                    for field in context.value_fields_schema:
+                        if field.name == vector_field_name:
+                            continue
+                        if field.name in value:
+                            field_value = value[field.name]
+                            # Use document field for text content if present
+                            if field.name in ("content", "text", "document"):
+                                if isinstance(field_value, str):
```
Haleshot: If a field is named content/text/document, we pass it as ChromaDB's docs param (enables their _built-in_ full-text search). Other fields go to metadata.

python/cocoindex/targets/chromadb.py
```diff
@@ -0,0 +1,358 @@
...
+        # For complex types, JSON serialize
+        return json.dumps(key, sort_keys=True, default=str)
+
+
+def _convert_value_to_metadata(value: Any) -> str | int | float | bool | None:
+    if value is None:
+        return None
+    if isinstance(value, (str, int, float, bool)):
+        return value
+    return json.dumps(value, sort_keys=True, default=str)
```
Haleshot, line 122: Complex types get JSON-stringified - not ideal but better than dropping data?

jeffchuber:
awesome! please also support `CloudClient` (that uses SPANN+SPFresh)

badmonster0:
thanks a lot @Haleshot !! will get to this in a bit! @georgeh0

python/cocoindex/targets/chromadb.py
georgeh0: This is ad-hoc. How about adding a field in the `ChromaDB` spec to specify the field name for the document field?

python/cocoindex/targets/chromadb.py
georgeh0: Sounds good!

python/cocoindex/targets/chromadb.py
georgeh0: Sounds good!

python/cocoindex/targets/chromadb.py
```diff
@@ -0,0 +1,372 @@
...
+        if previous is not None:
+            should_delete = current is None or (
+                previous.key_field_schema != current.key_field_schema
+                or previous.distance_metric != current.distance_metric
+            )
+            if should_delete:
+                try:
+                    await asyncio.to_thread(
+                        client.delete_collection, key.collection_name
+                    )
```
georgeh0: Does chroma client offer async API natively?  Please directly use async API if there is.

(If there's only sync API, directly call sync API and change `apply_setup_change()` to sync function. Same for `prepare()`)

georgeh0:
Hi @Haleshot, can you briefly mention how this PR is tested? Thanks!

python/cocoindex/targets/chromadb.py
Haleshot: ChromaDB has [`AsyncHttpClient`](https://cookbook.chromadb.dev/core/clients/) for server-based setups but `PersistentClient` is _sync-only_ (no async variant exists).
For this preliminary PR version, keeping everything sync and letting the framework's `to_async_call` handle it. Can add native `AsyncHttpClient` support (which can also be configured for cloud endpoints) in a _follow-up_.

Some more references I referred to:
- https://github.com/chroma-core/chroma/blob/main/chromadb/__init__.py
- https://deepwiki.com/chroma-core/chroma/2.1-frontend-services

Haleshot:
> This is ad-hoc. How about adding a field in the `ChromaDB` spec to specify the field name for the document field?

Ahh, thanks. Added a `document_field` spec option instead. 

_Re:_ testing of the PR/integration on the whole: validation was done w/ with `uv run dmypy run` and `uv run pytest python/`.
**No integration test yet** (similar to [`lance`](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/tests/test_lancedb_index.py)) - plan to add that alongside the examples/docs _follow-up_ once the connector is settled. (that works?)

python/cocoindex/targets/chromadb.py
georgeh0: Thanks for looking into this.  Then using sync functions consistently sounds good.

georgeh0:
> _Re:_ testing of the PR/integration on the whole: validation was done w/ with `uv run dmypy run` and `uv run pytest python/`. **No integration test yet** (similar to [`lance`](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/tests/test_lancedb_index.py)) - plan to add that alongside the examples/docs _follow-up_ once the connector is settled. (that works?)

I think basically the new code is not tested at runtime (`uv run pytest python/` will trigger no tests covering this).

I think it makes more sense to do e2e test for it with an example, similar to other connectors, before merging this PR. It's the only way to verify the code is working.

Haleshot:
> I think basically the new code is not tested at runtime (`uv run pytest python/` will trigger no tests covering this).
> 
> I think it makes more sense to do e2e test for it with an example, similar to other connectors, before merging this PR. It's the only way to verify the code is working.

Ahh yup, should be adding the main testing for the integration in this PR (sorry for not having done that initially). The unit tests I had earlier on/using covered the connector in isolation(?) & didn't actually verify the what actually matters here. I'll add a relevant example shortly (will look up other existing tests for other targets).

badmonster0:
nw, thanks a lot for looking into this!! @Haleshot

Haleshot:
I’ve added the ChromaDB example now, and I referenced the other targets as a template since I said I’d look at existing examples; hope that’s alright. For `v1`, I plan to make it more unique and aligned with the updated APIs.

_Edit:_ It seems the current CI failure looks tied to the recent change in #1357 (merged after this branch was cut).

georgeh0:
Thanks @Haleshot ! The PR looks good! Can you merge with the latest main and retry the checks?

Haleshot:
> Would be nice if you can also help with adding the document for this :)

Ah yes, will add a _follow-up_ PR for this to add Chroma in [targets](https://cocoindex.io/docs/targets) docs. Thanks for the reminder (and as per always, the review comments).

PR #1556, opened 2026-01-25
feat: Add PineCone as Target Connector

### LINKED ISSUE #237: [FEATURE] Add PineCone as Target Connector

prithvi-moonshot:
Closes #237

PR adds PineCone as a Built-In Target
Includes text embedding example

### DIFF
--- a/python/cocoindex/targets/pinecone.py
+++ b/python/cocoindex/targets/pinecone.py
@@ -7,6 +7,7 @@ Supports:
 - Namespace isolation
 """
 
+import json
 import asyncio
 import dataclasses
 import logging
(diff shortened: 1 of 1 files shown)


python/cocoindex/targets/pinecone.py
```diff
@@ -0,0 +1,393 @@
...
+    if isinstance(value, (str, int, float, bool)):
+        return value
+
+    # Lists of primitives
+    if isinstance(value, (list, tuple)):
+        return [_convert_value_for_pinecone(v) for v in value]
+
+    # Convert complex types to JSON strings
+    if isinstance(value, dict):
+        import json
```
georgeh0: We can simply import `json` at top level

georgeh0 (approved):
Thanks for contributing! Most parts looks good, with minor comment.

python/cocoindex/targets/pinecone.py
prithvi-moonshot: Fixed

georgeh0:
Thanks a lot for contributing!

PR #1564, opened 2026-01-26
`ci`: migrate from `pre-commit` to `prek`

### LINKED ISSUE #1552: [FEATURE] Consider migrating from `pre-commit` to `prek`
**What is the use case?**

Following up on my https://github.com/cocoindex-io/cocoindex/pull/1482#issuecomment-3761111745, I wanted to put together a more detailed proposal for switching from pre-commit to [`prek`](https://github.com/j178/prek).

[`prek`](https://prek.j178.dev/) is a Rust-based _reimagining_ of the `pre-commit` framework designed to be a _drop-in replacement_ that's significantly faster and uses less disk space ([already being used by projects like CPython, Apache Airflow, FastAPI, Ruff](https://prek.j178.dev/#who-is-using-prek), etc.).                   

In the spirit of continuing to adopt Rust-based tooling (like we did with https://github.com/cocoindex-io/cocoindex/pull/1425), this feels like a natural fit for the project.                                                                                                                    
                                                                                                                                                   
I went through our current `.pre-commit-config.yaml` and verified that prek supports everything we're using: namely: [local hooks with language support](https://prek.j178.dev/languages/), the remote hooks from `pre-commit-hooks`, `ruff-pre-commit`, ... are [config compatible](https://prek.j178.dev/configuration/). The `SKIP` env var we use in CI becomes `PREK_SKIP` but otherwise works the same way. 

The migration itself should be minimal. The existing `.pre-commit-config.yaml` file can stay as-is since `prek` reads the same format. The CI workflows would need updating to use [`prek-action`](https://github.com/j178/prek-action) instead of installing pre-commit via pip (& the `SKIP` references would become `PREK_SKIP` as previously stated). Documentation in the [contributing guide](https://cocoindex.io/docs/contributing/guide) and [dev setup docs](https://cocoindex.io/docs/contributing/setup_dev_environment) would need the command names updated from `pre-commit` to `prek`. 

**Additional context**

The one thing that changes is pre-commit.ci. Prek doesn't have an equivalent hosted service for automatic updates (yet*?), but we can handle this through `Dependabot`. Given we've set autoupdate_schedule: 'monthly' and autofix_prs: false, this seems like a _low-friction_ tradeoff. 


---

Haleshot:
Closes #1552 

Swaps out [pre-commit](https://pre-commit.com/) for [prek](https://prek.j178.dev/).

[`prek`](https://github.com/j178/prek) is a single binary with no Python deps, so we can ditch the `virtualenv` stuff (and all those related deps like `cfgv`, `nodeenv`, `virtualenv`, etc.; read the lockfile diff for reference). CI gets cleaner too since [prek-action](https://github.com/j178/prek-action) simplifies setup to a single step; no more separate cache and install steps.

Also added a `Dependabot` entry for the `pre-commit` ecosystem (yes, *pre-commit* — Dependabot reads the same config format) for hook versions to update automatically.

Everything else is just s/pre-commit/prek/ in the docs and `SKIP` → `PREK_SKIP` in the workflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@
 [![CI](https://github.com/cocoindex-io/cocoindex/actions/workflows/CI.yml/badge.svg?event=push&color=5B5BD6)](https://github.com/cocoindex-io/cocoindex/actions/workflows/CI.yml)
 [![release](https://github.com/cocoindex-io/cocoindex/actions/workflows/release.yml/badge.svg?event=push&color=5B5BD6)](https://github.com/cocoindex-io/cocoindex/actions/workflows/release.yml)
 [![Link Check](https://github.com/cocoindex-io/cocoindex/actions/workflows/links.yml/badge.svg)](https://github.com/cocoindex-io/cocoindex/actions/workflows/links.yml)
+[![prek](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/j178/prek/master/docs/assets/badge-v0.json)](https://github.com/j178/prek)
 [![Discord](https://img.shields.io/discord/1314801574169673738?logo=discord&color=5B5BD6&logoColor=white)](https://discord.com/invite/zpA9S2DR7s)
 
 </div>


.github/dependabot.yml
Haleshot, line 1: Dependabot's pre-commit ecosystem works w/ `prek` since both read the same `.pre-commit-config.yaml` format. Changed from monthly (the old [`autoupdate_schedule`](https://github.com/cocoindex-io/cocoindex/blob/fefc5ad343eba538416b1754adab7093ae3d0392/.pre-commit-config.yaml#L3) in the ci block) to `weekly` to match the `github-actions` update cadence.

.github/workflows/_test.yml
```diff
@@ -58,14 +58,10 @@ jobs:
...
-          uses: actions/cache@v4
+        - name: Install prek
+          uses: j178/prek-action@v1
           with:
-            path: ~/.cache/pre-commit
-            key: precommit-${{ matrix.platform.runner }}-${{ matrix.python-version }}
-
-        - name: Install pre-commit
-          run: pip install pre-commit
+            install-only: true
```
Haleshot, line 64: `prek-action` handles caching automatically via actions/cache internally, so we get the same benefit with less YAML.

.github/dependabot.yml
Haleshot: Apologies for the oversight; I thought `Dependabot` had a `pre-commit` ecosystem, but after checking the [Dependabot options reference | GitHub docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#package-ecosystem) (prompted by the failed [CI run](https://github.com/cocoindex-io/cocoindex/runs/61637039437)), turns out it doesn't.

Will remove this entry in a _follow-up_; for hook version updates we'd need to run `prek autoupdate` manually or set up a scheduled action.

PR #1565, opened 2026-01-26
v1-example quickstart

badmonster0:


georgeh0:
@badmonster0 I think you need to install pre-commit hooks locally. It'll automatically run various formatter and checks before commit

https://cocoindex.io/docs/contributing/setup_dev_environment#:~:text=Install%20and%20enable%20pre%2Dcommit%20hooks

PR #1574, opened 2026-01-27
Clarify usage of `.env`<->`.env.example` across examples

### LINKED ISSUE #1568: Standardize `.env`/`.env.example` file handling across examples
This came up in https://github.com/cocoindex-io/cocoindex/pull/1470 — there's an inconsistency/ambiguity in how env files are handled across examples.

The current heuristic (as @badmonster0 explained) is: if an example contains `XYZ_API_KEY`, we provide a `.env.example` and expect users to copy it over to `.env`. Otherwise, we just use `.env` directly. This doesn't seem to be applied consistently; some examples document it:

https://github.com/cocoindex-io/cocoindex/blob/0d360cb36e267043c38192c0e74dca108aa0f57b/examples/custom_source_hn/.env.example#L2

...others don't.

I'm thinking we either standardize on `.env.example` everywhere (with clear comments like the above ^) or document a consistent pattern in a central place.

Haleshot:
This PR closes #1568 & makes the `.env.example` files explicitly say to copy to `.env` and export the .`env` (to stay consistent throughout w/ the heuristic followed across the codebase examples here: https://github.com/cocoindex-io/cocoindex/pull/1470#issuecomment-3731981049).

Let me know if I missed out on something :)

### DIFF
--- a/examples/meeting_notes_graph/README.md
+++ b/examples/meeting_notes_graph/README.md
@@ -50,10 +50,10 @@ export GOOGLE_SERVICE_ACCOUNT_CREDENTIAL=/absolute/path/to/service_account.json
 export GOOGLE_DRIVE_ROOT_FOLDER_IDS=folderId1,folderId2
 ```
 
-Alternatively, fill in your values in `.env.example` and source it:
+Alternatively, copy `.env.example` to `.env`, fill in your values, and source it:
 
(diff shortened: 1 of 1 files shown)


georgeh0:
Thanks for the cleanup!

PR #1580, opened 2026-01-28
`ci`: replace invalid dependabot > pre-commit section

Haleshot:
Follow-up to #1564 (more specifically https://github.com/cocoindex-io/cocoindex/pull/1564#discussion_r2734717631).

The `pre-commit` entry in dependabot.yml was invalid (sorry about that); Dependabot doesn't support this in their `package-ecosystem` [`config options`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-options-for-dependabotyml). This was caught by the [failed CI run](https://github.com/cocoindex-io/cocoindex/runs/61637039437).

Replaced with a scheduled GitHub Action that runs `prek autoupdate` weekly.

### DIFF
--- a/.github/workflows/prek_autoupdate.yml
+++ b/.github/workflows/prek_autoupdate.yml
@@ -20,8 +20,8 @@ jobs:
         with:
           install-only: true
 
-      - name: Run prek autoupdate
-        run: prek autoupdate
+      - name: Run prek auto-update
+        run: prek auto-update
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v7


.github/workflows/prek_autoupdate.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install prek
+        uses: j178/prek-action@v1
+        with:
+          install-only: true
+
+      - name: Run prek autoupdate
+        run: prek autoupdate
```
georgeh0: I think this should be `auto-update`?

https://prek.j178.dev/cli/#prek-auto-update

Haleshot: Ahh yes, good catch 😅

Issue #1527, opened 2026-01-22
[FEATURE] add options to update and drop subcommands for cocoindex v1 CLI


shannon06437:
**What is the use case?**
Add to these options to v1 update subcommand:
- [ ] --quiet
- [ ] --reset
- [ ] --full-reprocess

And also add to v1 drop subcommand:
- [ ] --quiet

**Describe the solution you'd like**
All options should behave the same as the v0 subcommand options.

**Additional context**

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

shannon06437:
Can I work on this issue?

badmonster0:
assigned, thanks!

badmonster0:
thanks @shannon06437, love your contribution!

PR #1590, opened 2026-01-29
Feat/v1 cli options

shannon06437:
This PR adds the update and drop options for CocoIndex v1 projects, resolving https://github.com/cocoindex-io/cocoindex/issues/1527.

Changes:

- Added `--quiet`, `--reset`, `--full-reprocess` options to update CLI subcommand 
- Added `--quiet` option to drop CLI subcommand
- Added tests (TestUpdateFlags, TestDropQuiet) for the new CLI command options

### DIFF
--- a/docs/docs/programming_guide/app.md
+++ b/docs/docs/programming_guide/app.md
@@ -62,15 +62,18 @@ Call `update()` to execute the pipeline:
 
 ```python
 # Async API
-await app.update(report_to_stdout=True)
+await app.update(report_to_stdout=True, full_reprocess=False)
 ```
 
 ```python
 # Sync API
-app.update(report_to_stdout=True)
+app.update(report_to_stdout=True, full_reprocess=False)
 ```
 
-The `report_to_stdout` option prints periodic progress updates during execution.
+**Parameters:**
+
+- `report_to_stdout` option prints periodic progress updates during execution.
+- `full_reprocess` option reprocesses everything and invalidates existing caches. This forces all components to re-execute and all target states to be re-applied, even if they haven't changed.
 
 When you update an App, CocoIndex:
 
--- a/python/tests/core/test_full_reprocess.py
+++ b/python/tests/core/test_full_reprocess.py
@@ -1,0 +1,208 @@
+"""Tests for full_reprocess behavior."""
+
+import cocoindex as coco
+
+from typing import NamedTuple
+
+from tests import common
+from tests.common.target_states import GlobalDictTarget, DictDataWithPrev, Metrics
+
+
+coco_env = common.create_test_env(__file__)
+
+
+class SourceDataEntry(NamedTuple):
+    name: str
+    version: int
+    content: str
+
+    def __coco_memo_key__(self) -> object:
+        return (self.name, self.version)
+
+
+_source_data: dict[str, SourceDataEntry] = {}
+_metrics = Metrics()
+
+
+@coco.function(memo=True)
+def _declare_dict_entry(entry: SourceDataEntry) -> None:
+    """Memoized function that declares a target state."""
+    _metrics.increment("calls")
+    coco.declare_target_state(GlobalDictTarget.target_state(entry.name, entry.content))
+
+
+@coco.function(memo=True)
+def _declare_dict_data_memoized(data_version: int) -> None:
+    """Main function that mounts child components. Memoized to test component memoization.
+
+    Takes data_version as argument to enable component memoization.
+    """
+    _metrics.increment("root_component")
+    for entry in _source_data.values():
+        coco.mount(coco.component_subpath(entry.name), _declare_dict_entry, entry)
+
+
+@coco.function
+def _declare_dict_data() -> None:
+    """Main function that mounts child components. Not memoized."""
+    _metrics.increment("root_component")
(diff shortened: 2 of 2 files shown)


shannon06437:
Please take a look!

python/cocoindex/cli.py
```diff
@@ -535,10 +567,31 @@ def update(app_target: str) -> None:
...
+                        return
+
+                persisted_names = _get_persisted_app_names(env)
+                if app._name in persisted_names:
+                    await _drop_app(app, report_to_stdout=not quiet)
+
+            # --full-reprocess: invalidate memoization caches before running
+            if full_reprocess:
+                core_app = await app._get_core()
+                _core.clear_all_memoization(core_app)
```
georgeh0: Clearing memoization is not sufficient. We also keep track of target states. If a target states is unchanged, we skip writing to the target. But clearing target states is also not the right way, since we don't want to leak previous states (if they no longer exists) into the target.

Instead of clearing memoization before running the app, I think the right way is to pass an option down to indicate if we want to reset or not, and respect the option in the engine code. You may have a look at the existing implementation of this flag to get an idea.

shannon06437:
please take a look!

rust/py/src/context.rs
```diff
@@ -26,6 +26,10 @@ impl PyComponentProcessorContext {
         Ok(())
     }
 
+    fn full_reprocess(&self) -> bool {
+        self.0.full_reprocess()
+    }
+
```
georgeh0: This seems not needed. It's not used by the Python code.

rust/core/src/engine/execution.rs
```diff
@@ -785,15 +794,24 @@ pub(crate) async fn submit<Prof: EngineProfile>(
...
+                };
+                let prev_states = if full_reprocess {
+                    vec![]
+                } else {
+                    item.states
+                        .iter()
+                        .filter_map(|(_, s)| s.as_ref())
+                        .map(|s_bytes| Prof::TargetStateTrackingRecord::from_bytes(s_bytes))
+                        .collect::<Result<Vec<_>>>()?
+                };
```
georgeh0: We want to keep `prev_states` even on `full_reprocess`. Otherwise the target handler may think the states not exist for sure, and may always do a "insert" (which should actually be "update" or "replace" when we want to do a full reprocess).

rust/core/src/engine/execution.rs
```diff
@@ -1123,13 +1143,16 @@ fn finalize_fn_call_memoization<Prof: EngineProfile>(
         result.all_memos_fps.insert(*fp);
 
         if memo.already_stored {
-            for child_component in &memo.child_components {
-                stable_path_set.add_child(child_component.as_ref(), StablePathSet::Component)?;
+            if !full_reprocess {
```
georgeh0: The added check here is unnecessary. We already ignore the already stored memo entries on read time. `already_stored` should always be `false` for full reprocessing.

rust/core/src/engine/execution.rs
```diff
@@ -1142,23 +1165,25 @@ fn finalize_fn_call_memoization<Prof: EngineProfile>(
...
-    while let Some(fp) = deps_to_process.pop_front() {
-        if !result.all_memos_fps.insert(fp) {
-            continue;
-        }
-        let Some(memo) = read_fn_call_memo_with_txn(&rtxn, db, comp_ctx, fp)? else {
-            continue;
-        };
-        for child_component in &memo.child_components {
-            stable_path_set.add_child(child_component.as_ref(), StablePathSet::Component)?;
+    if !full_reprocess {
```
georgeh0: This is unnecessary either, since `deps_to_process` is always empty in this case.

But this is a good optimization actually (avoid creating another `rtxn`). I think we can generalize the condition to `if !deps_to_process.is_empty()` here.

rust/core/src/engine/context.rs
```diff
@@ -130,6 +130,7 @@ struct ComponentProcessorContextInner<Prof: EngineProfile> {
     components_readiness: ComponentBgChildReadiness,
 
     processing_stats: ProcessingStats,
+    full_reprocess: bool,
```
georgeh0: This is only meaningful for "Build" and not meaningful for "Delete".

I think we can consider adding it into `ComponentProcessingAction::Build`. We may create a new struct like this (similar to `ComponentDeleteContext`):

```rust
struct ComponentBuildContext {
  state: Mutex<Option<ComponentBuildingState<Prof>>>, // The existing field
  full_reprocess: bool
}
```

docs/docs/cli_commands.md
georgeh0, line 1: Since we have added new argument to our `App.update()` API, we'll also want to update the doc for the API part.

python/tests/cli/test_cli.py
georgeh0, line 1: Besides tests for CLI, it'll also be great if we add or modify some tests under `tests/core` to cover the behavior of full-reprocessing, e.g. make sure memoized functions are executed / unchanged target states are applied again by testing metrics.

shannon06437:
please take a look!

georgeh0 (approved):
Looks good! Thanks a lot!

Issue #1559, opened 2026-01-25
[FEATURE] Require target state keys to fit into type StableKey (similar to component key)


georgeh0:
Currently, our target state keys ([Rust](https://github.com/cocoindex-io/cocoindex/blob/76eecb79f01c4fcba32f1119b126333f4d0ba832/rust/core/src/engine/profile.rs#L42-L49), [Python](https://github.com/cocoindex-io/cocoindex/blob/76eecb79f01c4fcba32f1119b126333f4d0ba832/python/cocoindex/_internal/target_state.py#L27-L28)) can be any Python type that supports eq/hash/etc. We want it to be `StableKey` ([Rust](https://github.com/cocoindex-io/cocoindex/blob/76eecb79f01c4fcba32f1119b126333f4d0ba832/rust/core/src/state/stable_path.rs#L5-L16), [Python](https://github.com/cocoindex-io/cocoindex/blob/76eecb79f01c4fcba32f1119b126333f4d0ba832/python/cocoindex/_internal/core.pyi#L17-L18)). This will allow various tools (CLI, CocoInsight) to read/display them easily in standard ways, and also improves efficiency.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

Geoff-Robin:
@georgeh0 Hey, I'd like to pick this issue up!

georgeh0:
Thanks @Geoff-Robin for taking this! Note that this is a `v1` feature request and please work on the `v1` branch.

badmonster0:
mark as done, thanks @Geoff-Robin #1625

PR #1625, opened 2026-02-05
Feat/require target state keys to fit into type stable key

Geoff-Robin:
### Key Changes

* **Removed `PyKey`**
    * The custom `PyKey` struct and its serialization logic were entirely deleted from `rust/py/src/value.rs`. 
    * This type previously allowed wrapping arbitrary Python objects as keys by serializing them.

* **Adoption of `StableKey`**
    * The code was updated to use `StableKey` instead. 
    * This enforces a stricter, more deterministic type system for keys used in target states, likely improving reliability and cross-connector compatibility.

* **Rust Implementation Update**
    * **Target State:** Updated `TargetState` in `rust/py/src/target_state.rs` to interact with this new key type.
    * **Path Bridging:** Added support code in `rust/core/src/state/stable_path.rs` and `rust/py/src/stable_path.rs` to bridge the new key references.
    
Resolves this particular [issue](https://github.com/cocoindex-io/cocoindex/issues/1559)

### DIFF
--- a/rust/py/src/stable_path.rs
+++ b/rust/py/src/stable_path.rs
@@ -33,6 +33,8 @@ impl<'py> IntoPyObject<'py> for PyStableKey {
             }
 
             StableKey::Fingerprint(fp) => PyBytes::new(py, fp.as_ref()).into_bound_py_any(py),
+
+            StableKey::Symbol(s) => PySymbol(s).into_pyobject(py)?.into_bound_py_any(py),
         }
     }
 }


badmonster0:
thanks a lot @Geoff-Robin will get back on this shortly! thanks a lot for your contributions!

rust/core/src/state/stable_path.rs
```diff
@@ -186,6 +186,34 @@ impl storekey::Decode for StableKey {
...
+                let mut fingerprinter = utils::fingerprint::Fingerprinter::default();
+                let bytes = crate::engine::profile::Persist::to_bytes(self)
+                    .expect("StableKey encoding for fingerprint");
+                fingerprinter.write_raw_bytes(&bytes);
+                fingerprinter.into_fingerprint()
+            }
+        }
+    }
+}
+
```
georgeh0: We don't need to add these. `StableKey` itself is already serializable and deserializable.

Since `StableKey` is a type defined in the `core` crate, we can remove [`TargetStateKey`](https://github.com/cocoindex-io/cocoindex/blob/5ceeebf1f89368b7823c2d231c8a6deff917602c/rust/core/src/engine/profile.rs#L42-L49) and directly use `StableKey`. So we can directly serialize/deserialize it in the engine code.

rust/py/src/stable_path.rs
```diff
@@ -1,10 +1,33 @@
 use crate::prelude::*;
 
+use pyo3::conversion::IntoPyObjectExt;
 use pyo3::exceptions::PyTypeError;
 use pyo3::types::{PyBool, PyBytes, PyInt, PyList, PyString, PyTuple};
 
 use cocoindex_core::state::stable_path::{StableKey, StablePath};
 
+impl PyStableKey {
```
georgeh0: I think we want to implement `IntoPyObject` for `PyStableKey` instead (which matches `FromPyObject` below)

rust/py/src/target_state.rs
```diff
@@ -152,19 +154,18 @@ impl PyTargetStateProvider {
 
 #[pyfunction]
 pub fn declare_target_state<'py>(
-    py: Python<'py>,
+    _py: Python<'py>,
```
georgeh0: We no longer need this argument.

python/cocoindex/connectors/lancedb/_target.py
```diff
@@ -670,6 +670,9 @@ def reconcile(
         coco.TargetReconcileOutput[_TableAction, _TableTrackingRecord, _RowHandler]
         | None
     ):
+        if isinstance(key, tuple) and not isinstance(key, _TableKey):
+            key = _TableKey(*key)
+
```
georgeh0: The type annotation `key: _TableKey` is not right now. I think we want to use `key: StableKey` instead, since it's the actual type being passed around. We can remove [`KeyT` / `KeyT_contra`](https://github.com/cocoindex-io/cocoindex/blob/5ceeebf1f89368b7823c2d231c8a6deff917602c/python/cocoindex/_internal/target_state.py#L27-L28) here, and directly use `StableKey`.

This will also force each `reconcile()` implementation explicitly convert the type, so implementers won't forget.

Geoff-Robin: @georgeh0 I've tried doing this but I've encountered a data loss bug in SQLite caused by unstable instance-bound sinks and fixed it by using stable global sinks. Is this fine or maybe I could take a different approach
The tests that i started to fail were:
- test_sqlite_target.py::test_regular_table_vs_vec0_switch
 started failing around line 1232 > ` assert len(data) == 0 `
- test_sqlite_target.py::test_vec0_virtual_table_basic
 start failing around line 660  > `assert len(data) == 1`

I will be pushing the code for this to get the whole view

rust/core/src/engine/profile.rs
```diff
@@ -39,14 +39,7 @@ pub trait EngineProfile: Debug + Clone + PartialEq + Eq + Hash + Default + 'stat
...
     type TargetHdl: TargetHandler<Self>;
-    type TargetStateKey: Clone
-        + std::fmt::Debug
-        + Send
-        + Eq
-        + Hash
-        + Persist
-        + StableFingerprint
-        + 'static;
+    // TargetStateKey is removed in favor of using StableKey directly.
```
georgeh0: Why do we need to keep this line of comment?

rust/core/src/state/target_state_path.rs
```diff
@@ -1,11 +1,13 @@
 use crate::prelude::*;
 use serde::{Deserialize, Serialize};
 
+use crate::state::stable_path::StableKey;
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub struct TargetStatePath(Arc<[utils::fingerprint::Fingerprint]>);
+pub struct TargetStatePath(pub Arc<[StableKey]>);
```
georgeh0: We don't want to change this part. We still want to use fingerprint here. (We only change whatever was `TargetStateKey` to `StableKey`, but `Fingerprint` is a different thing)

For the `TargetStatePath`, the only thing we want to change is when we create/concat the path with a new key part, i.e. we compute fingerprint from `StableKey` instead of the old key.

rust/py/src/target_state.rs
```diff
@@ -211,7 +210,7 @@ pub fn register_root_target_states_provider(
         .unwrap()
         .register(
             TargetStatePath::new(
-                utils::fingerprint::Fingerprint::from(&name).into_py_result()?,
+                cocoindex_core::state::stable_path::StableKey::Str(Arc::from(name)),
```
georgeh0: We don't need to change this. It's OK to still get fingerprint from the name directly.

python/cocoindex/connectors/lancedb/_target.py
georgeh0: > a data loss bug in SQLite caused by unstable instance-bound sinks and fixed it by using stable global sinks.

Thanks for trying out. We want to understand what happens here.

Why the instance-bound sinks were unstable? They're instance-bound but they're stable. And I don't know why a key type change will affect the logic there.

As I commented above, I think this PR is changing more than we need: we still want to use Fingerprint as `TargetStatePath`. I'm not sure if the failed tests are related to some unexpected changes like this. Please fix that and try again if the sqlite tests will work with the previous code (with necessary changes for key types). Thanks!

Geoff-Robin: It seems like using StableKey directly in TargetStatePath caused the errors, I will be reverting back to fingerprint.
This will undo the changes i've made just now

python/cocoindex/connectors/lancedb/_target.py
```diff
@@ -434,12 +437,22 @@ def _build_pyarrow_schema(self) -> pa.Schema:
...
+        # Key conversion
+        if isinstance(key, tuple):
+            # _RowKey is tuple[Any, ...]
+            pass
+        else:
+            # Should be tuple for row key
+            if not isinstance(key, tuple):
+                # This might happen if single-value PK is passed as scalar
+                key = (key,)
+        key = cast(_RowKey, key)
```
georgeh0: If the key type is not a expected, raise an exception.

Then we won't need the `cast` (when the `isinstance()` pass, it's always a tuple that satisfies `_RowKey`). `cast` itself doesn't turn value in an unexpected type to the expected type.

Please update `reconcile()` for other targets accordingly.

python/cocoindex/connectors/sqlite/_target.py
```diff
@@ -416,131 +424,38 @@ def __init__(
         self._table_name = table_name
         self._table_schema = table_schema
         self._is_virtual_table = is_virtual_table
-        self._sink = coco.TargetActionSink[_RowAction, None].from_fn(
```
georgeh0, line 419: Can we revert this change, and only keep necessary changes related to `StableKey`?

rust/core/src/engine/execution.rs
```diff
@@ -840,7 +839,8 @@ pub(crate) async fn submit<Prof: EngineProfile>(
                                 );
                                 continue;
                             };
-                            let effect_key = Prof::TargetStateKey::from_bytes(item.key.as_ref())?;
+                            let effect_key: StableKey = storekey::decode(item.key.as_ref())
+                                .map_err(|e| internal_error!("Failed to decode StableKey: {e}"))?;
```
georgeh0: Let's avoid this `map_err()` here. It'll strip structured information (like backtrace) from the original error. A simple `?` will just work.

rust/core/src/engine/execution.rs
```diff
@@ -884,18 +884,19 @@ pub(crate) async fn submit<Prof: EngineProfile>(
 
             // Deal with new target states
             for (target_state_path, target_state) in declared_target_states_to_process {
-                let effect_key_bytes = target_state.key.to_bytes()?;
+                let effect_key_bytes = storekey::encode_vec(&target_state.item_key)
+                    .map_err(|e| internal_error!("Failed to encode StableKey: {e}"))?;
```
georgeh0, line 887: Same here, `map_err` is not needed.

georgeh0:
Hi @Geoff-Robin, thanks for contributing! I think the PR is almost there.

Since this is a breaking change, we want to make this PR merged sooner.  Let me know if you have any pending changes and want to commit soon. Otherwise, I can also merge the PR first and do a quick cleanup on my side for these comments. Thanks again!

Geoff-Robin:
@georgeh0 Yeah, I've just got started. I'll push them in around an hour.

Geoff-Robin:
@georgeh0 I have made the changes completely however
[this review](https://github.com/cocoindex-io/cocoindex/pull/1625#discussion_r2795173048) couldn't be adressed because because its trait-object error type doesn't satisfy the Sized bound required for automatic conversion to the internal error type.

I'm not sure why the tests are failing on merging with v1 and before. Any idea why it's causing? To my knowledge python tests are failing due to circular imports

georgeh0:
> @georgeh0 I have made the changes completely however [this review](https://github.com/cocoindex-io/cocoindex/pull/1625#discussion_r2795173048) couldn't be adressed because because its trait-object error type doesn't satisfy the Sized bound required for automatic conversion to the internal error type.
> 
> I'm not sure why the tests are failing on merging with v1 and before. Any idea why it's causing? To my knowledge python tests are failing due to circular imports

The actual problem I can see from the failed test is an enum variant (`StableKey::Symbol`, which just added yesterday) is not handled in your new function:

Screenshot 2026-02-12 at 11 06 59 AM

Geoff-Robin:
Yeah just realised I'll commit it right now

georgeh0:
Thanks for the contribution! Merging.

Issue #1622, opened 2026-02-04
[FEATURE] Support listing paths as trees for v1 CLI


shannon06437:
**What is the use case?**
Add an option to the show command to allow better discoverability and inspection of stable paths.

**Describe the solution you'd like**
Add a --tree flag to the show command that renders stable paths in a hierarchical tree view.
Nodes in the tree should be annotated to indicate which are components.

**Additional context**

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

shannon06437:
Can I work on this issue?

badmonster0:
assigned, thanks for proposing the feature and taking it!

Haleshot:
Can be closed via #1627 :)

PR #1627, opened 2026-02-06
Feat/v1 cli tree option

shannon06437:
This PR adds the --tree flag to the show CLI subcommand for CocoIndex v1 projects, resolving https://github.com/cocoindex-io/cocoindex/issues/1622.

Changes:

- Added list_stable_paths_with_types method 
- Added parts() function to StablePath to get the sequence of StableKeys
- Added tests (TestShowTree) for the new option

### DIFF
--- a/rust/core/src/inspect/db_inspect.rs
+++ b/rust/core/src/inspect/db_inspect.rs
@@ -34,9 +34,9 @@ pub fn list_stable_paths<Prof: EngineProfile>(app: &App<Prof>) -> Result<Vec<Sta
     Ok(result)
 }
 
-/// Represents a stable path with its node type information.
+/// Represents a stable path with metadata (e.g. node type); more properties may be added.
 #[derive(Clone, Debug)]
-pub struct StablePathWithType {
+pub struct StablePathInfo {
     pub path: StablePath,
     pub node_type: db_schema::StablePathNodeType,
 }
@@ -44,13 +44,13 @@ pub struct StablePathWithType {
 // Re-export StablePathNodeType for use in Python bindings
 pub use db_schema::StablePathNodeType;
 
-/// Returns a stream of stable paths with their node types.
+/// Returns a stream of stable paths with their metadata (e.g. node type).
 /// LMDB iteration runs on a dedicated thread (RoTxn/cursors are !Send); items are sent over a channel.
-pub fn iter_stable_paths_with_types<Prof: EngineProfile>(
+pub fn iter_stable_paths<Prof: EngineProfile>(
     app: &App<Prof>,
-) -> impl Stream<Item = Result<StablePathWithType>> + Send + 'static {
+) -> impl Stream<Item = Result<StablePathInfo>> + Send + 'static {
     let app_ctx = app.app_ctx().clone();
-    let (tx, rx) = tokio::sync::mpsc::channel::<Result<StablePathWithType>>(128);
+    let (tx, rx) = tokio::sync::mpsc::channel::<Result<StablePathInfo>>(128);
 
     std::thread::spawn(move || {
         let result: Result<()> = (|| {
@@ -86,7 +86,7 @@ pub fn iter_stable_paths_with_types<Prof: EngineProfile>(
                     }
                 };
 
-                let item = StablePathWithType { path, node_type };
+                let item = StablePathInfo { path, node_type };
                 if tx.blocking_send(Ok(item)).is_err() {
                     break;
                 }
--- a/rust/py/src/inspect.rs
+++ b/rust/py/src/inspect.rs
@@ -51,29 +51,27 @@ impl PyStablePathNodeType {
     }
 }
 
-#[pyclass(name = "StablePathWithType")]
+#[pyclass(name = "StablePathInfo")]
 #[derive(Clone)]
-pub struct PyStablePathWithType {
+pub struct PyStablePathInfo {
     #[pyo3(get)]
     pub path: PyStablePath,
     #[pyo3(get)]
     pub node_type: PyStablePathNodeType,
 }
 
-/// Python async iterator that yields `StablePathWithType` items one-by-one (no blocking calls, no forwarder).
-#[pyclass(name = "StablePathWithTypeAsyncIterator")]
-pub struct PyStablePathWithTypeAsyncIterator {
+/// Python async iterator that yields `StablePathInfo` items one-by-one (no blocking calls, no forwarder).
+#[pyclass(name = "StablePathInfoAsyncIterator")]
+pub struct PyStablePathInfoAsyncIterator {
     /// Stream wrapped in async Mutex to allow &self access without blocking Python thread.
     /// Pin<Box<...>> is needed because streams are not Unpin.
     stream: Arc<
-        tokio::sync::Mutex<
-            Pin<Box<dyn Stream<Item = Result<db_inspect::StablePathWithType>> + Send>>,
-        >,
+        tokio::sync::Mutex<Pin<Box<dyn Stream<Item = Result<db_inspect::StablePathInfo>> + Send>>>,
     >,
 }
 
 #[pymethods]
-impl PyStablePathWithTypeAsyncIterator {
+impl PyStablePathInfoAsyncIterator {
     fn __aiter__(slf: PyRef<'_, Self>) -> PyRef<'_, Self> {
         slf
     }
@@ -91,7 +89,7 @@ impl PyStablePathWithTypeAsyncIterator {
                     Python::attach(|py| {
                         Py::new(
                             py,
-                            PyStablePathWithType {
+                            PyStablePathInfo {
                                 path: PyStablePath(item.path),
                                 node_type: PyStablePathNodeType(item.node_type),
                             },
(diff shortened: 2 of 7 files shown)


shannon06437:
Please take a look!

rust/py/src/stable_path.rs
```diff
@@ -71,4 +73,42 @@ impl PyStablePath {
...
+}
+
+fn stable_key_to_py(py: Python<'_>, key: &StableKey) -> PyResult<Py<PyAny>> {
+    match key {
+        StableKey::Null => Ok(py.None().into()),
+        StableKey::Bool(b) => {
+            let builtins = py.import("builtins")?;
+            let bool_type = builtins.getattr("bool")?;
+            let obj = bool_type.call1((*b,))?;
+            Ok(obj.into())
```
georgeh0: Converting a boolean to Python value can be simpler than this. Don't need import anything.

(and whenever we need to import, we usually only import once instead of per call)

rust/py/src/stable_path.rs
```diff
@@ -71,4 +73,42 @@ impl PyStablePath {
...
+        }
+        StableKey::Int(i) => Ok(i.into_pyobject(py)?.into()),
+        StableKey::Str(s) => Ok(s.as_ref().into_pyobject(py)?.into()),
+        StableKey::Bytes(b) => Ok(PyBytes::new(py, b.as_ref()).into_pyobject(py)?.into()),
+        StableKey::Uuid(u) => {
+            let uuid_module = py.import("uuid")?;
+            let uuid_class = uuid_module.getattr("UUID")?;
+            let uuid_obj = uuid_class.call1((u.to_string(),))?;
+            Ok(uuid_obj.into_pyobject(py)?.into())
```
georgeh0: This conversion natively supported by PyO3:

https://pyo3.rs/main/doc/pyo3/uuid/

rust/py/src/stable_path.rs
```diff
@@ -71,4 +73,42 @@ impl PyStablePath {
...
+        StableKey::Null => Ok(py.None().into()),
+        StableKey::Bool(b) => {
+            let builtins = py.import("builtins")?;
+            let bool_type = builtins.getattr("bool")?;
+            let obj = bool_type.call1((*b,))?;
+            Ok(obj.into())
+        }
+        StableKey::Int(i) => Ok(i.into_pyobject(py)?.into()),
+        StableKey::Str(s) => Ok(s.as_ref().into_pyobject(py)?.into()),
+        StableKey::Bytes(b) => Ok(PyBytes::new(py, b.as_ref()).into_pyobject(py)?.into()),
```
georgeh0: For most type conversions supported by PyO3, I think we can directly use `into_py_any()`

rust/core/src/inspect/db_inspect.rs
```diff
@@ -31,6 +32,54 @@ pub fn list_stable_paths<Prof: EngineProfile>(app: &App<Prof>) -> Result<Vec<Sta
...
+            continue;
+        };
+
+        let node_type = get_path_node_type(db, &txn, parent_ref, key)?
+            .unwrap_or(db_schema::StablePathNodeType::Directory);
+        out.push((path, node_type == db_schema::StablePathNodeType::Component));
+    }
+
+    Ok(out)
+}
```
georgeh0: Ideally we want to stream out the output (similar to `list_stable_paths()`; that one was created for a quick bootstrapping so was a simple implementation).

Let's return an [async stream](https://docs.rs/async-stream/latest/async_stream/) (an alternative way is using a `Iterator` since our currently db read has no async logic, but later we'll support more types of internal databases, so starting from an async interface is a good idea).

rust/core/src/inspect/db_inspect.rs
```diff
@@ -31,6 +32,54 @@ pub fn list_stable_paths<Prof: EngineProfile>(app: &App<Prof>) -> Result<Vec<Sta
     Ok(result)
 }
 
+/// List stable paths along with whether each node is a component
+pub fn list_stable_paths_with_types<Prof: EngineProfile>(
+    app: &App<Prof>,
+) -> Result<Vec<(StablePath, bool)>> {
```
georgeh0: For a function's return type, it's usually a good idea by putting it in a more structured way (more readable, stronger type enforcement, future proof):

- Use `StablePathNodeType` instead of `bool`. The meaning will be more clear.
- Define a struct.

shannon06437:
Please take a look!

rust/core/src/inspect/db_inspect.rs
```diff
@@ -31,6 +33,62 @@ pub fn list_stable_paths<Prof: EngineProfile>(app: &App<Prof>) -> Result<Vec<Sta
...
+                get_path_node_type(db, &txn, parent_ref, key)?
+                    .unwrap_or(db_schema::StablePathNodeType::Directory)
+            } else {
+                db_schema::StablePathNodeType::Component
+            }
+        };
+        results.push(Ok(StablePathWithType { path, node_type }));
+    }
+
+    Ok(stream::iter(results))
```
georgeh0: This buffers all results in a `Vec` first. It vanishes the benefit of using streaming interface.

You may have have a look at the existing usage of `try_stream!`

https://github.com/cocoindex-io/cocoindex/blob/fe6b6789ffdf13c1cb5eba9bf8a56d6c53b3caa9/rust/cocoindex/src/execution/db_tracking.rs#L372

shannon06437:
Note: 
I added both a streaming and a collect variant of list_stable_paths_with_types.

The streaming version (list_stable_paths_with_types_stream) uses `try_stream!` to avoid buffering all results upfront and to establish an async-friendly API surface for future backends.

However, the current LMDB/heed backend relies on thread-affine read transactions (RoTxn), which do not implement Send. As a result, the stream cannot be safely moved across threads or wrapped as a Python Future (this is the Send error observed when attempting an async PyO3 bridge in inspect.rs):
image

As a result, Rust core exposes a streaming API, while Python bindings rely on a synchronous collector for now.

shannon06437:
Please take a look!

georgeh0:
Hi @shannon06437,

Thanks for trying out!

One pattern is to use a dedicated thread to iterate in the LMDB and use a [channel](https://docs.rs/tokio/latest/tokio/sync/index.html) to communicate. We may use [`LocalSet`](https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html#method.block_on) to run a `!Send` future on the dedicated thread.

Please have another try, and let me know if you have any questions. Thanks!

shannon06437:
Please take a look!

rust/core/src/inspect/db_inspect.rs
```diff
@@ -31,6 +34,100 @@ pub fn list_stable_paths<Prof: EngineProfile>(app: &App<Prof>) -> Result<Vec<Sta
...
+        if let Err(err) = result {
+            let _ = tx.blocking_send(Err(err));
+        }
+    });
+
+    try_stream! {
+        while let Some(item) = rx.recv().await {
+            yield item?;
+        }
+    }
```
georgeh0: Since we're using channel now, we can try this to convert the receiver to a stream: https://docs.rs/tokio-stream/0.1.18/tokio_stream/wrappers/struct.ReceiverStream.html

rust/py/src/inspect.rs
```diff
@@ -12,6 +15,69 @@ pub fn list_stable_paths(app: &PyApp) -> PyResult<Vec<PyStablePath>> {
...
+        pin_mut!(stream);
+        let mut results = Vec::new();
+        while let Some(item) = stream.next().await {
+            let item = item.into_py_result()?;
+            results.push(PyStablePathWithType {
+                path: PyStablePath(item.path),
+                node_type: PyStablePathNodeType(item.node_type),
+            });
+        }
+        Ok(results)
```
georgeh0: We're still buffering all results in memory here.

What ideally it needs to be an object that implements Python's [`AsyncIterator`](https://docs.python.org/3/library/collections.abc.html#collections.abc.AsyncIterator) protocol.

shannon06437:
Please take a look!

rust/py/src/inspect.rs
```diff
@@ -12,6 +19,111 @@ pub fn list_stable_paths(app: &PyApp) -> PyResult<Vec<PyStablePath>> {
...
+                    Python::attach(|py| {
+                        Py::new(
+                            py,
+                            PyStablePathWithType {
+                                path: PyStablePath(item.path),
+                                node_type: PyStablePathNodeType(item.node_type),
+                            },
+                        )
+                        .map(|p| p.into_any())
+                    })
```
georgeh0, line 98: Wondering is the `Python::attach(|py| { Py:new(py, ...) })` layer really needed here? i.e. will  `PyStablePathWithType { ... }` just work?

(just wonder if it can be simpler, it's OK to keep the out layer if it has to be there)

rust/py/src/inspect.rs
```diff
@@ -12,6 +19,111 @@ pub fn list_stable_paths(app: &PyApp) -> PyResult<Vec<PyStablePath>> {
...
+                    .map_err(|e| e.into())
+                }
+            }
+        })
+    }
+}
+
+#[pyfunction]
+pub fn iter_stable_paths_with_types<'py>(
```
georgeh0: We may just call it `iter_stable_paths`, and the type `PyStablePathWithType` can be something like `PyStablePathInfo`. We'll unlikely have separate functions for different properties, and very likely more properties will be added into it.

(BTW, since we have it now, we can clean up the existing `list_stable_paths()` which only provides a subset of functionality; the `list_stable_paths()` may be handy for testing purpose, so it's OK to leave a function on Python side which delegates to this new API)

rust/py/src/stable_path.rs
```diff
@@ -141,4 +142,33 @@ impl PyStablePath {
...
+            let mut items: Vec<Py<PyAny>> = Vec::with_capacity(arr.len());
+            for item in arr.iter() {
+                items.push(stable_key_to_py(py, item)?);
+            }
+            let py_tuple = PyTuple::new(py, items)?;
+            Ok(py_tuple.into_py_any(py)?)
+        }
+        StableKey::Fingerprint(fp) => Ok(fp.to_string().into_py_any(py)?),
+    }
 }
```
georgeh0: This one seems overlapped with the `into_pyobject()` above. We may not need this.

rust/py/src/inspect.rs
shannon06437: I believe the outer layer has to be there because we can’t use the `py` token directly in the async block. `future_into_py` requires `Send`, and `Python<'_>` isn’t `Send` (it would be held across an await). `Python::attach` re-acquires the GIL inside the future so `py` isn’t captured and the future stays `Send`.

shannon06437:
Please take a look!

rust/py/src/inspect.rs
shannon06437: image

georgeh0: Makes sense! Thanks for trying out!

georgeh0 (approved):

PR #1663, opened 2026-02-12
fix: ChromaDB collection not created on first run

### LINKED ISSUE #1660: [BUG] Chromadb export does not create a new collection
If you start a flow with an export to chromadb for a new collection, the flow will file with the error:
chromadb.errors.NotFoundError: Collection [Notes] does not exists
with of course your collection name instead of Notes

I am using cocoindex 0.3.30

I already fix the problem by changing line 286 in chromadb.py from:
`        collection = client.get_collection(spec.collection_name)`
to:
`        collection = client.get_or_create_collection(spec.collection_name)`

Haleshot:
Fixes #1660

[`prepare()`](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/targets/chromadb.py#L281) was using `get_collection()` which throws if the collection doesn't exist yet. Switched to [`get_or_create_collection()`](https://docs.trychroma.com/docs/collections/manage-collections#getting-collections); idempotent, returns existing or creates new.

Thanks @RodTol for flagging this!

### DIFF
--- a/python/cocoindex/targets/chromadb.py
+++ b/python/cocoindex/targets/chromadb.py
@@ -283,7 +283,7 @@ class _Connector:
         setup_state: _State,
     ) -> _MutateContext:
         client = _get_client(spec)
-        collection = client.get_collection(spec.collection_name)
+        collection = client.get_or_create_collection(spec.collection_name)
 
         return _MutateContext(
             client=client,


badmonster0:
thanks for the quick fix!

PR #1673, opened 2026-02-15
`feat`: add `Entire` Session Search example

Haleshot:
Semantic search over AI coding sessions captured by [Entire](https://entire.io/), powered by CocoIndex's `v1` async APIs.                               
                                                                                                                                  
Showcases `mount_each`, `map`, `TableTarget`, `ContextKey`, and _memo-based_ caching with a real-world use case; searching across accumulated coding (/agentic) sessions for things like "how did I fix that auth bug" or "which session touched the config parser?", etc. 

Will update this as I test with more sessions locally. Also planning a blog post around this if you deem it worthy of one.

cc: @khaong, @Soph: hope you don't mind the ping! I've been using [`entire/cli`](https://github.com/entireio/cli) to capture my coding sessions and figured the [checkpoint format](https://docs.entire.io/web/checkpoints) would be a nice fit for CocoIndex's incremental pipelines; sessions accumulate over time, and CocoIndex only _reprocesses_ what changed. Seemed to make sense with a mental model initially, so I built this out. Would love to hear if anything looks off _w.r.t._ the checkpoint format assumptions.

### DIFF
--- a/examples/entire_session_search/.env
+++ b/examples/entire_session_search/.env
@@ -1,1 +1,5 @@
 COCOINDEX_DB=./cocoindex.db
+POSTGRES_URL=postgres://cocoindex:cocoindex@localhost/cocoindex
+
+# Fallback to CPU for operations not supported by MPS on Mac.
+PYTORCH_ENABLE_MPS_FALLBACK=1


examples/entire_session_search/.env
```diff
@@ -0,0 +1,6 @@
+COCOINDEX_DB=postgres://cocoindex:cocoindex@localhost/cocoindex
```
georgeh0: We no longer use Postgres for CocoIndex internal database in v1. This needs to be a local file, e.g. `./cocoindex.db`

examples/entire_session_search/main.py
```diff
@@ -0,0 +1,376 @@
...
+    emb_table: postgres.TableTarget[SessionEmbeddingRow],
+    meta_table: postgres.TableTarget[SessionMetadataRow],
+) -> None:
+    info = extract_session_info(file)
+    filename = file.file_path.path.name
+
+    if filename == "full.jsonl":
+        content = file.read_text()
+        chunks = parse_transcript(content)
+        id_gen = IdGenerator()
```
georgeh0: Consider creating `IdGenerator()` outside all branches (before the `if`).

examples/entire_session_search/main.py
```diff
@@ -0,0 +1,376 @@
...
+
+        while True:
+            q = input("Enter search query (or Enter to quit): ").strip()
+            if not q:
+                break
+            await query_once(pool, q)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1 and sys.argv[1] == "query":
```
georgeh0: Actually we don't need users to provide `query` as `argv[1]`, since it only supports query when running as main.

examples/entire_session_search/.env
Haleshot: Done, matched the v1 storage approach w/ `./cocoindex.db`.

examples/entire_session_search/main.py
Haleshot: Removed the query subcommand; `python main.py` now drops straight into the search (with _optional_ inline args like python main.py "search term"). This felt more consistent.

Does this work?

Haleshot:
Ah, I meant to add a video walkthrough of this example in the PR description. I think I'll create an issue in the [`blog-proposal`](https://github.com/cocoindex-io/blog-proposal) repo now and add it there.

Thanks for the review :)

examples/entire_session_search/main.py
```diff
@@ -0,0 +1,373 @@
...
+    """Extract checkpoint_id and session_index from file path.
+
+    Entire layout: <shard>/<checkpoint_id>/<session_idx>/<filename>
+    """
+    parts = file.file_path.path.parts
+    # parts[-1] is the filename, [-2] is session_index, [-3] is checkpoint_id
+    return SessionInfo(
+        checkpoint_id=parts[-3],
+        session_index=parts[-2],
+    )
```
Soph, line 111: This is not correct. The checkpoint id is sharded yes, but the `<shard>` is part of the id, so the correct code would be: 

```  
  return SessionInfo(
      checkpoint_id=parts[-4] + parts[-3],  # "a3" + "b2c4d5e6f7" = "a3b2c4d5e6f7"
      session_index=parts[-2],
  )
```

examples/entire_session_search/main.py
```diff
@@ -0,0 +1,373 @@
...
+        meta_table.declare_row(
+            row=SessionMetadataRow(
+                checkpoint_id=info.checkpoint_id,
+                session_index=info.session_index,
+                prompt_summary=prompt_summary,
+                total_tokens=total_tokens,
+                files_touched=files_touched_str,
+                agent_percentage=agent_pct,
+            ),
+        )
```
Soph, line 270:
```suggestion
    elif filename == "metadata.json":
        content = file.read_text()
        try:
            meta = json.loads(content)
        except json.JSONDecodeError:
            return
  
        # Use summary.intent if available, otherwise empty
        summary = meta.get("summary", {})
        prompt_summary = summary.get("intent", "")
  
        # Token usage is nested
        token_usage = meta.get("token_usage", {})
        input_t = token_usage.get("input_tokens", 0) or 0
        output_t = token_usage.get("output_tokens", 0) or 0
        total_tokens = input_t + output_t
  
        # Files touched is correct
        files_touched = meta.get("files_touched", [])
  
        # Agent percentage is nested in initial_attribution
        initial_attr = meta.get("initial_attribution", {})
        agent_pct = initial_attr.get("agent_percentage")
```

I think this is the more correct parsing of the values.

But also a caveat: If the agent or the user decides to split the outcome of a single prompt into multiple commits the current design will create a checkpoint for each commit but the content of the checkpoint will be basically the same for each. The reason for that is, that we want to make it easy to go from a commit to the relevant session / data and then do deduplication later if needed. So if you look at 2 commits coming out of a single prompt and then just add up tokens you get twice the tokens.

examples/entire_session_search/README.md
```diff
@@ -0,0 +1,98 @@
...
+    parse --> embed[SentenceTransformer]
+    embed_prompt --> embed
+    split --> embed
+    embed --> emb_table[(session_embeddings)]
+```
+
+## Entire checkpoint layout
+
+```
+<shard>/<checkpoint_id>/<session_idx>/
```
Soph, line 92:
```suggestion
<checkpoint_id[:2]>/<checkpoint_id[2:]>/<session_idx>/
```

examples/entire_session_search/main.py
```diff
@@ -0,0 +1,373 @@
...
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def extract_session_info(file: FileLike) -> SessionInfo:
+    """Extract checkpoint_id and session_index from file path.
+
+    Entire layout: <shard>/<checkpoint_id>/<session_idx>/<filename>
```
Soph, line 104:
```suggestion
    Entire layout: <checkpoint_id[:2]>/<checkpoint_id[2:]>/<session_idx>/
```

examples/entire_session_search/main.py
Haleshot: Thanks for the catch; I had read the sharding as a separate directory prefix rather than the first 2 chars of the ID itself.

_Fixed as suggested_

examples/entire_session_search/main.py
Haleshot: Good to know _re:_ the multi-commit approach.

In this example `total_tokens` is stored per checkpoint and only used for display / filtering on individual sessions, so there's no _cross-checkpoint aggregation_ happening (doesn't affect us here?); worth adding a note in the README though...

Haleshot:
Thanks for reviewing after merge, appreciate it! Drawing up a _follow-up_ PR with the changes suggested soon.

badmonster0:
@Soph thanks a lot for the review,  @Haleshot thanks a lot for the follow up, looking forward to it!

PR #1677, opened 2026-02-17
Feat: Support dataclass and pydantic models for memo keys

### LINKED ISSUE #1557: [FEATURE] Support dataclass and pydantic models for memo keys
It's about [memo key](https://cocoindex.io/docs-v1/advanced_topics/memoization_keys) computation.

We want to support dataclass and pydantic models in addition to list/dict/etc. ([existing code](https://github.com/cocoindex-io/cocoindex/blob/76eecb79f01c4fcba32f1119b126333f4d0ba832/python/cocoindex/_internal/memo_key.py#L116-L133))

---

prithvi-moonshot:
- Auto-canonicalize dataclass instances using field definitions in order
- Auto-canonicalize Pydantic v2 models via __pydantic_fields__ (includes unset fields)
- Type tags prevent collisions between different dataclass/Pydantic types
- __coco_memo_key__() still takes precedence for custom behavior
- Tests added

Closes #1557

### DIFF
--- a/python/cocoindex/_internal/core.pyi
+++ b/python/cocoindex/_internal/core.pyi
@@ -104,7 +104,9 @@ class FnCallMemoGuard:
     @property
     def cached_memo_states(self) -> list[Any] | None: ...
     def update_memo_states(self, memo_states: list[Any]) -> None: ...
-    def resolve(self, fn_ctx: FnCallContext, ret: Any, memo_states: list[Any] | None = None) -> bool: ...
+    def resolve(
+        self, fn_ctx: FnCallContext, ret: Any, memo_states: list[Any] | None = None
+    ) -> bool: ...
     def close(self) -> None: ...
 
 # --- ComponentMountHandle ---
@@ -139,7 +141,6 @@ class StablePathInfo:
     path: StablePath
     node_type: StablePathNodeType
 
-
 class StablePathInfoAsyncIterator:
     """Async iterator of StablePathInfo; use with async for."""
 
@@ -148,7 +149,9 @@ class StablePathInfoAsyncIterator:
 
 # --- App ---
 class App:
-    def __new__(cls, name: str, env: Environment, max_inflight_components: int | None = None) -> App: ...
+    def __new__(
+        cls, name: str, env: Environment, max_inflight_components: int | None = None
+    ) -> App: ...
     def update(
         self,
         root_processor: ComponentProcessor[T_co],
--- a/python/cocoindex/_internal/setting.py
+++ b/python/cocoindex/_internal/setting.py
@@ -58,6 +58,8 @@ class Settings:
 
     db_path: os.PathLike[str] | None = None
     global_execution_options: GlobalExecutionOptions | None = None
+    lmdb_max_dbs: int = 1024
+    lmdb_map_size: int = 0x1_0000_0000  # 4 GiB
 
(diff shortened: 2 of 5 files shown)


python/cocoindex/_internal/memo_key.py
```diff
@@ -136,7 +179,15 @@ def _canonicalize(obj: object, _seen: dict[int, int] | None) -> Fingerprintable:
...
             return ("hook", base.__module__, base.__qualname__, _canonicalize(k, _seen))
 
-    # 3) Cycle / shared-reference tracking
+    # 3) Dataclass instances (check before containers since dataclasses may be sequences)
+    if _is_dataclass_instance(obj):
+        return _canonicalize_dataclass(obj)
+
+    # 4) Pydantic v2 models (check before containers since they may be sequences)
+    if _is_pydantic_model(obj):
+        return _canonicalize_pydantic(obj)
```
georgeh0: I think this should be after "cycle / shared-reference tracking", since there can also be cycles within dataclasses / pydantic models.

Preferably also after "Containers", since they seem to be less common and we can let containers check happen earlier.

python/cocoindex/_internal/memo_key.py
```diff
@@ -23,6 +24,48 @@
...
+    Format: ("dataclass", module, qualname, ((field_name, value), ...))
+    """
+    typ = type(obj)
+    # Get fields in definition order
+    fields = dataclasses.fields(obj)  # type: ignore[arg-type]
+    field_items: list[tuple[str, Fingerprintable]] = []
+    for field in fields:
+        value = getattr(obj, field.name)
+        field_items.append((field.name, _canonicalize(value, _seen=None)))
+    return ("dataclass", typ.__module__, typ.__qualname__, tuple(field_items))
```
georgeh0: nit: We can use a comprehension to create the tuple from the generator:

```python
tuple(
  (field.name, _canonicalize(getattr(obj, field.name), _seen=None))
  for field in fields
)
```

This saves an intermediate list.

Same for  `_canonicalize_pydantic()` below

python/cocoindex/_internal/memo_key.py
```diff
@@ -23,6 +24,48 @@
...
+
+def _canonicalize_pydantic(obj: object) -> Fingerprintable:
+    """Canonicalize a Pydantic v2 model instance.
+
+    Includes all fields (set and unset) to ensure determinism.
+    Format: ("pydantic", module, qualname, ((field_name, value), ...))
+    """
+    typ = type(obj)
+    # Get all field names from model fields
+    field_names = list(obj.__pydantic_fields__.keys())  # type: ignore[attr-defined]
```
georgeh0: nit: we don't need to convert `obj.__pydantic_fields__.keys()` to `list` first. We can directly iterate over it.

python/cocoindex/_internal/memo_key.py
prithvi-moonshot: Done

python/cocoindex/_internal/memo_key.py
prithvi-moonshot: Done

python/cocoindex/_internal/memo_key.py
prithvi-moonshot: Moved

Issue #1685, opened 2026-02-19
[BUG] `test_cli.py::TestShowTree` is flaky: Fatal Python error: PyGILState_Release


georgeh0:
The test `python/tests/cli/test_cli.py::TestShowTree::test_show_tree_with_nested_structure` is flaky. It fails non-deterministically on GitHub runner. It happened twice today, both on Ubuntu-24.04-arm on Python 3.11.

- [Failure 1](https://github.com/cocoindex-io/cocoindex/actions/runs/22194484188/job/64191068207?pr=1682) 
- [Failure 2](https://github.com/cocoindex-io/cocoindex/actions/runs/22201218344/job/64214333830?pr=1684)

The first failure has more information in output:

```
  FAILED python/tests/cli/test_cli.py::TestShowTree::test_show_tree_with_nested_structure - AssertionError: Command failed: ['cocoindex', 'show', './tree_test_app.py', '--tree']
  returncode=-6
  stdout:
  Found 6 stable paths:
  / [component]
  ├── direct [component]
  ├── files
  │   ├── file1.txt [component]
  │   └── file2.txt [component]
  └── setup [component]

  stderr:
  Fatal Python error: PyGILState_Release: thread state 0xff171c001d00 must be current when releasing
  Python runtime state: finalizing (tstate=0x0000ff184ea06480)

  Thread 0x0000ff184ea7a020 (most recent call first):
    <no Python frame>

  Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg (total: 2)
```

The message "Fatal Python error: PyGILState_Release: thread state 0xff171c001d00 must be current when releasing" suggests somewhere we're trying to release GIL when we don't have it? Maybe some race conditions?

After retries, both succeeded.

georgeh0:
@shannon06437 , do you want to have a look? Please let me know if you're stuck. Thanks!

shannon06437:
I'll take a look!

georgeh0:
Hi @shannon06437, the flakiness problem comes back: https://github.com/cocoindex-io/cocoindex/actions/runs/22366810229/job/64734899559

Would you have a look? Thanks!

georgeh0:
Hi @shannon06437 , I encountered one more case of flakiness on `PyGILState_Release`

https://github.com/cocoindex-io/cocoindex/actions/runs/22738523646/job/65946038372

Would you please have a look? Thanks!

georgeh0:
[...]

PR #1687, opened 2026-02-20
fix: test_cli.py::TestShowTree is flaky

shannon06437:
Fix for the bug at https://github.com/cocoindex-io/cocoindex/issues/1685 

Add clean up for environments and their background threads before the subprocess exits, which was causing a race condition where threads tried to call PyGILState_Release after CPython had torn down thread state resulting in SIGSEGV (-11) or SIGABRT (-6) crash.

### DIFF
--- a/python/cocoindex/cli.py
+++ b/python/cocoindex/cli.py
@@ -640,23 +640,27 @@ def show(app_target: str, tree: bool) -> None:
     """
     app = _load_app(app_target)
 
-    if tree:
-        items = list_stable_paths_info_sync(app)
-        component_node_type = _core.StablePathNodeType.component()
-        paths: list[StablePath] = []
-        component_paths: set[StablePath] = set()
-        for item in items:
-            path = StablePath(item.path)
-            paths.append(path)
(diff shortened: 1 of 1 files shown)


python/cocoindex/cli.py
```diff
@@ -640,23 +640,27 @@ def show(app_target: str, tree: bool) -> None:
...
-        output = _render_paths_as_tree(paths, component_paths)
-        click.echo(output)
-    else:
-        paths = list_stable_paths_sync(app)
-        click.echo(f"Found {len(paths)} stable paths:")
-        for path in paths:
-            click.echo(f"  {path}")
+    try:
+        if tree:
+            items = list_stable_paths_info_sync(app)
```
georgeh0, line 645: (This is missed from the previous PR review)

This part should also be streaming, i.e. instead of buffering the whole thing in memory, we should use an async iterator and print out things as soon as each arrives.

Since we already use `asyncio.run_coroutine_threadsafe` below, consider encapsulating the whole thing within an async function.

PR #1702, opened 2026-02-24
fix: adding streaming to cli.py for show --tree command

shannon06437:
Fixes streaming issue as mentioned here: https://github.com/cocoindex-io/cocoindex/pull/1687

- Updated cocoindex show to be fully streaming (prints paths as they arrive)
- Updated cocoindex show --tree to a streaming, indented bullet list format (no buffering) and adjusted CLI tests accordingly


shannon06437:
Please take a look!

Issue #1703, opened 2026-02-24
[FEATURE] Make LMDB max dbs and map size configurable


georgeh0:
Currently we have LMDB max DBs and map size hardcoded:

https://github.com/cocoindex-io/cocoindex/blob/4a3679b9decbf3f26d6a9878b2f94e07f30c075b/rust/core/src/engine/environment.rs#L8-L10

We want to expose them as configurable env settings, and also environment variables.

Existing settings as reference: [`COCOINDEX_DB`](https://github.com/cocoindex-io/cocoindex/blob/4a3679b9decbf3f26d6a9878b2f94e07f30c075b/python/cocoindex/_internal/setting.py#L19)

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

Harry-kp:
I'd like to work on this. Will send a PR shortly.

badmonster0:
Thanks a lot!! assigned!

Haleshot:
Can be closed via https://github.com/cocoindex-io/cocoindex/pull/1704?

georgeh0:
> Can be closed via [#1704](https://github.com/cocoindex-io/cocoindex/pull/1704)?

Yes, thanks! Closed.

PR #1704, opened 2026-02-24
feat: make LMDB max_dbs and map_size configurable

Harry-kp:
## Summary

- Exposes the hardcoded LMDB `max_dbs` (1024) and `map_size` (4 GiB) as configurable settings
- Configurable via Python `Settings(lmdb_max_dbs=..., lmdb_map_size=...)` or environment variables `COCOINDEX_LMDB_MAX_DBS` / `COCOINDEX_LMDB_MAP_SIZE`
- Defaults are set on the Python side (matching the `source_max_inflight_rows` pattern), so fields are always concrete values
- Invalid values are validated on the Rust side with `client_bail!`

**Fixes** #1703

## Changes

- `rust/core/src/engine/environment.rs` — Added `lmdb_max_dbs: u32` and `lmdb_map_size: usize` to `EnvironmentSettings` with `#[serde(default = "...")]` for backward compatibility. Validates `max_dbs >= 1` and `map_size > 0` via `client_bail!`. Removed unused `u32` import and the `// TODO: Expose these as settings` comment.
- `python/cocoindex/_internal/setting.py` — Added `lmdb_max_dbs: int = 1024` and `lmdb_map_size: int = 0x1_0000_0000` to `Settings`, loaded from env vars in `from_env()`
- `python/cocoindex/setting.py` — Same changes (public API mirror, noted as deprecated in v1)

## Usage

```python
import cocoindex

settings = cocoindex.Settings(
    db_path="/tmp/cocoindex_db",
    lmdb_max_dbs=2048,
    lmdb_map_size=0x2_0000_0000,  # 8 GiB
)
```

Or via environment variables:
```bash
export COCOINDEX_LMDB_MAX_DBS=2048
export COCOINDEX_LMDB_MAP_SIZE=8589934592
```

### DIFF
--- a/python/cocoindex/_internal/setting.py
+++ b/python/cocoindex/_internal/setting.py
@@ -58,8 +58,8 @@ class Settings:
 
     db_path: os.PathLike[str] | None = None
     global_execution_options: GlobalExecutionOptions | None = None
-    lmdb_max_dbs: int | None = None
-    lmdb_map_size: int | None = None
+    lmdb_max_dbs: int = 1024
+    lmdb_map_size: int = 0x1_0000_0000  # 4 GiB
 
     @classmethod
     def from_env(cls, db_path: os.PathLike[str] | None = None) -> Self:
--- a/python/cocoindex/setting.py
+++ b/python/cocoindex/setting.py
@@ -46,8 +46,8 @@ class Settings:
 
     db_path: os.PathLike[str] | None = None
     global_execution_options: GlobalExecutionOptions | None = None
-    lmdb_max_dbs: int | None = None
-    lmdb_map_size: int | None = None
+    lmdb_max_dbs: int = 1024
+    lmdb_map_size: int = 0x1_0000_0000  # 4 GiB
 
     @classmethod
     def from_env(cls, db_path: os.PathLike[str] | None = None) -> Self:
--- a/rust/core/src/engine/environment.rs
+++ b/rust/core/src/engine/environment.rs
@@ -3,18 +3,25 @@ use crate::{
 };
 
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeSet, path::PathBuf, u32};
+use std::{collections::BTreeSet, path::PathBuf};
 
 const DEFAULT_MAX_DBS: u32 = 1024;
 const DEFAULT_LMDB_MAP_SIZE: usize = 0x1_0000_0000; // 4GiB
 
+fn default_max_dbs() -> u32 {
+    DEFAULT_MAX_DBS
+}
+fn default_lmdb_map_size() -> usize {
+    DEFAULT_LMDB_MAP_SIZE
+}
+
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct EnvironmentSettings {
     pub db_path: PathBuf,
-    #[serde(default)]
-    pub lmdb_max_dbs: Option<u32>,
-    #[serde(default)]
-    pub lmdb_map_size: Option<usize>,
+    #[serde(default = "default_max_dbs")]
+    pub lmdb_max_dbs: u32,
+    #[serde(default = "default_lmdb_map_size")]
+    pub lmdb_map_size: usize,
 }
 
 struct EnvironmentInner<Prof: EngineProfile> {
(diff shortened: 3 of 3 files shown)


python/cocoindex/_internal/setting.py
```diff
@@ -58,6 +58,8 @@ class Settings:
 
     db_path: os.PathLike[str] | None = None
     global_execution_options: GlobalExecutionOptions | None = None
+    lmdb_max_dbs: int | None = None
+    lmdb_map_size: int | None = None
```
georgeh0: I think we can put the default values at Python side (similar to `GlobalExecutionOptions.source_max_inflight_rows`). This will make the default more clear from Python API. Then these fields are always not `None` and we don't need to apply default by constants from Rust side.

python/cocoindex/setting.py
georgeh0, line 1: FYI, this file is no longer used in v1 (I'll do a cleanup soon). It's OK to either keep or revert current changes.

rust/core/src/engine/environment.rs
```diff
@@ -37,10 +40,12 @@ impl<Prof: EngineProfile> Environment<Prof> {
         std::fs::create_dir_all(&db_path)?;
         // Backward compatibility: migrate LMDB files from old layout into mdb/.
         Self::migrate_legacy_db_files(&settings.db_path, &db_path)?;
+        let max_dbs = settings.lmdb_max_dbs.unwrap_or(DEFAULT_MAX_DBS);
+        let map_size = settings.lmdb_map_size.unwrap_or(DEFAULT_LMDB_MAP_SIZE);
```
georgeh0: To resolve the copilot comments on the `setting.py`, we can validate these values are positive here and raise a error otherwise (use `client_bail!`)

python/cocoindex/_internal/setting.py
Harry-kp: Done in f2308d6d — moved defaults to the Python side, matching the `source_max_inflight_rows` pattern. Fields are now always concrete `int` values, never `None`. Removed `Option<>` wrappers on the Rust side as well, with `#[serde(default = "...")]` for backward compatibility.

rust/core/src/engine/environment.rs
Harry-kp: Done in f2308d6d — added validation with `client_bail!` before calling `heed::EnvOpenOptions`:

```rust
if settings.lmdb_max_dbs < 1 {
    client_bail!("lmdb_max_dbs must be >= 1, got {}", settings.lmdb_max_dbs);
}
if settings.lmdb_map_size == 0 {
    client_bail!("lmdb_map_size must be > 0, got {}", settings.lmdb_map_size);
}
```

This also resolves the Copilot comments about Python-side validation — invalid values are now caught at the Rust layer with clear error messages.

python/cocoindex/setting.py
Harry-kp: Thanks for the heads up! I have kept the changes for consistency with `_internal/setting.py`. Happy to revert if you prefer — just let me know.

rust/core/src/engine/environment.rs
```diff
@@ -5,13 +5,16 @@ use crate::{
 use serde::{Deserialize, Serialize};
 use std::{collections::BTreeSet, path::PathBuf, u32};
```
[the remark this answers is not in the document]
Harry-kp: Fixed in f2308d6d.

rust/core/src/engine/environment.rs
```diff
@@ -37,10 +40,12 @@ impl<Prof: EngineProfile> Environment<Prof> {
...
         // Backward compatibility: migrate LMDB files from old layout into mdb/.
         Self::migrate_legacy_db_files(&settings.db_path, &db_path)?;
+        let max_dbs = settings.lmdb_max_dbs.unwrap_or(DEFAULT_MAX_DBS);
+        let map_size = settings.lmdb_map_size.unwrap_or(DEFAULT_LMDB_MAP_SIZE);
         let db_env = unsafe {
             heed::EnvOpenOptions::new()
-                .max_dbs(MAX_DBS)
-                .map_size(LMDB_MAP_SIZE)
+                .max_dbs(max_dbs)
+                .map_size(map_size)
```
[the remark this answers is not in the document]
Harry-kp: Addressed — validation is now handled on the Rust side with `client_bail!` per @georgeh0's guidance.

python/cocoindex/setting.py
```diff
@@ -66,9 +68,24 @@ def from_env(cls, db_path: os.PathLike[str] | None = None) -> Self:
...
+            "COCOINDEX_LMDB_MAX_DBS",
+            parse=int,
+        )
+        _load_field(
+            lmdb_kwargs,
+            "lmdb_map_size",
+            "COCOINDEX_LMDB_MAP_SIZE",
+            parse=int,
+        )
+
```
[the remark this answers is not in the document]
Harry-kp: Validation is now handled on the Rust side with `client_bail!` as suggested by @georgeh0, so Python-side validation is not needed.

python/cocoindex/setting.py
```diff
@@ -66,9 +68,24 @@ def from_env(cls, db_path: os.PathLike[str] | None = None) -> Self:
...
+            "lmdb_max_dbs",
+            "COCOINDEX_LMDB_MAX_DBS",
+            parse=int,
+        )
+        _load_field(
+            lmdb_kwargs,
+            "lmdb_map_size",
+            "COCOINDEX_LMDB_MAP_SIZE",
+            parse=int,
+        )
```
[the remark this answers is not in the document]
Harry-kp: @georgeh0 The existing `GlobalExecutionOptions.source_max_inflight_rows` (the pattern we're following) doesn't have dedicated unit tests for env var loading either. Would you like me to add test coverage for the LMDB env vars, or is the Rust-side validation sufficient for now?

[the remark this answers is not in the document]
georgeh0: It's good for now. Thanks!

PR #1705, opened 2026-02-25
Fix: flaky tree test

shannon06437:
Fixes bug at: https://github.com/cocoindex-io/cocoindex/issues/1685

- Stabilized CLI async behavior and env shutdown so cocoindex commands no longer rely on the shared background event loop, avoiding cross-thread asyncio issues (especially on Python 3.14/free-threaded builds)
- While running CI pipeline, found errors in full-reprocess test with non-deterministic behaviour so also fixed full-reprocess semantics in the Rust core so function memoization respects full_reprocess=True, and made the test harness’ metrics collection thread-safe to eliminate flakiness under parallel execution


shannon06437:
Please take a look!

rust/core/src/engine/function.rs
```diff
@@ -72,9 +72,12 @@ pub async fn reserve_memoization<Prof: EngineProfile>(
...
             if let FnCallMemoEntry::Pending = &*guard {
-                let stored_fn_call_memo = read_fn_call_memo(comp_exec_ctx, memo_fp)?;
-                if let Some(fn_call_memo) = stored_fn_call_memo {
-                    *guard = FnCallMemoEntry::Ready(Some(fn_call_memo));
+                // Under full_reprocess, force execution (do not load cached memo).
+                if !comp_exec_ctx.full_reprocess() {
+                    let stored_fn_call_memo = read_fn_call_memo(comp_exec_ctx, memo_fp)?;
+                    if let Some(fn_call_memo) = stored_fn_call_memo {
+                        *guard = FnCallMemoEntry::Ready(Some(fn_call_memo));
+                    }
```
georgeh0, line 80: Nice catch!

Is it possible to add a test case to cover fixed problem?

georgeh0 (approved):
Thanks for the quick fix!

georgeh0:
I'll merge first, since it's fixing real flaky issue :)

PR #1735, opened 2026-03-07
Fix: Flaky cli tests

shannon06437:
Fixes flaky cli tests as per most recent comment: https://github.com/cocoindex-io/cocoindex/issues/1685, [Ubuntu 3.11 build error](https://github.com/cocoindex-io/cocoindex/actions/runs/22738523646/job/65946038372)

- Explicitly call`drop(processor)` and `drop(context)` before calling `guard.resolve()` so that all Python objects are released while the interpreter is still running and there's nothing to clean up once Python starts shutting down
- ManuallyDrop suppresses the destructor so the raw Python pointers are never decremented at process exit to cause SIGSEGV

After the first fix, a timeout error occurred: [Windows 3.14 build error](https://github.com/shannon06437/cocoindex/actions/runs/22752692099/job/65990926494), so a second fix was made:

After the 1-second join, unconditionally call `loop.close()` regardless of thread state. Closing the Windows ProactorEventLoop IOCP handle unblocks `GetQueuedCompletionStatus` and allows the daemon thread to exit. A second `thread.join(timeout=1.0)` waits for the clean exit before returning. Previously, `loop.call_soon_threadsafe(loop.stop)` did not reliably wake up `GetQueuedCompletionStatus`.

### DIFF
--- a/rust/core/src/engine/component.rs
+++ b/rust/core/src/engine/component.rs
@@ -352,6 +352,9 @@ impl<Prof: EngineProfile> Component<Prof> {
                     Ok((outcome, output)) => (outcome, Ok(output)),
                     Err(err) => (ComponentRunOutcome::exception(), Err(err)),
                 };
+                drop(processor);
+                drop(context);
+                drop(self);
                 child_readiness_guard.map(|guard| guard.resolve(outcome));
                 output?
                     .ok_or_else(|| internal_error!("component deletion can only run in background"))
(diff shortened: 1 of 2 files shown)


shannon06437:
Please take a look!

rust/core/src/engine/component.rs
```diff
@@ -352,6 +352,9 @@ impl<Prof: EngineProfile> Component<Prof> {
                     Ok((outcome, output)) => (outcome, Ok(output)),
                     Err(err) => (ComponentRunOutcome::exception(), Err(err)),
                 };
+                drop(processor);
+                drop(context);
+                drop(self);
```
georgeh0, line 357: Good catch!

PR #1747, opened 2026-03-10
Fix: flaky cli tests

shannon06437:
Fixing issue: https://github.com/cocoindex-io/cocoindex/issues/1685

Fix 1: pyo3 POOL destroyed before statics (TOKIO_RUNTIME and ROOT_TARGET_STATE_PROVIDER_REGISTRY) that hold Py<PyAny> objects
Add ManuallyDrop for both

Fix 2: Tokio threads calling PyGILState_Release after _PyGILState_Fini
Added shutdown_runtime() — uses ptr::read to extract the Runtime from its ManuallyDrop wrapper and calls shutdown_timeout(5s) to drain tasks and join all threads


shannon06437:
Please take a look!

python/cocoindex/_internal/environment.py
```diff
@@ -519,7 +519,29 @@ def _shutdown_background_loop_at_exit() -> None:
...
+    can race with the runtime shutdown.
+    """
+    try:
+        core.shutdown_tokio_runtime()
+    except Exception:
+        pass
+
+
 atexit.register(_shutdown_background_loop_at_exit)
+atexit.register(_shutdown_tokio_runtime_at_exit)
```
georgeh0: I think we only need to register one hook, to shutdown both.

georgeh0 (commented):
Thanks for debugging and fixing!

shannon06437:
Please take a look!

Issue #1560, opened 2026-01-25
[FEATURE] Support registration of exception handler for mounted components


georgeh0:
When a component is mounted by [`mount()`](https://cocoindex.io/docs-v1/programming_guide/processing_component#mount-1), execution happens in background. When there's an exception, currently we [log the exception](https://github.com/cocoindex-io/cocoindex-v1/blob/1809328f33e62d1a893319d6f6bdac9f72344e11/rust/core/src/engine/component.rs#L338-L341) instead of failing the parent component, so the exceptions in processing a single input (file, row) won't affect others.

(note: this is `mount()`'s behavior; `mount_run()` will populate the error in the `result()` call, since `mount_run()` means parent has a dependency on the child component's processing)

Sometimes users want to be able to monitor and handle these exceptions in their own way. We should allow users to register exception handlers.

- There can be exception handler registered globally (i.e. at the environment level). Registration can happen in the lifespan function. Logging can be behavior of the default exception handler if users don't explicitly register one.
- Users can setup a more specific exception handler in a specific `Scope`, which creates a new `Scope`. e.g. `scope.with_exception_handler(specific_handler)`. In this case, the `specific_handler` will be invoked first on any exception mounted with this scope and any sub scope, but if `specific_handler` itself raises an exception, the previous exception handler will be invoked on this exception raised by `specific_handler`. 

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

prabhath004:
@georgeh0 @badmonster0  is this still open I can contribute

badmonster0:
hi @prabhath004 https://github.com/cocoindex-io/cocoindex/pull/1754 there is a PR in progress on this! Really appreciate your interest in helping, if there's other issues you are interested to take please let us know! Cc @shannon06437

shannon06437:
Just pushed an update!

PR #1754, opened 2026-03-12
Feat: Mount exception handler

shannon06437:
Resolves this issue: https://github.com/cocoindex-io/cocoindex/issues/1560

Summary:
- Added environment and scope-level exception handlers for background mount() failures.
- Wired handler chaining so inner handlers run first and fall back to outer/global handlers on failure.

Changes:

- Extended ComponentContext with _exception_handlers, ExceptionContext, ExceptionHandler, and coco.exception_handler(...).
- Added EnvironmentBuilder.set_exception_handler(...) and plumbed the handler into Environment and the root ComponentContext.
- Updated mount() and mount_each() to call a new core.mount_async_with_handler(...) when handlers are present.
- Modified Rust Component::run_in_background(...) and the PyO3 layer to accept and invoke an async on_error callback instead of always logging.
- Added test_exception_handlers.py to verify behaviour.


shannon06437:
Please take a look!

python/cocoindex/_internal/environment.py
```diff
@@ -88,10 +89,12 @@ class EnvironmentBuilder:
 
     _settings: setting.Settings
     _context_provider: ContextProvider
+    _exception_handler: "ExceptionHandler | None"
```
georgeh0: nit: since we already have `from __future__ import annotations`, the quotes around `ExceptionHandler | None` is not necessary

rust/py/src/component.rs
```diff
@@ -260,7 +260,65 @@ pub fn mount_async<'py>(
...
+
+#[pyfunction]
+pub fn mount_async_with_handler<'py>(
+    py: Python<'py>,
+    processor: PyComponentProcessor,
+    stable_path: PyStablePath,
+    comp_ctx: PyComponentProcessorContext,
+    fn_ctx: &PyFnCallContext,
+    handler_callback: Py<PyAny>,
+) -> PyResult<Bound<'py, PyAny>> {
```
georgeh0: We may merge this one with `mount_async()` into one API, and take `handler_callback` as an optional argument. Because the main behaviors of these two are identical. This helps avoid exponential explosion after we have more options like this.

python/cocoindex/_internal/component_ctx.py
```diff
@@ -53,9 +82,19 @@ def _with_extended_path(self, *parts: StableKey) -> ComponentContext:
...
+            self._exception_handlers,
+        )
+
+    def _with_exception_handler(self, handler: ExceptionHandler) -> ComponentContext:
+        return ComponentContext(
+            self._env,
+            self._core_path,
+            self._core_processor_ctx,
+            self._core_fn_call_ctx,
+            (*self._exception_handlers, handler),
```
georgeh0: (optional) There can be a slight more efficient way to represent it, something like this (similar to a list in Lisp):

```python
class ExceptionHandlerChain(NamedTuple):
  handler: ExceptionHandler
  base: ExceptionHandlerChain | None = None
```

With this, we don't need to create a new tuple with size N each time. 

And it's also slightly easier to write the resolve logic: just doing a tail recursion.

shannon06437:
Please take a look!

georgeh0:
PR looks good! Can you also help to update the documentation? Consider add a new document under [`advanced_topics`](https://github.com/cocoindex-io/cocoindex/tree/v1/docs/docs/advanced_topics). Thanks!

shannon06437:
Please take a look!

georgeh0 (approved):
Thanks!!

Issue #1744, opened 2026-03-09
[FEATURE] Batched functions should support additional arguments


georgeh0:
Currently, in CocoIndex v1, a [batched function](https://cocoindex.io/docs-v1/programming_guide/function#batching) takes exactly one argument. e.g.

```python
@coco.fn.as_async(batching=True, runner=coco.GPU, max_batch_size=64)
def embed(self, texts: list[str]) -> list[_NDArray[_np.float32]]:
  ...
```

It exposes a method with this signature:

```python
async def embed(self, text: str) -> _NDArray[_np.float32]
  ...
```

However, sometimes more arguments are needed. e.g. when calling `embed()`, users may also provide arguments like `query_prompt_name`. So we need a method signature like this:

```python
async def embed(self, text: str, query_prompt_name: str) -> _NDArray[_np.float32]
  ...
```

The then the batching behavior should be: for all calls on `embed()` that matches `self` + `query_prompt_name`, they can be batched together. i.e. the batched function should take this signature:

```python
@coco.fn.as_async(batching=True, runner=coco.GPU, max_batch_size=64)
def embed(self, texts: list[str], query_prompt_name: str) -> list[_NDArray[_np.float32]]:
  ...

```

We want to update the batching support to allow this

Note:
- The current [`sentence_transformers.py`](https://github.com/cocoindex-io/cocoindex/blob/69dba970e5c25463c83e17b14284767399172f03/python/cocoindex/ops/sentence_transformers.py#L124-L148) is manually doing this by manually maintain a dict to group by additional arguments. After this issue is resolved, we can save the boilerplate and do it automatically.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

PR #1766, opened 2026-03-15
feat: Batched functions support additional arguments

Davda-James:
Added support for extra args to pass in embed and updated the logic to handle that on each call embed as well removed from sentence_transformer.py manual grouping logic.
Also added 2 tests for the change.

Closes Issue: #1744 

Pls review it whenever you get a chance.


badmonster0:
thanks a lot @Davda-James for the PR!!
@georgeh0 george can help take a look!

python/cocoindex/_internal/function.py
```diff
@@ -878,42 +888,68 @@ async def runner_batch_fn_async(inputs: list[Any]) -> list[R_co]:
 
         # User function is a batch function: list[T] -> list[R]
         if self_obj is None:
-            return self._any_fn  # type: ignore
+            if not extra_args and not extra_kwargs:
+                return self._any_fn  # type: ignore
+            fn = self._any_fn
+            if inspect.iscoroutinefunction(fn):
```
georgeh0: I think here we can use the same condition as the branch for `self_obj` non-None case (`(orig_async_fn := self._orig_async_fn) is not None`), and use `orig_async_fn` / `orig_sync_fn` in two branches like below.

python/tests/core/test_function_batching.py
```diff
@@ -845,6 +845,33 @@ async def test_gpu_runner_lazy_env_var(
     assert coco.GPU._should_use_subprocess() is True  # still cached
 
 
+@coco.fn.as_async(batching=True)  # type: ignore[arg-type]
+def batched_with_extra_arg(inputs: list[str], prefix: str) -> list[str]:
```
georgeh0, line 849: Can we also add additional kwargs to the test?

And it'll also be great if we add test for a member method (with `self`) with extra args and kwargs.

Thanks!

georgeh0 (commented):
Sorry for the late review. The PR is great! A few minor comments.

python/tests/core/test_function_batching.py
Davda-James: done

python/cocoindex/_internal/function.py
Davda-James: done

georgeh0 (approved):

georgeh0:
Thanks!!

PR #1798, opened 2026-04-03
fix(postgres): preserve NUL bytes in internal JSONB tracking state

### LINKED ISSUE #933: [BUG] Make U+0000 (NUL) works with Postgres
U+0000 (NUL) is valid in Unicode, but unfortunately it's not supported by Postgres ([Postgres documentation](https://www.postgresql.org/docs/current/datatype-character.html#:~:text=the%20character%20with%20code%20zero%20(sometimes%20called%20NUL)%20cannot%20be%20stored)).
Sometimes the input text has U+0000 (usually test cases to intentionally contain control characters, [example](https://github.com/BurntSushi/toml/blob/master/internal/toml-test/tests/invalid/control/multi-null.toml)), and the flow will fail when writing into Postgres.

Our plan:
- For Postgres target, sanitize and strip U+0000, document the behavior clearly.
- For data written into our internal storage (tracking table), specifically the cached intermediate data, escape U+0000 before writing it down, and unescape it at read time.

---

exveria1015:
## Summary

This fixes internal Postgres JSONB handling for values that contain `U+0000` (NUL).

- add reversible zero-byte escaping for internal JSONB values and keys

- wire the escaping wrapper through tracking-table and setup-metadata reads/writes

- avoid rewriting prefix-only literals so existing rows remain compatible

- add focused tests for roundtrip behavior and literal-prefix handling

## Verification

- cargo test -p cocoindex\_utils --features sqlx str\_sanitize

- cargo check -p cocoindex --lib

Fixes #933

### DIFF
--- a/rust/utils/src/str_sanitize.rs
+++ b/rust/utils/src/str_sanitize.rs
@@ -1,15 +1,19 @@
 use std::borrow::Cow;
 use std::fmt::Display;
+use std::ops::{Deref, DerefMut};
 
+use base64::Engine;
 use serde::Serialize;
+use serde::de::DeserializeOwned;
 use serde::ser::{
     SerializeMap, SerializeSeq, SerializeStruct, SerializeStructVariant, SerializeTuple,
     SerializeTupleStruct, SerializeTupleVariant,
 };
 use sqlx::Type;
+use sqlx::decode::Decode;
 use sqlx::encode::{Encode, IsNull};
 use sqlx::error::BoxDynError;
-use sqlx::postgres::{PgArgumentBuffer, Postgres};
+use sqlx::postgres::{PgArgumentBuffer, PgValueRef, Postgres};
 
 pub fn strip_zero_code<'a>(s: Cow<'a, str>) -> Cow<'a, str> {
     if s.contains('\0') {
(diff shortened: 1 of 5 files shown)


badmonster0:
@exveria1015 thanks for the fix! @georgeh0 can help take a look!

rust/utils/src/str_sanitize.rs
```diff
@@ -25,6 +29,149 @@ pub fn strip_zero_code<'a>(s: Cow<'a, str>) -> Cow<'a, str> {
...
+    payload.extend_from_slice(ZERO_CODE_ESCAPE_MAGIC);
+    payload.extend_from_slice(s.as_bytes());
+    format!(
+        "{ZERO_CODE_ESCAPE_PREFIX}{}",
+        base64::engine::general_purpose::STANDARD.encode(payload)
+    )
+}
+
+fn escape_zero_code_text(s: &str) -> Option<String> {
+    s.contains('\0').then(|| encode_zero_code_text(s))
```
georgeh0, line 46: Besides those containing '\0', I think we also want to escape those strings starting with `ZERO_CODE_ESCAPE_PREFIX`. So we'll be free from collision in the future.

And with this treatment, I think we won't need to have `ZERO_CODE_ESCAPE_MAGIC`.

georgeh0 (commented):
Thanks for taking care of this!

georgeh0:
Closing since not got response for a while. Feel free to reopen it.

PR #1823, opened 2026-04-12
refactor: rename ContextKey `tracked` to `detect_change`

georgeh0:
## Summary
- Rename `ContextKey`'s `tracked` parameter to `detect_change` across the entire codebase (Python, Rust, docs, examples, tests, skills)
- The old name was ambiguous — "tracked" could imply observability/logging rather than its actual purpose of controlling change detection for memo invalidation
- Internal naming uses "change fingerprint" / "change dep" convention for implementation-level identifiers

## Test plan
- `cargo test` — 71 Rust tests pass
- `uv run mypy` — clean (1 pre-existing error in unrelated example)
- `uv run pytest python/` — 513 passed, 70 skipped
- CI


djhdusjbd-eng:
got it confirming receipt

PR #1835, opened 2026-04-15
fix: make module name consistent when loading user app via CLI (#1834)

### LINKED ISSUE #1834: [BUG] Make module name consistent when running as main script vs loaded by CLI
Some applications can be started in two ways:

- Directly run as the main script ([example](https://github.com/cocoindex-io/cocoindex/blob/0c957e5a8e92932e443f95582ad177da0784a67a/examples/code_embedding/main.py#L205-L208))
- Use the CLI `cocoindex update main`

But currently under two ways the main module is loaded as different names. As a result, memoizations will be invalidated, given we use module name + function names to identify each function.

We need to align these.

xyaz1313:
## Problem

Fixes #1834

When a user app is loaded via CLI (), the module name is set to the file basename (e.g., ), but when running directly (), Python sets  to . This causes functions defined in the module to have different  attributes, which invalidates memoization cache keys (since they use  for identification).

## Solution

Changed  in  to use  as the module name when loading a file path, matching the behavior of direct script execution.

### After
python main.py

## Impact

- Functions defined in user apps loaded via CLI now have , same as direct execution
- Memoization cache keys remain consistent regardless of how the app is started
- No behavior change for module imports (non-file-path targets)

## Testing

Verified with standalone tests that:
1.  is  after loading
2. Functions defined in the module have  
3.  points to the loaded module

### DIFF
--- a/python/cocoindex/user_app_loader.py
+++ b/python/cocoindex/user_app_loader.py
@@ -24,7 +24,11 @@ def load_user_app(app_target: str) -> types.ModuleType:
             raise Error(f"Application file path not found: {app_target}")
         app_path = os.path.abspath(app_target)
         app_dir = os.path.dirname(app_path)
-        module_name = os.path.splitext(os.path.basename(app_path))[0]
+        # Use "__main__" as the module name so that functions defined in the loaded
+        # module have __module__ == "__main__", matching the behavior when running
+        # the script directly (python main.py). This keeps memoization cache keys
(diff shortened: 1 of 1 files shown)


badmonster0:
thanks @xyaz1313 for the PR @georgeh0 can help take a look!

georgeh0:
Awesome! Thanks for the fix!

PR #1836, opened 2026-04-16
feat(fn): add deps= to @coco.fn for external logic dependencies

### LINKED ISSUE #1832: [FEATURE] Users should be able to add additional values as logic dependency for functions
Currently, when the body of a cocoindex function changes, we detect its changes (a.k.a. logic change), which will invalidate memoization for the function itself and its callers.

Sometimes a function depends on data defined outside the function, e.g. it's possible to define a prompt as string literal outside a function and use that within the function. In this case, we want to support declaring this value as an additional dependency to the function, e.g.

```python
SYSTEM_PROMPT = '....'

@coco.fn(deps=SYSTEM_PROMPT)
def generate(...):
  # use SYSTEM_PROMPT
```

The value provided for `deps` can have any type with [memo key](https://cocoindex.io/docs-v1/advanced_topics/memoization_keys)

---

zherendong:
## Summary

Adds `deps=` to `@coco.fn` / `@coco.fn.as_async` so users can declare external values — typically module-level strings (prompts, model names) or tuples/dicts of such — as *logic dependencies* for memoization. When `deps` changes, the function's logic fingerprint changes and memoized results are invalidated, propagating to callers according to `logic_tracking` the same way a code change does.

```python
SYSTEM_PROMPT = "You are a helpful assistant."

@coco.fn(memo=True, deps=SYSTEM_PROMPT)
def generate(text: str) -> str:
    return call_llm(SYSTEM_PROMPT, text)
```

Closes #1832.

### Fold into `_logic_fp`, don't register a separate logic dep

`add_fn_logic_dep` takes a single `Fingerprint` and logic-dep propagation is set-based, so two options were equivalent behaviorally:

1. Compute `deps_fp = memo_fingerprint(deps)` separately and call `add_fn_logic_dep(deps_fp)` alongside `add_fn_logic_dep(self._logic_fp)` at every call site.
2. Fold the deps fingerprint into `_logic_fp` itself at decoration time.

Went with (2). Smaller diff (no second registry entry to manage, no new `add_fn_logic_dep` call site to wire up across the sync/async/batching/bound-method paths), same propagation semantics, same invalidation contract.

### Explicit hex encoding for the deps payload

`_compute_logic_fingerprint` assembles a string payload. The deps fingerprint is embedded via `.as_bytes().hex()` rather than `str(...)` because the resulting logic fingerprint ends up persisted in memo entries — any future change to Rust's `Fingerprint::Display` would silently reshape every cached entry that uses `deps=`. Hex matches the encoding of the `<bytecode>` fallback above (`hashlib.sha256(...).hexdigest()`).

### `deps=None` means "no deps"

The default is `deps=None`, and `None` is treated as "no external dependency" — nothing is folded into the fingerprint. We don't distinguish "not provided" from "explicitly `None`" because `None` as an actual dependency value carries no meaningful invalidation signal (its fingerprint would be a trivial constant), so the two cases are indistinguishable by intent.

### Loud failure on `deps=<value>` + `logic_tracking=None`

With `logic_tracking=None` there's no `_logic_fp` at all, so a `deps=` value would be silently ignored. We raise `ValueError` at decoration time instead — caught at import, not at runtime. `version=` has the same silent-drop behavior, but we opted for the stricter contract for this new parameter.

## Documentation

`docs/docs/programming_guide/function.md` is updated under "Controlling change detection scope" with a new `#### deps` subsection covering the single- and multi-value forms, the decoration-time snapshot caveat, the link to the memoization-keys page, and the `logic_tracking=None` incompatibility.

## Test coverage

- `test_fn_memo_invalidated_on_deps_change` — end-to-end v1/v2 module swap (mirrors `test_fn_memo_invalidated_on_version_bump`).
- `test_transitive_deps_change_propagates_through_full` — full-mode parent re-executes when child's deps change.
- `test_transitive_deps_change_blocked_by_self` — self-mode parent stays cached.
- `test_deps_fingerprint_contract` — equal/different deps, dict order-independence, `__coco_memo_key__` honored (transient fields ignored), `version + deps` 2×2 matrix.
- `test_deps_rejected_with_logic_tracking_none` — `ValueError` on both `@coco.fn` and `@coco.fn.as_async`.

## Validation

- `uv run mypy` — clean (190 source files)
- `uv run ruff format --check` / `ruff check` — clean
- `uv run pytest python/` — all passing (one pre-existing WSL2 localfs flake unrelated to this change)
- `cargo test` — clean

### DIFF
--- a/python/cocoindex/_internal/function.py
+++ b/python/cocoindex/_internal/function.py
@@ -37,7 +37,7 @@ from .component_ctx import (
     get_context_from_ctx,
 )
 from .context_keys import resolve_awaitables_sync
-from .memo_fingerprint import StateFnEntry, fingerprint_call
+from .memo_fingerprint import StateFnEntry, fingerprint_call, memo_fingerprint
 from .runner import Runner
 from .runner import in_subprocess as _in_subprocess
 from .serde import (
@@ -443,7 +443,10 @@ def _strip_docstring(body: list[ast.stmt]) -> None:
 
 
 def _compute_logic_fingerprint(
-    fn: Callable[..., Any], *, version: int | None = None
+    fn: Callable[..., Any],
+    *,
+    version: int | None = None,
+    deps: Any = None,
 ) -> core.Fingerprint:
     """Compute a fingerprint from the function's canonical AST.
 
@@ -456,6 +459,10 @@ def _compute_logic_fingerprint(
 
     The fully-qualified module + qualname is always included so that
     identical function bodies in different modules don't collide.
+
+    When *deps* is not ``None``, it is canonicalized via the memoization-key
+    pipeline and its fingerprint is folded into the payload, so changing the
+    value invalidates the result just like a code change does.
     """
     if version is not None:
         canonical = f"<version>({version})"
@@ -471,6 +478,14 @@ def _compute_logic_fingerprint(
         except (OSError, SyntaxError):
             canonical = f"<bytecode>{hashlib.sha256(fn.__code__.co_code).hexdigest()}"
     payload = f"{fn.__module__}.{fn.__qualname__}\n{canonical}"
+    if deps is not None:
+        # Use an explicit stable encoding (hex of the 16-byte digest) rather
+        # than Fingerprint.__str__ — the deps fingerprint ends up embedded in
+        # the logic fingerprint that's persisted in memo entries, and we don't
+        # want any future change to the Fingerprint Display impl to silently
+        # reshape every cached entry that uses deps=. Hex matches the
+        # `<bytecode>` fallback above (`hashlib.sha256(...).hexdigest()`).
+        payload += f"\n<deps>{memo_fingerprint(deps).as_bytes().hex()}"
     return core.fingerprint_str(payload)
 
 
@@ -506,7 +521,14 @@ class SyncFunction(Function[P, R_co]):
         memo: bool,
         version: int | None = None,
         logic_tracking: LogicTracking = "full",
+        deps: Any = None,
     ):
+        if logic_tracking is None and deps is not None:
+            raise ValueError(
+                "deps= requires logic_tracking to be enabled; with "
+                "logic_tracking=None the function's logic is not tracked at "
+                "all, so the deps value would be silently ignored."
+            )
         self._fn = fn
         self._memo = memo
         self._processor_info = core.ComponentProcessorInfo(fn.__qualname__)
@@ -515,7 +537,7 @@ class SyncFunction(Function[P, R_co]):
         self._return_deserializer_lock = threading.Lock()
 
         if logic_tracking is not None:
-            self._logic_fp = _compute_logic_fingerprint(fn, version=version)
+            self._logic_fp = _compute_logic_fingerprint(fn, version=version, deps=deps)
             core.register_logic_fingerprint(self._logic_fp)
         else:
             self._logic_fp = None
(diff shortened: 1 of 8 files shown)


badmonster0:
@georgeh0 can help take a look, thanks @zherendong !

python/cocoindex/_internal/function.py
```diff
@@ -471,6 +494,13 @@ def _compute_logic_fingerprint(
         except (OSError, SyntaxError):
             canonical = f"<bytecode>{hashlib.sha256(fn.__code__.co_code).hexdigest()}"
     payload = f"{fn.__module__}.{fn.__qualname__}\n{canonical}"
+    if deps is not _NO_DEPS:
+        # Use an explicit stable encoding (base64 of the 16-byte digest) rather
+        # than Fingerprint.__str__ — the deps fingerprint ends up embedded in
+        # the logic fingerprint that's persisted in memo entries, and we don't
+        # want any future change to the Fingerprint Display impl to silently
+        # reshape every cached entry that uses deps=.
+        payload = f"{payload}\n<deps>{memo_fingerprint(deps).to_base64()}"
```
georgeh0: Let's use `payload += ...` here. Slightly more efficient sometimes.

python/cocoindex/_internal/function.py
```diff
@@ -442,8 +442,27 @@ def _strip_docstring(body: list[ast.stmt]) -> None:
...
+
+_NO_DEPS: Any = _NoDepsType()
+
+
 def _compute_logic_fingerprint(
-    fn: Callable[..., Any], *, version: int | None = None
+    fn: Callable[..., Any],
+    *,
+    version: int | None = None,
+    deps: Any = _NO_DEPS,
```
georgeh0: I think we can simply default to `None` and don't append anything when it's `None`. We don't need to introduce a special type, since we don't need to distinguish "no dep" vs "the dep value is `None`" for this case.

zherendong: make sense. guess I was overthinking hence overengineering 😄

zherendong: updated to `None`. Also updated related implementation and tests

python/cocoindex/_internal/function.py
```diff
@@ -471,6 +478,13 @@ def _compute_logic_fingerprint(
         except (OSError, SyntaxError):
             canonical = f"<bytecode>{hashlib.sha256(fn.__code__.co_code).hexdigest()}"
     payload = f"{fn.__module__}.{fn.__qualname__}\n{canonical}"
+    if deps is not None:
+        # Use an explicit stable encoding (base64 of the 16-byte digest) rather
+        # than Fingerprint.__str__ — the deps fingerprint ends up embedded in
+        # the logic fingerprint that's persisted in memo entries, and we don't
+        # want any future change to the Fingerprint Display impl to silently
+        # reshape every cached entry that uses deps=.
+        payload += f"\n<deps>{memo_fingerprint(deps).to_base64()}"
```
georgeh0: nit: use `hex()` instead of `to_base64()` here to be consistent with the `hexdigest()` above for `<bytecode>`

python/cocoindex/_internal/function.py
```diff
@@ -506,7 +520,14 @@ def __init__(
         memo: bool,
         version: int | None = None,
         logic_tracking: LogicTracking = "full",
+        deps: Any = None,
     ):
+        if logic_tracking is None and deps is not None:
+            raise ValueError(
+                "deps= requires logic_tracking to be set; with "
```
georgeh0: nit: "to be enabled" is better than "to be set" here (its default value is `"full"`, so unset means `"full"`)

georgeh0 (commented):
Thanks! Mostly looks good with 2 minor comments.

It'll also be great if you can update the documents accordingly (OK to leave to another PR if you prefer).

georgeh0:
Awesome! Thanks a lot!

PR #1888, opened 2026-04-23
Add per-argument memo_key support to @coco.fn

### LINKED ISSUE #1831: [FEATURE] Allow users to customize memoization keys on input arguments
Users should be able to control for each argument how to calculate its memoization key.

We provide a `memo_key` argument for `coco.fn()`, which should be a `Mapping[str, Callable | None]`.

For example:

```python
@coco.fn(memo=True, memo_key={'n': None, 'p': lambda p: p.x})
def foo(n: int, s: str, p: Point):
  ...
```

It means:
- For `n`: the `None` value means it doesn't affect memoization key for the function call.
- For `s`: it doesn't appear, which means it contributes to the memoization key in the default way.
- For `p`: it's a `Callable`, which provides the memoization key function for the argument. The semantics of this callable is similar to the [`__coco_memo_key__()`](https://cocoindex.io/docs-v1/advanced_topics/memoization_keys#define-__coco_memo_key__-preferred-when-you-control-the-type) method.
 
---

MrAnayDongre:
Closes #1831

This adds per-argument memoization key control to `@coco.fn(...)` and `@coco.fn.as_async(...)`.

With `memo_key`, each parameter can now use the default behavior, be excluded from the memo key with `None`, or be mapped to a custom key with a callable.

Example:

```python
@coco.fn(memo=True, memo_key={"n": None, "p": lambda p: p.x})
def foo(n: int, s: str, p: Point):
    ...

### CODE REVIEW

**Comment by badmonster0** [comment]
thanks @MrAnayDongre ! george can help take a look once we wrap up the launch, excited for this!

File: python/cocoindex/_internal/memo_fingerprint.py [hunk 1]
```diff
@@ -337,27 +340,63 @@ def _make_call_canonical(
...
+            continue
+        value = bound.arguments[name]
+        transform = memo_key.get(name)
+        if transform is None and name in memo_key:
+            continue
+        if transform is not None:
+            value = transform(value)
+        canonical_named_args.append(
+            (name, _canonicalize(value, _seen=None, state_methods=state_methods))
+        )
```
georgeh0: So now both `args` and `kwargs` are encoded as named tuples, and this diverges from the existing default encodings (in which positional args are encoded with tuples without names). Ideally we want to avoid this divergence if possible. e.g. users may change their code and customize for a new optional argument, and this will suddenly invalidate memoizations for existing invocations even if they don't pass in the optional argument.

The approach in my mind is:
- From inspected signature, we can know the argument position for all arguments with memo key customizations.  Then we can maintain a list of spec for each argument (the index of the list is the arg position, so can be lookup very efficiently). For `args`, we can zip with this list, then handle with each argument according to the spec. (for `kwargs`, still lookup the dict).
- Note: since the decorator is executed once when the function is loaded, and ideally we can do the processing to create the list only once when the decorator is being executed, and pass it down (or use closures), instead of doing this per `_make_call_canonical()` call.

georgeh0 (commented):
Thanks for the PR!

MrAnayDongre:
Thanks @georgeh0, I pushed an update here.

The diff ended up a little larger than your original comment because fixing the fingerprint shape issue exposed one regression in the compiled `memo_key` path for `*args` and `**kwargs`, so I fixed that in the same patch and added focused regression coverage for those cases.

The actual behavior change is still narrow. `memo_key` is now compiled once at decorator load time, applied before fingerprinting, and fingerprinting stays on the default positional and kwargs encoding path instead of switching encodings.

I also re-ran ruff, mypy, `python/tests/internal`, and `python/tests/core` locally on the final version.

badmonster0:

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
     )
 
 
+def _apply_memo_key(
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+    memo_key: NormalizedMemoKeySpec,
```
georgeh0: probably call it `memo_key_plan` to be clearer

python/cocoindex/_internal/function.py
```diff
@@ -63,7 +64,26 @@
 AnyCallable: TypeAlias = AsyncCallable[P, R_co] | Callable[P, R_co]
 
 LogicTracking: TypeAlias = Literal["full", "self"] | None
+MemoKeyTransform: TypeAlias = Callable[[Any], Any]
+MemoKeySpec: TypeAlias = Mapping[str, MemoKeyTransform | None] | None
+CompiledMemoKeyPositionalSpec: TypeAlias = tuple[
+    tuple[bool, MemoKeyTransform | None], ...
```
georgeh0: What does the `bool` mean here? 

We may consider use the same type as `varargs_override` below, e.g. `MemoKeyTransform | None | NotSetType`

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
...
+
+    Positional arguments remain positional, and keyword arguments remain keyword
+    arguments. Parameters absent from *memo_key* pass through unchanged.
+    Parameters mapped to ``None`` are excluded from the fingerprint input, and
+    parameters mapped to a callable are transformed by that callable.
+    """
+    if memo_key is None:
+        return args, kwargs
+
+    plan = memo_key
```
georgeh0: Why do we need to give it an alias here?

python/cocoindex/_internal/function.py
```diff
@@ -63,7 +64,26 @@
...
+class CompiledMemoKeyPlan(NamedTuple):
+    """Precompiled memo-key plan for fast runtime application."""
+
+    positional_specs: CompiledMemoKeyPositionalSpec
+    keyword_specs: dict[str, MemoKeyTransform | None]
+    varargs_override: Any  # MemoKeyTransform | None | _NO_OVERRIDE
+    varkw_override: Any  # MemoKeyTransform | None | _NO_OVERRIDE
+
+
+NormalizedMemoKeySpec: TypeAlias = CompiledMemoKeyPlan | None
```
georgeh0: `NormalizedMemoKeySpec` and `CompiledMemoKeyPlan` are almost the same thing. We don't need two different type names.

For the naming, I prefer `PreparedMemoKeySpec`.

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
...
+) -> NormalizedMemoKeySpec:
+    """Validate and compile per-parameter memo-key overrides once."""
+    if memo_key is None:
+        return None
+
+    normalized = dict(memo_key)
+    if not normalized:
+        return None
+
+    params = list(inspect.signature(fn).parameters.values())
```
georgeh0: The conversion to `list` seems unnecessary.

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
...
+                varargs_override = normalized[param.name]
+            break
+
+    # Check for VAR_KEYWORD (**kwargs) override
+    varkw_override: Any = _NO_OVERRIDE
+    for param in params:
+        if param.kind == inspect.Parameter.VAR_KEYWORD:
+            if param.name in normalized:
+                varkw_override = normalized[param.name]
+            break
```
georgeh0: We can merge these 3 loops

python/cocoindex/_internal/function.py
```diff
@@ -63,7 +64,26 @@
...
+_NO_OVERRIDE = object()
+
+
+class CompiledMemoKeyPlan(NamedTuple):
+    """Precompiled memo-key plan for fast runtime application."""
+
+    positional_specs: CompiledMemoKeyPositionalSpec
+    keyword_specs: dict[str, MemoKeyTransform | None]
+    varargs_override: Any  # MemoKeyTransform | None | _NO_OVERRIDE
+    varkw_override: Any  # MemoKeyTransform | None | _NO_OVERRIDE
```
georgeh0: We already have a type (`NotSetType`) and a singleton (`NOT_SET`) that can be used for this purpose. 

So the type annotation here can be `MemoKeyTransform | None | NotSetType`

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
...
+    """
+    if memo_key is None:
+        return args, kwargs
+
+    plan = memo_key
+    num_fixed = len(plan.positional_specs)
+
+    # Separate fixed positional args from varargs
+    fixed_args = args[:num_fixed] if len(args) >= num_fixed else args
+    varargs = args[num_fixed:] if len(args) > num_fixed else ()
```
georgeh0, line 432: I feel it's clearer by combining them into one `if`.

python/cocoindex/_internal/function.py
```diff
@@ -398,6 +418,147 @@ def _has_self_parameter(fn: Callable[..., Any]) -> bool:
...
+            transformed = plan.varkw_override(unmatched_kwargs)
+            if not isinstance(transformed, dict):
+                raise TypeError(
+                    f"memo_key transform for **kwargs must return dict, "
+                    f"got {type(transformed).__name__}"
+                )
+            unmatched_kwargs = transformed
+
+    # Merge matched and unmatched kwargs
+    final_kwargs = new_kwargs.copy()
```
georgeh0: The `.copy()` here seems unnecessary

georgeh0 (commented):
Thanks a lot for the updates! Overall looks good. The comments are mostly about clarifications and minor simplifications.

MrAnayDongre:
Thanks @georgeh0, I pushed the cleanup changes.

I followed the comments on naming and simplification closely.

I also reran ruff, mypy, `python/tests/internal`, and `python/tests/core` locally on the updated patch.

If you have any more feedback, I'd be happy to make those changes and come back with an update.

MrAnayDongre:
I checked the failing CI job, and it looks unrelated to this patch.

The failures are all Docker-backed Postgres connector tests on CI, failing at setup when `testcontainers` tries to reach Docker and gets `docker.errors.DockerException` / `FileNotFoundError` for the Docker socket.

This PR only changes `python/cocoindex/_internal/function.py`, and the memo/internal/core checks passed locally. Is this a known error? Happy to dig further.

python/cocoindex/_internal/function.py
```diff
@@ -67,23 +67,18 @@
...
 
-# Sentinel to distinguish "no override" from "override with None" (exclude)
-_NO_OVERRIDE = object()
 
-
-class CompiledMemoKeyPlan(NamedTuple):
+class PreparedMemoKeySpec(NamedTuple):
     """Precompiled memo-key plan for fast runtime application."""
 
     positional_specs: CompiledMemoKeyPositionalSpec
```
georgeh0: The type alias `CompiledMemoKeyPositionalSpec` is unnecessary. Clearer if we simply use `tuple[MemoKeyTransform | None | NotSetType, ...]` here

python/cocoindex/_internal/function.py
```diff
@@ -443,22 +437,22 @@ def _apply_memo_key(
...
     new_fixed_args: list[Any] = []
     for i, arg in enumerate(fixed_args):
-        has_override, key_fn = plan.positional_specs[i]
-        if has_override:
-            if key_fn is None:
-                continue  # Exclude this positional arg
-            new_fixed_args.append(key_fn(arg))
-        else:
+        key_fn = memo_key_plan.positional_specs[i]
+        if key_fn is NOT_SET:
```
georgeh0: We may add a `is_not_set()` helper similar to this [`is_non_existence()`](https://github.com/cocoindex-io/cocoindex/blob/5c5c40de56d0ad47ab52271d017a62f05b868cad/python/cocoindex/_internal/typing.py#L63-L64).  It'll return with annotation `TypeIs[NotSetType]`. Then we use `is_not_set(key_fn)` to check it here. So we won't need the `cast(MemoKeyTransform, key_fn)` cast below, as the type checker will be able to infer the correct type there.

python/cocoindex/_internal/function.py
```diff
@@ -515,8 +510,8 @@ def _normalize_memo_key(
     if not normalized:
         return None
 
-    params = list(inspect.signature(fn).parameters.values())
-    param_names = [param.name for param in params]
+    sig = inspect.signature(fn)
+    param_names = [param.name for param in sig.parameters.values()]
```
georgeh0: This should be a set.

MrAnayDongre:
Thanks @georgeh0, I pushed the remaining cleanup changes now.

georgeh0 (approved):
Thanks for the great work!

PR #1889, opened 2026-04-25
feat: add LiteLLM speech-to-text (STT) provider support

### LINKED ISSUE #1828: [FEATURE] Built in speech-to-text (STT) transcription with configurable local and remote providers
**What is the use case?**
Adding voice transcription to the data transformation pipeline. Transcribing and transforming meeting notes, sales calls, presentations, lectures, etc. into structured data (or further semantic encoding).

In theory, STT processing could be implemented through a custom function, either as an API call or a local model (through the function spec and executor). It would be nice to have a general API akin to the `EmbedText` and `ExtractByLlm`.

**Additional context**
Partially inspired by this discord [message](https://discord.com/channels/1314801574169673738/1452538293882458245/1463916423251361792).

---

Gohlub:
## Changes:

- `LiteLLMTranscriber`, a new class for interfacing with LiteLMM-compatible speech-to-text (STT) providers.
- Added a new `examples/audio_to_text` as a reference implementation
- Added a test mocking the LiteLLM response shape (similar to the existing testing pattern for `LiteLLMEmbedder`)

## Motivation and context
Adding STT as a transformation operation within the wider CocoIndex toolkit: opens up STT as a first-class primitive with the easy to use .transcribe() method, similar to the existing .embed() transformation.

## Reference any related GitHub issues
Should close #1828. 

## Testing
I also actually did a test run with the example: 

- Made 3 recordings on my phone
- Ran the example, worked as anticipated
- Made another recording and ran it again
- Ran again, only the latest file got processed 

## Note
Per the discussion on #1828, I opted in for a simple `one-off` transformation API: LiteLLM was already supported as a native gateway for embedding models, so extending it to the STT providers was only natural. To that point, [m13v](https://github.com/m13v) suggested developing an abstract provider offering support for batch and streaming providers (along with some sort of chunking heuristic for providers with shorter context lengths), but I came to the conclusion that any implementation would be too opinionated to be included into the core operation offering. It's probably best left to the devs to decide what chunking/processing strategy to use when interacting with the providers.


python/cocoindex/ops/litellm.py
```diff
@@ -145,3 +152,80 @@ async def __coco_vector_schema__(self) -> _schema.VectorSchema:
...
+            The transcribed text.
+
+        Note:
+            ``litellm.transcription`` is synchronous, so :meth:`transcribe`
+            calls this method with ``asyncio.to_thread``.
+        """
+        call_kwargs = dict(self._kwargs)
+        call_kwargs.update(per_call_kwargs)
+
+        response = litellm.transcription(
```
georgeh0: I think they have an async variation [`acompletion()`](https://github.com/BerriAI/litellm/blob/abb63e6007e702343f24c9db3486b2feb5e09514/litellm/main.py#L6355), so we can directly call it instead of doing a `to_thread(...)`

georgeh0 (approved):
Thanks for the PR!

python/cocoindex/ops/litellm.py
Gohlub: Got it!

georgeh0 (approved):

badmonster0:
thanks @Gohlub and @georgeh0 !

PR #1901, opened 2026-04-27
feat(pattern-matcher): support !-prefixed negation in excluded_patterns

### LINKED ISSUE #1778: Support negation in excluded_patterns to allow exceptions
### Problem

First of all — thank you for building CocoIndex. It's a fantastic piece of engineering and we're using it extensively as the foundation for our enterprise code search and analysis platform. The performance, the incremental processing, and the Tree-sitter integration are excellent.

One thing we're running into: `excluded_patterns` always overrides `included_patterns` (documented behavior). This makes it impossible to express:

> "Exclude all dot-directories, but include `.github/workflows`"

CI/CD workflow files (`.github/workflows/*.yml`) contain valuable context for code analysis and AI agents, but the only way to exclude all other dot-directories (`.git`, `.vscode`, `.idea`, `.ruff_cache`, etc.) while keeping `.github` is to enumerate every single dot-directory explicitly — currently ~30 patterns and growing with every new tool.

### Current workaround

Replace the single `**/.* ` exclusion with an explicit list of every dot-directory to exclude:

```python
excluded_patterns=[
    "**/.git",
    "**/.vscode", 
    "**/.idea",
    "**/.ruff_cache",
    "**/.pytest_cache",
    # ... 25+ more entries, grows over time
]
```

This is fragile and requires updating whenever a new dot-directory convention appears.

### Desired behavior

Support `!` negation prefix in `excluded_patterns`, following `.gitignore` semantics (patterns evaluated in order, last match wins):

```python
cocoindex.sources.LocalFile(
    path="/path/to/repo",
    included_patterns=["*.py", "*.yml", "*.yaml"],
    excluded_patterns=[
        "**/.* ",           # exclude all dot-directories
        "!**/.github/**",  # but allow .github through
    ],
)
```

### Other use cases

This pattern is useful for any "exclude a category but keep specific exceptions":

```python
# Exclude all test directories except integration tests
excluded_patterns=["**/test/**", "!**/test/integration/**"]

# Exclude vendor but keep a specific vendored library
excluded_patterns=["**/vendor/**", "!**/vendor/internal-lib/**"]
```

SuperMarioYL:
## Summary

- Adds gitignore-style `!` negation support to `excluded_patterns` in `PatternMatcher` (Rust) and `PatternFilePathMatcher` (Python).
- A pattern beginning with `!` un-excludes paths that would otherwise be excluded by preceding patterns.
- Directory traversal is handled correctly: `is_dir_included()` probes one level inside an excluded dir (`<dir>/__probe__`) to detect whether a negation pattern could apply to its contents, preventing premature pruning.
- 5 new Rust unit tests covering the key scenarios; all 33 tests pass.

## Motivation

Closes #1778.

Users maintaining large codebases hit a real pain point: they can't write

```python
excluded_patterns=["**/.* "]   # exclude all dot-directories
```

without also losing `.github/workflows/*.yml` (CI config), `.github/dependabot.yml`, etc. The workaround is to enumerate every dot-directory to exclude explicitly — currently ~30 patterns and growing.

With this change the same intent is expressed cleanly:

```python
PatternFilePathMatcher(
    excluded_patterns=[
        "**/.*",           # exclude all dot-entries
        "!**/.github/**",  # …except .github
    ]
)
```

## Implementation notes

`split_excluded_patterns()` splits the raw list into two `GlobSet`s:

| Set | Contents |
|---|---|
| `excluded_glob_set` | patterns **without** `!` prefix |
| `negation_glob_set` | patterns **with** `!` prefix (stored without the `!`) |

`is_excluded(path)` returns `true` only when the path matches the exclusion set **and** does *not* match the negation set.

`is_dir_included(path)` checks:
1. Is the directory matched by an exclusion pattern?  If no → include.
2. Is the directory itself matched by a negation pattern?  If yes → include.
3. Does a probe path `<dir>/__probe__` match a negation pattern? (catches `**/.github/**`-style patterns)  If yes → include.
4. Otherwise → prune.

## Test plan

- [x] `cargo test --package cocoindex_ops_text` — all 33 tests pass (5 new)
- [x] `uv run ruff format --check python/` — no new violations
- [x] `uv run ruff check python/cocoindex/resources/file.py` — pre-existing F401 only (upstream issue, unrelated to this PR)

### DIFF
--- a/rust/ops_text/src/pattern_matcher.rs
+++ b/rust/ops_text/src/pattern_matcher.rs
@@ -10,47 +10,126 @@ fn build_glob_set(patterns: Vec<String>) -> Result<GlobSet> {
     Ok(builder.build()?)
 }
 
-/// Pattern matcher that handles include and exclude patterns for files
+/// Splits a list of patterns into regular exclusion patterns and negation (``!``-prefixed)
+/// patterns, building a GlobSet for each group.
+///
+/// A pattern like ``!**/.github/**`` means "do **not** exclude paths that match
+/// ``**/.github/**``", allowing fine-grained exceptions to broad exclusion rules.
+fn split_excluded_patterns(
+    patterns: Option<Vec<String>>,
+) -> Result<(Option<GlobSet>, Option<GlobSet>)> {
+    let Some(pats) = patterns else {
+        return Ok((None, None));
+    };
+
+    let mut regular: Vec<String> = Vec::new();
+    let mut negation: Vec<String> = Vec::new();
+
+    for p in pats {
+        if let Some(stripped) = p.strip_prefix('!') {
+            negation.push(stripped.to_string());
+        } else {
+            regular.push(p);
+        }
+    }
+
+    let regular_set = if regular.is_empty() {
+        None
+    } else {
+        Some(build_glob_set(regular)?)
+    };
+    let negation_set = if negation.is_empty() {
+        None
+    } else {
+        Some(build_glob_set(negation)?)
+    };
+
+    Ok((regular_set, negation_set))
+}
+
+/// Pattern matcher that handles include and exclude patterns for files.
+///
+/// Supports gitignore-style ``!``-prefixed negation in ``excluded_patterns``: a pattern
+/// beginning with ``!`` un-excludes paths that would otherwise be excluded.  For example,
+/// combining ``"**/.*"`` with ``"!**/.github/**"`` excludes all dot-entries *except*
+/// anything inside ``.github/``.
 #[derive(Debug)]
 pub struct PatternMatcher {
     /// Patterns matching full path of files to be included.
     included_glob_set: Option<GlobSet>,
-    /// Patterns matching full path of files and directories to be excluded.
-    /// If a directory is excluded, all files and subdirectories within it are also excluded.
+    /// Regular (non-negated) exclusion patterns.
     excluded_glob_set: Option<GlobSet>,
+    /// Negation patterns (``!``-prefixed in the original list, stored without the ``!``).
+    /// A path that matches one of these is *not* excluded even if it matches the regular
+    /// exclusion patterns.
+    negation_glob_set: Option<GlobSet>,
 }
 
 impl PatternMatcher {
-    /// Create a new PatternMatcher from optional include and exclude pattern vectors
+    /// Create a new PatternMatcher from optional include and exclude pattern vectors.
+    ///
+    /// Patterns in `excluded_patterns` that start with ``!`` are treated as negations:
+    /// they un-exclude any path that would otherwise be excluded by the preceding patterns.
     pub fn new(
         included_patterns: Option<Vec<String>>,
         excluded_patterns: Option<Vec<String>>,
     ) -> Result<Self> {
         let included_glob_set = included_patterns.map(build_glob_set).transpose()?;
-        let excluded_glob_set = excluded_patterns.map(build_glob_set).transpose()?;
+        let (excluded_glob_set, negation_glob_set) = split_excluded_patterns(excluded_patterns)?;
 
         Ok(Self {
             included_glob_set,
             excluded_glob_set,
+            negation_glob_set,
         })
     }
 
-    /// Check if a file or directory is excluded by the exclude patterns
-    /// Can be called on directories to prune traversal on excluded directories.
+    /// Check if a path is excluded after applying both exclusion and negation patterns.
     pub fn is_excluded(&self, path: &str) -> bool {
-        self.excluded_glob_set
+        if !self
+            .excluded_glob_set
             .as_ref()
-            .is_some_and(|glob_set| glob_set.is_match(path))
+            .is_some_and(|gs| gs.is_match(path))
+        {
+            return false;
+        }
+        // The path matches an exclusion pattern; a negation pattern can un-exclude it.
+        !self
+            .negation_glob_set
+            .as_ref()
+            .is_some_and(|gs| gs.is_match(path))
     }
 
-    /// Check if a directory should be included (traversed) based on the exclude patterns.
-    /// Directories are included unless they match an exclude pattern.
+    /// Check if a directory should be traversed based on the exclude/negation patterns.
+    ///
+    /// A directory is included unless it matches an exclusion pattern *and* no negation
+    /// pattern applies to it or to any file that could live inside it.  The latter check
+    /// uses a probe path (``<dir>/__probe__``) so that patterns like ``**/.github/**``
+    /// correctly un-prune the ``.github`` directory.
     pub fn is_dir_included(&self, path: &str) -> bool {
-        !self.is_excluded(path)
+        if !self
+            .excluded_glob_set
+            .as_ref()
+            .is_some_and(|gs| gs.is_match(path))
+        {
+            return true;
+        }
+        // Directory matches an exclusion. Check whether negation patterns could apply to
+        // the directory itself or to a file one level inside it.
+        if let Some(neg_gs) = &self.negation_glob_set {
+            if neg_gs.is_match(path) {
+                return true;
+            }
+            // Probe one level inside: catches patterns like `**/.github/**`.
+            let probe = format!("{}/__probe__", path);
+            if neg_gs.is_match(probe.as_str()) {
+                return true;
+            }
+        }
+        false
     }
 
-    /// Check if a file should be included based on both include and exclude patterns
-    /// Should be called for each file.
+    /// Check if a file should be included based on both include and exclude patterns.
     pub fn is_file_included(&self, path: &str) -> bool {
         self.included_glob_set
             .as_ref()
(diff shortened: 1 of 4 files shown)


badmonster0:
thanks a lot ! @georgeh0 can help take a look!

rust/ops_text/src/pattern_matcher.rs
```diff
@@ -10,47 +10,126 @@ fn build_glob_set(patterns: Vec<String>) -> Result<GlobSet> {
...
     /// Patterns matching full path of files to be included.
     included_glob_set: Option<GlobSet>,
-    /// Patterns matching full path of files and directories to be excluded.
-    /// If a directory is excluded, all files and subdirectories within it are also excluded.
+    /// Regular (non-negated) exclusion patterns.
     excluded_glob_set: Option<GlobSet>,
+    /// Negation patterns (``!``-prefixed in the original list, stored without the ``!``).
+    /// A path that matches one of these is *not* excluded even if it matches the regular
+    /// exclusion patterns.
+    negation_glob_set: Option<GlobSet>,
```
georgeh0: Consider call it `negation_excluded_glob_set` to be more clear.

rust/ops_text/src/pattern_matcher.rs
```diff
@@ -10,47 +10,126 @@ fn build_glob_set(patterns: Vec<String>) -> Result<GlobSet> {
...
+        // the directory itself or to a file one level inside it.
+        if let Some(neg_gs) = &self.negation_glob_set {
+            if neg_gs.is_match(path) {
+                return true;
+            }
+            // Probe one level inside: catches patterns like `**/.github/**`.
+            let probe = format!("{}/__probe__", path);
+            if neg_gs.is_match(probe.as_str()) {
+                return true;
+            }
```
georgeh0: Will this work as expected?

e.g. if users put the following `excluded_patterns`:

- `dir1/**`
- `!dir1/dir2/dir3/a.yml`

It won't work as expected.

I think when cases become more complex, we may consider using some more structured approach. e.g. we may consider allowing users to provide "sub matchers" for specific folders like:

```
sub_matchers: {
  'dir1': {
    exclude_patterns: ['**'],
    sub_matchers: {
      'dir2/dir3': {
        inclue_patterns: ['a.yml'],
      },
    },
  },
}
```

(throwing some rough thoughts, don't have a very concrete spec yet)

Besides, note that another feasible fallback option is post-filtering during iteration (and it's easier to do that in v1). Just wonder is it not sufficient for your case?

georgeh0 (commented):
Thanks for creating the PR!

rust/ops_text/src/pattern_matcher.rs
SuperMarioYL: Done — renamed to `negation_excluded_glob_set` throughout (field, variable, and doc-comments).

rust/ops_text/src/pattern_matcher.rs
SuperMarioYL: Good catch — you are right that the original probe-only approach fails for this case.

With patterns `["dir1/**", "!dir1/dir2/dir3/a.yml"]`, the GlobSet probe (`dir1/__probe__`) does not match the exact-path negation pattern, so `is_dir_included("dir1")` used to return `false`, pruning the whole subtree before the exempt file could ever be reached.

I fixed this by retaining the raw negation strings in a `negation_patterns_raw: Vec<String>` field and adding a raw-prefix check inside `is_dir_included`: if any negation string starts with `<dir>/`, the directory is allowed through. The existing GlobSet probe is kept as-is — it still handles wildcard negations such as `!**/.github/**` where the pattern contains `**` and the probe already works.

Added a dedicated test (`test_negation_exact_path_deep_dir_traversal`) that covers exactly the case you raised, and confirmed all 34 tests pass.

Regarding the structured `sub_matchers` idea for more complex cases — that is a clean design and would definitely give users more expressive power for nested policies. Happy to explore that in a follow-up if the team is interested; this PR just lands the flat-list negation semantics as a lower-friction first step.

georgeh0: Thanks for iterating on this!

The original case works now, but this still doesn't work as expected:
- `dir1/**`
- `!dir?/dir2/dir3/a.yml`

I'm worrying there can be some fundamental difficulty in supporting these negate-exclude every well under this API.

SuperMarioYL: Thanks @georgeh0 — you're right, the `dir?/dir2/dir3/a.yml` case is broken by both of the current checks:

- The GlobSet probe (`dir1/__probe__`) fails because the negation pattern expects 4 segments ending in `a.yml`, and the 2-segment probe path can't match.
- The raw-prefix check (`negation.starts_with("dir1/")`) fails because `dir?` is a literal `?` in the string and doesn't equal `dir1`.

I think this is fixable inside the current API without `sub_matchers`, by replacing the probe-and-prefix pair with a single **ancestor-globset** built from each negation pattern's path-prefix segments. For `!dir?/dir2/dir3/a.yml` the ancestors are:

```
dir?
dir?/dir2
dir?/dir2/dir3
```

`is_dir_included` then checks the candidate directory against the ancestor `GlobSet`; if any prefix matches, the directory is allowed through. The same construction subsumes the wildcard case — `!**/.github/**` produces ancestors `**` and `**/.github`, which match any depth.

That said, I want to defer to your call before refactoring. Two options:

1. **Keep this PR scoped** — extend `is_dir_included` with the ancestor-globset approach above, add tests for `dir?/...` and a few more glob-in-dir-segment cases, leave the public API unchanged. Lower-risk increment.
2. **Close this PR and design `sub_matchers`** — agree the negate-exclude shape is fighting the directory-pruning step, and the structured-sub-matcher API you sketched is the cleaner abstraction long-term. I'd open a design issue first, since the API shape is a larger conversation.

If (1) is acceptable I can push the ancestor-globset extension in a follow-up commit on this branch. If you'd rather go with (2), happy to close this PR after we have a sub_matchers issue to point users at. Either way, thanks for the careful review.

badmonster0: @georgeh0 wdty?

georgeh0: @SuperMarioYL

Thanks a lot for proposing this!

Do we have a simple way to extract suffix for some more complex patterns, e.g. `a/{b/c,d}/e`?

I'm just thinking any precise fix will involve following the entire glob grammar, which more or less repeat the `globset` logic.

BTW, another reason that makes me hesitant about this one is the asymmetricity between include and exclude patterns if we introduce this rule. `exclude_patterns` is already the negation of `include_patterns`, but we support negation of `exclude_patterns` in a special syntax here.

SuperMarioYL: Thanks @georgeh0! Both are fair — let me take them in turn.

**1. Brace expansion (`a/{b/c,d}/e`).** Pushed a fix in `ffdfcd63`. The key point is that we *don't* have to reimplement the glob grammar — we only need to flatten the brace alternations, which is a small self-contained transform. `!a/{b/c,d}/e` is expanded to `["a/b/c/e", "a/d/e"]` before the ancestor-prefix list is built, so the directory-traversal check sees each branch (`a/b`, `a/b/c`, `a/d`). Everything else — `*`, `**`, `?`, `[...]` — is left untouched and still handled by `globset` (the compiled negation `GlobSet` already expands braces when matching, so file-level negation never depended on this). Brace expansion is the *only* grammar added: nested groups, multiple groups, and commas inside `[...]` are covered, and unbalanced braces are passed through so `globset` still reports the error. Added unit tests for `brace_expand` plus `a/{b/c,d}/e` and multi-group traversal cases.

**2. Asymmetry between include/exclude negation.** This one's a genuine judgment call, and it's yours to make. My argument for keeping it: the `!`-prefix here isn't a second negation API — it's the same affordance `.gitignore` gives, where `!` re-includes inside an otherwise-excluded subtree (`__pycache__/` + `!__pycache__/keep.py`). Users coming from `.gitignore`/`rsync`/`ripgrep` already read `!` as "but keep this", so it lands as one familiar rule rather than "negation of a negation". `excluded_patterns` being "the negation of includes" describes the field's role; the `!` is a per-entry exception *within* that field, not an inversion of the field's meaning. The motivating case (#1778: exclude `.github/**` but keep `copilot-instructions.md`) stays a one-liner.

That said — if you'd still prefer to model this structurally as `sub_matchers` (your 2026-04-30 sketch), I'm glad to close this and open a design issue for it; #1778 stays addressable either way. Just let me know which direction you'd like.

georgeh0:
Thanks a lot for your patience! The current version is great enough and  looks good to me.

Can you help to:
- Update the [documentation](https://github.com/cocoindex-io/cocoindex/blob/main/docs/src/content/docs/common_resources/data_types.mdx?plain=1).
- And run the precommit hooks to reformat the files (which has CI failure now)]

Thanks!

SuperMarioYL:
Thanks @georgeh0! Both done in 35adc013:

- **Docs** — added a *Gitignore-style negation* subsection to `FilePathMatcher` in `docs/src/content/docs/common_resources/data_types.mdx`, documenting `!`-prefixed patterns in `excluded_patterns` with the `.github` carve-out example and a note that traversal still descends into excluded dirs when a negation could match.
- **CI** — the `fast-check` failure was `cargo fmt` on the new test in `pattern_matcher.rs`; reformatted and pushed. `ruff format` was already clean.

Let me know if you'd like the negation behavior covered in the connector docs (localfs/amazon_s3) as well.

georgeh0:
Looks good! Thanks for making this!

PR #1908, opened 2026-04-28
feat(falkordb): add FalkorDB target connector

galshubeli:
Pure-Python connector under cocoindex.connectors.falkordb, modeled on the SurrealDB connector. Supports node tables, relation (edge) tables, per-graph multitenancy, and vector index attachments.

- _cypher.py: pure Cypher (MERGE upsert, DETACH DELETE, three-MERGE relationship insert with $-bound params, vector index DDL)
- _target.py: ConnectionFactory, TableSchema (with from_class + FalkorType override), TableTarget, RelationTarget, module-level entry points, root provider "cocoindex/falkordb/table"
- Two-level state via statediff.diff_composite, with prev_pk_field recovered from prior tracking records so DROP INDEX can identify the underlying artifact on table delete/replace
- Vector index attachment via attachments() (the plural form is what the engine actually introspects)
- pyproject.toml: falkordb=[falkordb>=1.1.0] extra + all=[...] + mypy ignore for falkordb.*
- 44 tests (38 unit, 6 integration vs live FalkorDB), mypy strict clean, docs build clean
- Docs page + sidebar entry under Connectors

### DIFF
--- a/docs/src/content/docs/connectors/falkordb.mdx
+++ b/docs/src/content/docs/connectors/falkordb.mdx
@@ -1,0 +1,389 @@
+---
+title: "*FalkorDB* connector"
+toc_max_heading_level: 4
+description: >
+  Write to FalkorDB — a Redis-backed graph database — with support for node
+  tables, relationship tables (edges), per-graph multitenancy, and vector
+  indexes with cosine / euclidean / inner-product distances.
+---
+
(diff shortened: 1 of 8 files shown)


georgeh0 (approved):
Thanks for the PR!

PR #1920, opened 2026-04-29
Feat: Cocoindex show full fledged inspection tool for debugging

### LINKED ISSUE #1830: [FEATURE] `cocoindex show` full fledged inspection tool for debugging
`cocoindex show` should support the following functionalities:

1. It should print out detailed information stored in the LMDB for each stable path (currently we only print the keys) under a flag `-l`

2. It should accept a stable path from argument. When provided, only information for this specific path. With the following flags
    - `-r`: if provided, also show all children for the stable path recursively.  e.g. if the path is `A/B/C`, information for `A/B/C/D` and `A/B/C/D/E` etc. will also be showed
    - `-p`: if provided, also show all parent for the stable path. e.g. if the path is `A/B/C`, information for `A` and `A/B` will also be showed
  

---

prithvi-moonshot:
- Extends CocoIndex show cli command as required. 
- Adds a LMDB query function to fetch stable path details 

Closes #1830


prithvi-moonshot:
@georgeh0 I believe the mypy type check error which is the only error I saw while running the pre-commit hooks locally is caused by an unrelated issue. It has something to do with ``testcontainers``.  Let me know if my assumption is incorrect and I will make changes.

badmonster0:
thanks a lot @prithvi-moonshot , @georgeh0  can help take a look!

python/cocoindex/_internal/core.pyi
```diff
@@ -183,6 +183,18 @@ class StablePathInfoAsyncIterator:
...
     def __anext__(self) -> Awaitable[StablePathInfo]: ...
 
+class StablePathDetail:
+    """Detailed information about a stable path from LMDB."""
+
+    path: StablePath
+    node_type: StablePathNodeType
+    version: int
+    processor_name: str
+    target_state_count: int
```
georgeh0, line 217: I'm thinking [`target_state_items`](https://github.com/cocoindex-io/cocoindex/blob/5674ad100a0c4ced214bcc103d2b15b9c6dcc8bb/rust/core/src/state/db_schema.rs#L344-L345) is more useful than the count.

Among these fields [https://github.com/cocoindex-io/cocoindex/blob/5674ad100a0c4ced214bcc103d2b15b9c6dcc8bb/rust/core/src/state/db_schema.rs#L271-L287]:
- `key` can be decoded into a `StableKey` and displayed.
- For `states`, we can display everything other than the specific bytes in the `TargetStateInfoItemState::Existing` enum variant (because deserialization for the bytes here needs specific target connector, and it's not very easy to locate them now).
- For others (`provider_schema_version`, `provider_generation`), we can also display.

Thanks!

python/cocoindex/cli.py
```diff
@@ -628,20 +714,145 @@ async def _show_from_database(db_path: str, app_name: str, tree: bool) -> None:
...
+) -> None:
+    """Print stable paths in table format with detailed info."""
+    if not items:
+        click.echo("Stable paths:")
+        click.echo("  (none)")
+        return
+
+    # Table header
+    click.echo(
+        f"{'PATH':<35} | {'TYPE':<10} | {'VERSION':<8} | {'PROCESSOR':<20} | {'TARGETS':<8} | {'MEMO':<5}"
```
georgeh0: I think it may not fit into a fixed-width table, e.g. path can be arbitrarily long, and there're more details.

I'm thinking about showing it in multiple lines, e.g. something like

```
/the/path/which/may/be/very/long
  {type} version:{version} processor:{processor} ...
  Target states:
    ......
```

georgeh0 (commented):
Sorry for the late review. Please see the comments.

python/cocoindex/_internal/core.pyi
prithvi-moonshot: Done

python/cocoindex/cli.py
prithvi-moonshot: Done

prithvi-moonshot:
Also I had to suppress the mypy error in order to commit my changes because I believe it is unrelated, but just wanna give a heads up

georgeh0:
Hi @prithvi-moonshot , there're `maturin build` failure in the CI, in `db_inspect.rs` which should be related to the PR. Can you have a look and fix? Thanks!

prithvi-moonshot:
@georgeh0 `maturin build` failure fixed

python/cocoindex/cli.py
```diff
@@ -647,20 +733,160 @@ async def _show_from_database(db_path: str, app_name: str, tree: bool) -> None:
...
+    items: AsyncIterator[Any],
+    app: App[Any, Any] | None = None,
+    tree: bool = False,
+    long_format: bool = False,
+    stable_path: str | None = None,
+    recursive: bool = False,
+    parents: bool = False,
+) -> None:
+    # Collect all items first for filtering
+    all_items = [item async for item in items]
```
georgeh0: We should avoid buffering all items in memory. Before the change, we're doing it in streaming way. We should keep it streaming.

python/cocoindex/cli.py
```diff
@@ -647,20 +733,160 @@ async def _show_from_database(db_path: str, app_name: str, tree: bool) -> None:
...
+        # Parents: item is a prefix of target
+        if parents:
+            if stable_path.startswith(item_path_str) and stable_path != item_path_str:
+                filtered.append(item)
+
+        # Children: target is a prefix of item
+        if recursive:
+            if item_path_str.startswith(stable_path) and item_path_str != stable_path:
+                if item not in filtered:
+                    filtered.append(item)
```
georgeh0: It's inefficient to scan everything and filter what we want.

- To get the item with a specific path, we can read entry for the specific stable path.
- To get all the children items, we can do a stable path prefix scan.
- To also get the parents, we can get parent stable path from the current path and read the entry for each one.

In all cases, we only need to read what we want.

rust/core/src/inspect/db_inspect.rs
```diff
@@ -56,3 +59,166 @@ fn receiver_to_stable_path_info_stream(
...
     env.storage().list_app_names().await
 }
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct TargetStateInfoItemSummary {
+    pub target_state_path: String,
+    pub key: String,
+    pub states: Vec<(u64, String)>,
+    pub provider_schema_version: u64,
+    pub provider_generation: Option<(u64, u64)>,
```
georgeh0: When reading the code here, for `states` and `provider_generation`, what each element of the tuple means is unclear. It'll be much clearer if we have structs with named fields for these.

Same for related fields in `PyTargetStateInfoItemSummary`.

rust/core/src/inspect/db_inspect.rs
```diff
@@ -56,3 +59,166 @@ fn receiver_to_stable_path_info_stream(
 pub async fn list_app_names<Prof: EngineProfile>(env: &Environment<Prof>) -> Result<Vec<String>> {
     env.storage().list_app_names().await
 }
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct TargetStateInfoItemSummary {
+    pub target_state_path: String,
+    pub key: String,
```
georgeh0: The type should be `StableKey` rather than a `String`. There's a `PyStableKey` type to be used under `rust/py` crate.

prithvi-moonshot: Done

rust/core/src/inspect/db_inspect.rs
prithvi-moonshot: Done

python/cocoindex/cli.py
prithvi-moonshot: Done

python/cocoindex/cli.py
prithvi-moonshot: Done

prithvi-moonshot:
@georgeh0 Link check failure seems unrelated to the changes I have pushed. But let me know if you think otherwise.

georgeh0 (approved):
Thanks for the contribution! merging.

PR #1945, opened 2026-05-06
Add preview mode for update actions

### LINKED ISSUE #1890: [FEATURE] support shadow run / preview
From users from Discord:

> One feature I’d genuinely love to see is a shadow run/preview mode. When someone changes chunking or something, can we somehow show what would actually change before anything is written?
This would reduce the breaking of pipelines in a sense

We want to support an option for `update` API and CLI subcommand to enable preview run. During preview run, all *Actions* will be:

- For API: emitted to a callback
- For CLI: printed out

instead of being applied to the real target.

---

MrAnayDongre:
## Summary

Adds preview mode for update actions in both the Python API and CLI.

With preview enabled, CocoIndex computes the same target actions as a normal update, but does not apply them to real targets and does not commit internal tracking state.

Closes #1890.

## API and CLI

Preview is supported in:

- `App.update(..., preview=True)`
- `App.update_blocking(..., preview=True)`
- `cocoindex update --preview`

For the Python API, preview returns the planned action objects.
For the CLI, preview prints the planned actions.

## Implementation

This reuses the existing planning path instead of introducing a separate planner.

In preview mode:

- reconciliation still runs through `pre_commit(...)`
- planning runs in a standalone write transaction
- that transaction is dropped without commit
- `sink.apply(...)` is skipped
- the later commit phase is skipped

This keeps preview aligned with normal update behavior while remaining read-only.

## First pass scope

This first pass supports flat / leaf target actions only.

If a target requires child target providers, preview fails clearly instead of returning partial output.

## CLI behavior

Added:

- `cocoindex update ... --preview`

## Tests

Added coverage for:

- sync preview returns actions without writing
- async preview returns actions without writing
- preview rejects child target providers
- CLI preview prints planned actions

## Validation

Ran locally:

- `cargo fmt --check`
- `cargo test`
- `uv run mypy`
- `uv run pytest python/tests/core -q`
- `uv run pytest python/tests/internal -q`
- `uv run pytest python/tests/cli/test_cli.py -q`

### DIFF
--- a/python/tests/cli/test_cli.py
+++ b/python/tests/cli/test_cli.py
@@ -707,6 +707,32 @@ class TestDropQuiet:
 # =============================================================================
 
 
+class TestPreview:
+    """Tests for the --preview flag on update."""
+
+    def test_preview_prints_actions(self) -> None:
+        """update --preview should print planned actions without writing."""
+        result = run_cli("update", "./flat_target_app.py", "--preview")
+        assert "Preview: planned target actions" in result.stdout
+        assert "('x', 42)" in result.stdout
+
+    def test_preview_reset_rejected(self) -> None:
+        """--preview --reset should be rejected."""
+        result = run_cli(
+            "update", "./single_app.py", "--preview", "--reset", check=False
+        )
+        assert result.returncode != 0
+        assert "cannot be used together" in result.stderr.lower()
+
+    def test_preview_live_rejected(self) -> None:
+        """--preview --live should be rejected."""
+        result = run_cli(
+            "update", "./single_app.py", "--preview", "--live", check=False
+        )
+        assert result.returncode != 0
+        assert "cannot be used together" in result.stderr.lower()
+
+
 class TestShowTree:
     """Tests for the show command with --tree flag."""
 
--- a/rust/core/src/engine/app.rs
+++ b/rust/core/src/engine/app.rs
@@ -3,7 +3,7 @@ use crate::engine::stats::{ProcessingStats, VersionedProcessingStats};
 use crate::prelude::*;
 
 use crate::engine::component::Component;
-use crate::engine::context::AppContext;
+use crate::engine::context::{AppContext, PreviewActionCollector};
 use crate::engine::live_component::{LIVE_COMPONENT_DRAIN_TIMEOUT_SECS, LiveComponentState};
 
 use crate::engine::environment::{AppRegistration, Environment};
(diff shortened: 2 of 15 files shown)


badmonster0:

python/tests/cli/test_cli.py
```diff
@@ -690,6 +690,31 @@ def test_drop_quiet_suppresses_informational_output(self) -> None:
 # =============================================================================
 
 
+class TestPreview:
+    """Tests for the --preview flag on update."""
+
+    def test_preview_prints_actions(self) -> None:
+        """update --preview should print planned actions without writing."""
+        result = run_cli("update", "./flat_target_app.py", "--preview")
+        assert "Preview: planned target actions" in result.stdout
```
georgeh0, line 716: Can we also check the specific actions really printed out in the output?

rust/core/src/engine/app.rs
```diff
@@ -98,18 +100,28 @@ impl<Prof: EngineProfile> App<Prof> {
...
+        } else {
+            None
+        };
         let context = self.root_component.new_processor_context_for_build(
             None,
             processing_stats.clone(),
             options.full_reprocess,
             options.live,
+            options.preview,
+            preview_collector.clone(),
```
georgeh0: It's possible to combine these two, i.e. everywhere we can use `preview_collector.is_some()` to determine if it's under preview mode.

We can even consider using `Option<PreviewActionCollector<Prof>>` to replace the `preview: bool` field in `AppUpdateOptions`. The `update()` API will remain returning a single handle (`PreviewCollector` is more like a side channel output, so keeping the return value a simple handle is attractive).

georgeh0 (commented):
Sorry for the late review!

Most parts look good. A few minor comments. Please also merge with latest and resolve conflict. Thanks!

rust/core/src/engine/app.rs
MrAnayDongre: Thanks @georgeh0, I pushed an update for the review comments.

The CLI preview test now checks the actual planned action in the output, not just the header. I also cleaned up the Rust side so preview mode is derived from the preview collector instead of carrying a separate preview flag through the engine.

I rebased this on the latest `main` as well. While doing the cleanup, I kept the preview path read-only and reran the focused Rust/Python checks.

georgeh0 (approved):
Thanks a lot for the contribution!

PR #1946, opened 2026-05-06
fix: resolve ruff format and lint issues (C408)

dashitongzhi:
## Summary

Fixes ruff formatting and lint issues found in the codebase:

1. **ruff format** in `python/tests/connectors/test_kafka_target.py`: Lambda expressions with multiline string concatenation were not properly parenthesized, causing `ruff format` to flag the file.

2. **ruff C408** in `python/cocoindex/setting.py` and `python/cocoindex/_internal/setting.py`: Replaced 6 unnecessary `dict()` calls with dict literal `{}` syntax, which is more idiomatic and consistent with the rest of the codebase.

## Changes

- `python/tests/connectors/test_kafka_target.py`: Wrap lambda expressions in parentheses for proper multiline formatting
- `python/cocoindex/setting.py`: `dict()` → `{}` in `Settings.from_env()` and `ServerSettings.from_env()`
- `python/cocoindex/_internal/setting.py`: Same `dict()` → `{}` fixes in the internal settings module

## Verification

```bash
ruff format --check .        # 219 files already formatted ✓
ruff check --select C408 .   # All checks passed ✓
```

### DIFF
--- a/python/cocoindex/_internal/setting.py
+++ b/python/cocoindex/_internal/setting.py
@@ -65,7 +65,7 @@ class Settings:
     def from_env(cls, db_path: os.PathLike[str] | None = None) -> Self:
         """Load settings from environment variables."""
 
-        exec_kwargs: dict[str, Any] = dict()
+        exec_kwargs: dict[str, Any] = {}
         _load_field(
             exec_kwargs,
             "source_max_inflight_rows",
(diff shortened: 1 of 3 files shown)


georgeh0:
Thanks for the change!

Issue #1435, opened 2025-12-28
[FEATURE] Support adding new columns in place for LanceDB


georgeh0:
When users want to add a new column, they only need to update the flow. When rerunning the code with `cocoindex update`, cocoindex will automatically update the schema of the target. 

This is what we do for most targets (e.g. postgres, kuzu, etc.).

For LanceDB target, there was some problem when I tried to support adding columns to existing table (I have [the code](https://github.com/cocoindex-io/cocoindex/blob/175c7a0bc5bd7322c036ac432fb1a1ccb90bd114/python/cocoindex/targets/lancedb.py#L260-L288) but commented out, because hit some errors somewhere). So now when a new column is added, we always drop the table and recreate again.

We want to have another try to make adding columns in place work for LanceDB.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

prrao87:
Hi @georgeh0 based on this [example file](https://github.com/lancedb/cocoindex-lancedb-demo/blob/main/main.py) and the fact that it runs well (while updating the LanceDB database schema as expected via `collect()`), do we still consider this to be an issue?

georgeh0:
Hi @prrao87, in most cases it doesn't matter, because an auto backfill will happen after "drop and recreate". But it's still not ideal. For example, if there're some other processes using the table in between, they'll get problem before backfill is done. So I'll still keep this issue open.

prrao87:
Related issue on the Lance repo:
https://github.com/lance-format/lance/issues/3443

qWaitCrypto:
@georgeh0 I’d like to work on this.
My planned first pass is to support adding new columns in place for LanceDB table targets, while keeping destructive/
lossy schema changes out of scope for now. I’ll also add tests to verify existing rows survive add-column updates.

PR #1951, opened 2026-05-08
feat(lancedb): add columns in place for table targets

qWaitCrypto:
Addresses #1435.

## Summary
- Add in-place additive schema evolution for LanceDB table targets.
- When a flow adds new collected fields, the LanceDB target now uses `AsyncTable.add_columns(...)` instead of forcing a drop/recreate cycle.
- Keep the implementation intentionally narrow: only additive column changes are handled in place. Rename, drop, and type-change paths still fall back to full table replacement.

## Motivation
- CocoIndex already supports incremental target maintenance for relational targets such as Postgres and SQLite.
- LanceDB table targets previously lacked an in-place path for additive schema evolution, so adding a new collected column could force a destructive table rebuild or fail to recover cleanly in edge cases.
- This is especially noticeable for embedding/indexing workloads where users want to add metadata columns without reinitializing the whole table.

## What Changed
- Teach the LanceDB target reconciler to distinguish additive column actions from non-additive schema changes.
- Apply additive `insert` / `upsert` column actions via `AsyncTable.add_columns(...)`.
- Materialize newly added columns as `nullable=True` on the LanceDB side, since existing rows must backfill with `null`.
- Skip already-existing physical columns during additive reconciliation so recovery paths do not fail when tracking state is missing but the table schema is already ahead.
- Upgrade non-additive column actions to full table replacement during reconcile, instead of letting them reach an unsupported in-place apply path.
- Normalize table existence checks behind a small helper that handles LanceDB API variants consistently.

### Before
- Adding a new collected field to a LanceDB table target did not have a dedicated in-place schema evolution path.
- Recovery-style `upsert` column actions could fail.
- Adding a non-nullable field could fail because LanceDB requires newly added columns for existing rows to be nullable.
- Non-additive column actions had incomplete semantics between reconcile and apply.

### After
- Additive schema changes are applied in place.
- Existing rows survive column addition.
- Recovery-style `upsert` column actions are handled safely.
- Non-nullable logical fields are materialized as nullable new columns in LanceDB, which matches the required backfill semantics.
- Non-additive schema changes cleanly fall back to destructive replacement.

### Reproducing the pre-fix failure modes
These are the concrete scenarios covered by the new tests. On the old implementation, at least some of them would fail:

- Add a new non-nullable field after the table already exists.
- Add multiple columns in one schema evolution step.
- Reconcile an additive column change through an `upsert` action when tracking state is behind the physical table schema.

### Verifying the fix
Run:

```bash
uv run --group format ruff check \
  python/cocoindex/connectors/lancedb/_target.py \
  python/tests/connectors/test_lancedb_target.py
```

```bash
env -u CONDA_PREFIX PATH="$HOME/.cargo/bin:$PATH" \
  uv run pytest -q -p no:cacheprovider \
  python/tests/connectors/test_lancedb_target.py
```

Optional regression check against the earlier incremental convergence work:

```bash
env -u CONDA_PREFIX PATH="$HOME/.cargo/bin:$PATH" \
  uv run pytest -q -p no:cacheprovider \
  python/tests/core/test_incremental_convergence.py
```

## Test Coverage
- `test_add_column_preserves_existing_rows`
  - Verifies that adding a column preserves existing rows and advances the LanceDB table version in-place instead of resetting via drop/recreate.
- `test_add_column_keeps_old_rows_before_backfill`
  - Verifies additive schema evolution works before all rows have been fully backfilled with the new column.
- `test_add_non_nullable_column_is_materialized_as_nullable`
  - Covers the LanceDB requirement that newly added columns over existing data must be nullable.
- `test_add_multiple_columns_in_place`
  - Verifies more than one column can be added in a single additive schema evolution step.
- `test_lancedb_async_table_supports_add_columns_api`
  - Guards the connector assumption that the async LanceDB table exposes `add_columns`.

## Local Results
- `uv run --group format ruff check python/cocoindex/connectors/lancedb/_target.py python/tests/connectors/test_lancedb_target.py`
- `env -u CONDA_PREFIX PATH="$HOME/.cargo/bin:$PATH" uv run pytest -q -p no:cacheprovider python/tests/connectors/test_lancedb_target.py`
- `env -u CONDA_PREFIX PATH="$HOME/.cargo/bin:$PATH" uv run pytest -q -p no:cacheprovider python/tests/core/test_incremental_convergence.py`

Observed locally:
- `python/tests/connectors/test_lancedb_target.py`: `5 passed`
- `python/tests/core/test_incremental_convergence.py`: `3 passed`

## Scope / Non-Goals
- No attempt to support in-place rename, drop, or type mutation for LanceDB columns in this PR.
- No changes to the generic incremental engine or other target connectors.
- No schema migration UI / CLI work; this PR is strictly connector behavior plus regression coverage.


prrao87:
Thanks a lot for this update! LGTM overall. Should be useful to folks who need to update their tables! 

My only concern with this is that each incremental add creates a new fragment in the Lance table. Over time, this can add up in the manifest. @georgeh0 @cocoindex-dev I know we briefly discussed this earlier but do you see scope for cocoindex automatically calling `table.optimize()` every fixed number of commits (say 500 or 1000) to perform compaction? This will avoid unnecessary surprises in performance for the user.

georgeh0 (approved):
Thanks for adding in-place column addition support!

badmonster0:
@qWaitCrypto pls take a look at the test.

@prrao87 thanks for the suggestion! maybe we can have a separate issue so we don't lose track of your suggestion?

qWaitCrypto:
Thanks for the heads up! I’ve updated the test typing in the latest commit. @badmonster0 

Regarding the issue mentioned above by @prrao87 , if needed, I can also submit a follow-up PR to address it.

prrao87:
Thanks @qWaitCrypto . 

With regard to compaction, I believe this is already accounted for in v0 of CocoIndex: https://github.com/cocoindex-io/cocoindex/issues/1429

When you have a moment, could you give it a check that the v1 API also calls `table.optimize()` at the required cadence per the PR linked there? Thanks!

qWaitCrypto:
> Thanks @qWaitCrypto .
> 
> With regard to compaction, I believe this is already accounted for in v0 of CocoIndex: #1429
> 
> When you have a moment, could you give it a check that the v1 API also calls `table.optimize()` at the required cadence per the PR linked there? Thanks!

I checked the v1 LanceDB target, and it doesn't seem to have the v0 `table.optimize()` cadence yet.

prrao87:
I created https://github.com/cocoindex-io/cocoindex/issues/2002 to track this, thanks!

qWaitCrypto:
@georgeh0  I just resolved the conflicts, so it should be ready to merge directly.

PR #1953, opened 2026-05-09
docs(ops/text): expand supported languages table for RecursiveSplitter

nuthalapativarun:
## Summary

- Replace the vague "And many more..." placeholder in the `RecursiveSplitter` docs with a complete table of all tree-sitter-backed languages
- Each entry shows the `language=` string to pass and the file extensions it covers
- Add a note that many more languages use separator-based splitting and that `detect_code_language()` can infer the language name from a filename

Solidity (`.sol`) was recently added (#1942 added Julia, earlier PRs added Svelte/Vue) but was not reflected in the docs.

## Test plan

- `uv run ruff format --check .`
- `uv run ruff check .`

### DIFF
--- a/docs/src/content/docs/ops/text.mdx
+++ b/docs/src/content/docs/ops/text.mdx
@@ -115,11 +115,43 @@ chunks = splitter.split(
 ```
 
 **Supported languages:**
-- Python
-- Rust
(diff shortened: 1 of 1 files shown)


georgeh0:
Thanks for the PR!

PR #1960, opened 2026-05-10
fix: correct lychee exit code handling

Haleshot:
Fixes the [link checker workflow](https://github.com/cocoindex-io/cocoindex/blob/main/.github/workflows/links.yml) that was silently passing despite broken links in the [latest run](https://github.com/cocoindex-io/cocoindex/actions/runs/25616104962).

Turned out there were a few compounding issues: `lychee` actually [exits with 2 for broken links](https://github.com/lycheeverse/lychee#exit-codes) (not 1), so the old `== 1` condition never _fired_. The issue creation step also needed to come **before** the fail step, otherwise it gets _skipped_.

And scheduled runs were restoring a stale cache, so URLs that were valid before the `v1` branch was deleted kept passing on daily checks; fixed by skipping cache restore on scheduled runs entirely. Also cleaned up v1 from the branch triggers since that branch no longer exists :)

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -2,14 +2,14 @@ name: Link Check
 
 on:
   pull_request:
-    branches: [main, v1]
+    branches: [main]
     paths:
       - "docs/**/*.md"
       - "docs/**/*.mdx"
       - ".lycheeignore"
       - ".github/workflows/links.yml"
   push:
-    branches: [main, v1]
+    branches: [main]
     paths:
       - "docs/**/*.md"
       - "docs/**/*.mdx"
@@ -33,6 +33,7 @@ jobs:
           fetch-depth: 1
 
       - name: Restore lychee cache
+        if: github.event_name != 'schedule'
         uses: actions/cache@v5
         with:
           path: .lycheecache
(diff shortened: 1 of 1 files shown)


Haleshot:
The CI fail (expected) is now correctly catching the broken v1 links still in main. Will go green once the doc-fix PR merges: #1959

.github/workflows/links.yml
```diff
@@ -2,14 +2,14 @@ name: Link Check
 
 on:
   pull_request:
-    branches: [main, v1]
+    branches: [main]
```
georgeh0, line 5: Can we add `v0` here?

georgeh0 (commented):
Thanks for the fix!

.github/workflows/links.yml
Haleshot: Do you still accept changes pushed to `v0`? 

If so, I can add `v0` to both the `pull_request` and `push` branch filters. If it is archived/frozen, keeping only `main` seems cleaner?

badmonster0: @georgeh0 may take a look?

Haleshot: Sure, let me know :)

PR #1964, opened 2026-05-11
Security: No Input Validation on File Path Resolution

tuanaiseo:
## Problem

In python/cocoindex/connectors/localfs/_common.py, the FilePath.resolve() method resolves paths without sanitization. If user-controlled input flows to FilePath, this could lead to path traversal attacks.

**Severity**: `medium`
**File**: `python/cocoindex/connectors/localfs/_common.py`

## Solution

Add path validation to ensure resolved paths stay within expected boundaries. Use os.path.realpath() and validate the result is within allowed directories.

## Changes

- `python/cocoindex/connectors/localfs/_common.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/python/cocoindex/connectors/localfs/_common.py
+++ b/python/cocoindex/connectors/localfs/_common.py
@@ -54,12 +54,24 @@ class FilePath(file.FilePath[pathlib.Path]):
 
     def resolve(self) -> pathlib.Path:
         """Resolve this FilePath to an absolute filesystem path."""
+        import os
+
         if self._base_dir is not None:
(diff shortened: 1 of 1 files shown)


georgeh0:
Thanks for creating the PR!

I'm thinking restricting the resolved directory within the base directory is not always the expected behavior.

Maybe we can consider add the restricted base dir as an explicit kwarg to pass in. What do you think?

PR #1969, opened 2026-05-12
feat: add Apache Iggy connector

countradooku:
## Summary

- add an Apache Iggy connector package with source and target support
- map Iggy stream/topic/partition addressing into CocoIndex live streams and target states
- preserve source at-least-once behavior by storing Iggy offsets only after downstream readiness
- add mocked source tests covering offset storage, payload streams, keyed map adaptation, and readiness safeguards
- add the `cocoindex[iggy]` optional dependency and lockfile entries

## Notes

- Iggy Python messages do not expose Kafka-style keys or tombstones, so `topic_as_map()` requires an application-level `key` extractor and optional deletion predicate.
- Multi-partition readiness requires an explicit `initial_high_watermark` until the Python SDK exposes per-partition watermarks.
- Live testing against `apache/iggy:latest` found that `iter_messages().__anext__()` returns a Future, so the connector uses `asyncio.ensure_future()` for SDK compatibility.

## Validation

- `uv run pytest python/tests/connectors/test_iggy_source.py`
- `uv run ruff check python/cocoindex/connectors/iggy/_source.py python/cocoindex/connectors/iggy/_target.py python/tests/connectors/test_iggy_source.py`
- `uv run mypy python/cocoindex/connectors/iggy/_source.py python/cocoindex/connectors/iggy/_target.py python/tests/connectors/test_iggy_source.py`
- Live smoke test against local `apache/iggy:latest` on `127.0.0.1:8090`: created stream/topic, sent via Iggy target handler, consumed via `topic_as_stream()`, observed readiness, and verified the same consumer group did not replay messages after offset storage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,7 @@ doris = ["aiohttp>=3.9.0", "pymysql>=1.1.0", "aiomysql>=0.2.0"]
 falkordb = ["falkordb>=1.1.0"]
 neo4j = ["neo4j>=5.18.0"]
 kafka = ["confluent_kafka>=2.6"]
+iggy = ["apache-iggy>=0.8.0"]
 oci = ["oci>=2.0"]
 entity_resolution = ["faiss-cpu>=1.7"]
 entity_resolution_llm = [
@@ -120,6 +121,7 @@ all = [
     "falkordb>=1.1.0",
     "neo4j>=5.18.0",
     "confluent_kafka>=2.6",
+    "apache-iggy>=0.8.0",
     "oci>=2.0",
     "faiss-cpu>=1.7",
     "instructor>=1.0",
--- a/python/cocoindex/connectors/iggy/__init__.py
+++ b/python/cocoindex/connectors/iggy/__init__.py
@@ -1,0 +1,5 @@
+from . import _source, _target
+from ._source import *
+from ._target import *
+
+__all__ = _source.__all__ + _target.__all__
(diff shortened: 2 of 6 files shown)


badmonster0:
@countradooku looks good, pls take a look at the test. what kind of e2e test did you run?

countradooku:
> @countradooku looks good, pls take a look at the test. what kind of e2e test did you run?

I reran against a real `apache/iggy:latest` Docker broker over TCP. That caught a live-only duplicate offset case (`0,1,1,2`) from the Iggy consumer while manual offset storage was catching up. I fixed the source to skip duplicate offsets within a watch call and added a regression test for that sequence. After the fix, the real broker smoke delivered exactly `0,1,2`, and `uv run prek run --all-files` passes.

badmonster0:
thanks a lot @countradooku . could you follow this example https://cocoindex.io/blogs/csv-to-kafka-live/
and make sure iggy works for it too? just want to make sure the examples we bring to cocoindex community are well tested

countradooku:
> thanks a lot @countradooku . could you follow this example https://cocoindex.io/blogs/csv-to-kafka-live/ and make sure iggy works for it too? just want to make sure the examples we bring to cocoindex community are well tested

@badmonster0 I ran the blog’s live CSV workflow against Iggy, using a real `apache/iggy:latest` broker over TCP. The live app used `mount_iggy_topic_target`, watched CSV files, and emitted JSON upsert/delete messages. I verified initial load, one-cell edit, row add, row delete, and file delete. The run passed: initial scan produced 9 upserts, the edit/add/delete operations each produced only the expected diff messages, and deleting `products.csv` emitted delete markers for the six product keys. Since Iggy does not support Kafka tombstones, deletes were encoded with `deletion_value_fn`.

badmonster0:
amazing, thanks!!

Issue #1060, opened 2025-09-28
Enhance error message population from Rust side to Python SDK


georgeh0:
Currently we're exposing too concise error info to users when an error occurs from Rust side, especially when `RUST_BACKTRACE` is not turned on.

For example, users are getting this error message, which is very hard to debug:

```
Error:

  File "/home/vivek/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/home/vivek/Data/experiement_projects/invoice_chatbot/venv/lib/python3.11/site-packages/cocoindex/flow.py", line 832, in setup_async
    await bundle.describe_and_apply_async(report_to_stdout=report_to_stdout)
  File "/home/vivek/Data/experiement_projects/invoice_chatbot/venv/lib/python3.11/site-packages/cocoindex/setup.py", line 72, in describe_and_apply_async
    await self.apply_async(report_to_stdout=report_to_stdout)
  File "/home/vivek/Data/experiement_projects/invoice_chatbot/venv/lib/python3.11/site-packages/cocoindex/setup.py", line 37, in apply_async
    await self._engine_bundle.apply_async(report_to_stdout=report_to_stdout)
Exception: version 1213486160 is not supported

```

See sub issues for detailed actions.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

CodeSakshamY:
I am working on it

thisisharsh7:
@georgeh0 can you please guide me a bit on this issue. Do we need to improve the error message?

georgeh0:
> [@georgeh0](https://github.com/georgeh0) can you please guide me a bit on this issue. Do we need to improve the error message?

This is a parent issue. The specific action items in my mind are created as sub issues.

nuthalapativarun:
[...]

PR #1986, opened 2026-05-17
fix(py): add context to Rust→Python error messages for App operations

nuthalapativarun:
## Summary

Wraps `App::new()`, `app.update()`, and `app.drop_app()` calls at the PyO3 boundary with `.context()` messages so terse internal Rust errors include the name of the operation that triggered them.

**Before:**
```
Exception: version 1213486160 is not supported
```

**After:**
```
Exception:
Context:
  1: failed to initialize app 'my_app'
version 1213486160 is not supported
```

Uses the existing `ContextExt::context()` infrastructure already in place in `cocoindex_utils::error`. No new dependencies, no API surface changes.

Part of #1060.

## Test plan

- `cargo test -p cocoindex_core -p cocoindex_utils -p cocoindex_ops_text` — all pass
- `uv run maturin develop --uv` — builds cleanly
- `uv run pytest python/tests/ -q` — 625 passed, 138 skipped

### DIFF
--- a/rust/py/src/app.rs
+++ b/rust/py/src/app.rs
@@ -170,7 +170,9 @@ impl PyApp {
         env: &PyEnvironment,
         max_inflight_components: Option<usize>,
     ) -> PyResult<Self> {
-        let app = App::new(name, env.0.clone(), max_inflight_components).into_py_result()?;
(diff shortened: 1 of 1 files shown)


georgeh0 (approved):
Thanks for improving error messages!

PR #2005, opened 2026-05-23
fix(id): make generate_uuid() return stable UUIDs matching memo=True contract

dashitongzhi:
The @memo=True decorator promises idempotent results but uuid4() was generating random UUIDs. This fix uses uuid5() with memo fingerprints for deterministic UUIDs.

### DIFF
--- a/python/cocoindex/resources/id.py
+++ b/python/cocoindex/resources/id.py
@@ -57,6 +57,15 @@ async def generate_id(_dep: _typing.Any = None) -> int:
     return await _component_ctx.next_id(None)
 
 
+# Namespace UUID for stable UUID generation (using UUID v5 algorithm)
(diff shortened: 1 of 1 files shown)


georgeh0 (commented):
Thanks a lot for creating the fix!

Note that [`generate_uuid()`](https://github.com/cocoindex-io/cocoindex/blob/a8d016211c5cb19be58d68264e892a8241fe769c/python/cocoindex/resources/id.py#L65-L66) is supposed to guarantee the UUID is the same within the same component (it's already guaranteed by `memo=True`), and will not be the thing across components: we'll generate distinct IDs in the case.

One primary use case for this is to generate stable target primary key, and avoid churns (i.e. no-op or in-place update instead of delete+insert). If it's already in different components, we cannot avoid churning any way. Also in general we want to avoid different component owns the same target primary key.

May I understand your use case? It'll be helpful for us to revisit if our design and assumption is sound. Thanks!


BTW, I think our current documentation is not super clear on this. We'll improve that.

PR #2006, opened 2026-05-23
Parallelize entity resolution via candidate-graph components

zherendong:
## Summary

- Partitions entities into connected components of the candidate-similarity graph (FAISS `range_search` + union-find), then resolves each component independently via `asyncio.gather`. Components are vertex-disjoint by construction, so dedup-map writes never conflict and the output is byte-for-byte identical to the prior sequential implementation. Resolver-call concurrency goes from 1 to (≈ component count).
- Event-callback ordering is preserved exactly: per-component runners collect events locally, then `resolve_entities` emits them in the same per-policy sort order the single-pass implementation used (PREFERRED: globally sorted; PINNED: pass-1 existings first then pass-2 non-existings, each sorted).
- Adds the benchmark scaffolding needed to actually *measure* the gain — `BenchmarkProfile` system, asymmetric `cluster_sizes` knob, `candidate_components` metric (range_search + union-find over captured embeddings, computed post-hoc inside the benchmark), and the `synthetic-many-components` / `synthetic-one-giant` / `openai-many-components` profiles.
- Follow-up fixes preserve the sequential edge-case semantics under parallel dispatch: exact-threshold candidate edges stay in the same component, resolver failures cancel sibling component work, already-completed events are still delivered before reraising, `ResolvedEntities` iteration remains deterministic, `top_n` remains a hard cap on surfaced candidates, and PINNED event ordering now tracks pass phases explicitly.

## Why `range_search` instead of `top-N` for graph edges

`top_n` caps the number of distinct canonical candidates surfaced to the resolver, but `_CandidateIndex.search` may inspect more than `top_n` raw vector hits to backfill after chain-walking aliases to canonicals. Building the partition graph from only top-N raw hits could miss threshold edges the runtime search would later surface, silently changing behavior. `range_search` builds the full threshold graph instead: extra edges only reduce parallelism, while missing edges can change correctness.

One subtlety: FAISS `IndexFlatIP.range_search` uses a strict `score > radius` boundary, while runtime candidate filtering keeps `score >= threshold`. The partition and benchmark component counter step the radius one float32 below the threshold so exact-boundary candidate edges are included.

## Measured Speedups

Synthetic (20 ms simulated resolver latency):

| Profile | Components | Before | After | Speedup |
|---|---|---|---|---|
| `synthetic-latency` | 200 | 6397 ms | 69 ms | **93×** |
| `synthetic-many-components` | 180 | 2448 ms | 47 ms | **52×** |
| `synthetic-one-giant` | 1 | 2110 ms | 2111 ms | 1.0× (no regression) |

Real OpenAI (`openai/gpt-5.4-nano` + `text-embedding-3-small`):

| Profile | `max_distance` | Components | Before | After | Speedup |
|---|---|---|---|---|---|
| `openai-small` | 0.3 | 2 | 10513 ms | 8691 ms | 1.21× |
| `openai-many-components` | 0.15 | 15 | 15063 ms | 3075 ms | **4.90×** |

`wrong_mixed_groups = 0` and canonical counts identical to the sequential implementation in every profile. Note: real-world speedup is bounded by `candidate_components`, which depends on both `max_distance` and how clustery the embedder is for your data — see the README in `benchmarks/entity_resolution/` for the threshold-sensitivity discussion.

## Change Structure

The branch is organized as a benchmark-first implementation with follow-up correctness commits:

1. Benchmark scaffolding: deterministic synthetic profiles, OpenAI-backed profiles, `cluster_sizes`, post-hoc `candidate_components`, and README numbers.
2. Candidate search improvements: backfill for distinct canonicals after chain-walking aliases, while preserving `top_n` as the maximum surfaced candidate count.
3. Refactors: extract `_CandidateIndex`, helper functions, event builders, and `_resolve_component` so the parallel dispatch change stays reviewable.
4. Parallel dispatch: partition entities by the conservative threshold graph and resolve independent components concurrently, with parity and concurrency tests.
5. Event/API semantics: document post-resolution event delivery order, keep PINNED phase ordering explicit, preserve deterministic `ResolvedEntities` iteration, and deliver completed partial events on resolver failure.
6. Error/correctness fixes: include exact-threshold graph edges, cancel sibling component tasks on resolver error, and keep benchmark component counting aligned with resolver partitioning.

## Test Plan

- [x] `uv run pytest python/tests/ops/test_entity_resolution.py python/tests/ops/test_llm_pair_resolver.py` — 42 passed
- [x] `uv run mypy python/cocoindex/ops/entity_resolution python/tests/ops/test_entity_resolution.py benchmarks/entity_resolution/benchmark.py` — clean
- [x] `uv run ruff check benchmarks/entity_resolution/benchmark.py python/cocoindex/ops/entity_resolution python/tests/ops/test_entity_resolution.py` — clean
- [x] `uv run ruff format --check benchmarks/entity_resolution/benchmark.py python/cocoindex/ops/entity_resolution python/tests/ops/test_entity_resolution.py` — clean
- [x] All four synthetic benchmark profiles match the README numbers within wall-clock jitter
- [x] Both OpenAI benchmark profiles produce identical structural metrics (components, canonicals, resolver_calls, max_concurrency, wrong_mixed_groups) to the README values
- [x] **Parity check**: a parametrized test runs every scenario through both the real component-graph dispatch and a monkey-patched single-component dispatch, asserting byte-identical dedup maps. The single-component dispatch is exactly the pre-partitioning sequential code, so equality is a strong correctness oracle.
- [x] **Concurrency check**: a dedicated test confirms peak in-flight resolver calls reaches ≥ 2 on disjoint components

### DIFF
--- a/benchmarks/entity_resolution/.env.example
+++ b/benchmarks/entity_resolution/.env.example
@@ -1,0 +1,21 @@
+# Copy to .env and fill in your real key. .env is intentionally gitignored.
+OPENAI_API_KEY=
+
+# Defaults for OpenAI-backed benchmark runs. Any CLI flag still overrides these.
+ER_BENCH_PROFILE=openai-small
+ER_BENCH_EMBEDDER=litellm
+ER_BENCH_RESOLVER=llm
+ER_BENCH_EMBEDDING_MODEL=text-embedding-3-small
+ER_BENCH_LLM_MODEL=openai/gpt-4o-mini
+ER_BENCH_ENTITY_TYPE=organization
+
+# Keep real LLM runs small at first; resolver calls are the main cost driver.
+ER_BENCH_GROUPS=20
+ER_BENCH_ALIASES_PER_GROUP=3
+ER_BENCH_ISOLATED=20
+ER_BENCH_MAX_DISTANCE=0.3
+ER_BENCH_TOP_N=5
+
+# Optional output/state overrides.
+# ER_BENCH_STATE=benchmarks/entity_resolution/.work/cocoindex
+# ER_BENCH_EXTRA_GUIDANCE=A parent organization and subsidiary are distinct.
--- a/benchmarks/entity_resolution/.gitignore
+++ b/benchmarks/entity_resolution/.gitignore
@@ -1,0 +1,3 @@
+.venv/
+.work/
+.env
--- a/benchmarks/entity_resolution/README.md
+++ b/benchmarks/entity_resolution/README.md
@@ -24,14 +24,43 @@ uv run --project benchmarks/entity_resolution python benchmarks/entity_resolutio
   --output-json benchmarks/entity_resolution/.work/synthetic-fast.json
 ```
 
-Simulate resolver latency to make the current sequential pair processing
-visible:
+Simulate resolver latency to make resolver-call latency the dominant cost:
 
 ```sh
 uv run --project benchmarks/entity_resolution python benchmarks/entity_resolution/benchmark.py \
   --profile synthetic-latency \
   --output-json benchmarks/entity_resolution/.work/synthetic-latency.json
 ```
+
+## Component parallelism
+
+`resolve_entities` partitions entities into connected components of the
+candidate-similarity graph and resolves each component concurrently. Two
+profiles measure the two extremes:
+
+```sh
+# Many independent clusters → high resolver-call concurrency.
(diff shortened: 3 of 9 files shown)


zherendong:
@georgeh0 looks like I don't have permissions to add reviewers so pinging here. This pr should be ready for review.

georgeh0 (approved):
Thanks for the optimizations and adding the benchmark!

PR #2008, opened 2026-05-23
feat(lancedb): optimize v1 target periodically

### LINKED ISSUE #2002: [FEATURE]: Add compaction and table optimization step in LanceDB target
**What is the use case?**

This is a follow-up to https://github.com/cocoindex-io/cocoindex/issues/1429: That was for the v0 LanceDB target. In v1, there doesn't seem to be a periodic `table.optimize()` call that keeps the number of fragments in the written data manageable. Not calling `table.optimize()` can result in a large number of fragments building up over time if users tend to make writes in small batches (or incrementally).

A similar solution to what was implemented for v0 in https://github.com/cocoindex-io/cocoindex/pull/1466 would work for v1, too (though the trigger to optimize may depend on the sequence of steps in the CocoIndex v1 pipeline).

**Additional context**

Tagging @qWaitCrypto and @cocoindex-dev here for any additional pointers and tips, if needed.

I do not have the bandwidth in the near term to take this on, but if @qWaitCrypto wants to, I'd be more than happy to watch and cheer on! 😁

qWaitCrypto:
## Summary

Add periodic LanceDB table optimization to the v1 target connector, matching the v0 behavior added in #1466.

- Adds `num_transactions_before_optimize` to the v1 LanceDB table target APIs, defaulting to 50.
- Schedules background `table.optimize()` calls after successful row mutation batches reach the configured threshold.
- Documents the new option and adds unit coverage for the cadence and parameter validation.

## Validation

- `ruff check python/cocoindex/connectors/lancedb/_target.py python/tests/connectors/test_lancedb_target.py`
- `python -m py_compile python/cocoindex/connectors/lancedb/_target.py python/tests/connectors/test_lancedb_target.py`
- `git diff --check`

Targeted pytest with `uv run --group build-test --extra lancedb pytest python/tests/connectors/test_lancedb_target.py -q` could not complete locally because dependency downloads for large wheels such as `lancedb`/`pyarrow` repeatedly stalled or timed out in this environment.

Fixes #2002

### DIFF
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -80,6 +80,7 @@ def declare_table_target(
     table_schema: TableSchema[RowT],
     *,
     managed_by: Literal["system", "user"] = "system",
+    num_transactions_before_optimize: int = 50,
 ) -> TableTarget[RowT, coco.PendingS]
 ```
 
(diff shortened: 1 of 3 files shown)


georgeh0:
Thanks for backporting v0 lancedb optimization support here! The change looks good. Please fix the `mypy` error.

georgeh0:
Thanks!

prrao87:
I addressed the minor feedback from the above in #2013. It should help avoid wasteful calling of `table.optimize()` when there are no appends happening. Cheers!

PR #2009, opened 2026-05-24
feat(docs): improve docs CI with proper triggers and build verification

dashitongzhi:
## Summary

This PR improves the docs CI workflow (`.github/workflows/docs_test.yml`) with the following changes:

### CI Improvements

1. **Add explicit quotes around path patterns** - For consistency and clarity with GitHub Actions syntax
2. **Add `uv.lock` to trigger paths** - The docs build depends on the Python package's lock file, so changes to `uv.lock` should trigger the docs CI
3. **Add explicit `push` trigger** - Alongside the PR trigger for better manual testing on branch pushes
4. **Add `workflow_dispatch`** - Allows manual triggering from the GitHub Actions UI

### Build Verification

5. **Verify `docs/dist` directory exists** - Confirms the build produced output
6. **Verify `index.html` was generated** - Uses `find` to check at least one page was built successfully

### Why These Changes Matter

- The docs CI was missing `uv.lock` as a trigger path, which meant updates to Python dependencies would not run the docs build verification
- Adding explicit build verification steps ensures the CI fails immediately if the build does not produce expected output (fail-fast behavior)
- The `workflow_dispatch` trigger enables debugging without needing to push commits

## Testing

- CI workflow runs on push/PR to `main` and `v1` branches when docs files, the workflow file itself, or `uv.lock` changes
- Build verification steps execute after `npm run build` to confirm output

## Files Changed

- `.github/workflows/docs_test.yml` - Added meaningful CI improvements

### DIFF
--- a/.github/workflows/docs_test.yml
+++ b/.github/workflows/docs_test.yml
@@ -4,8 +4,16 @@ on:
   pull_request:
     branches: [main, v1]
     paths:
-      - docs/**
+      - "docs/**"
       - ".github/workflows/docs_test.yml"
+      - "uv.lock"
+  push:
+    branches: [main, v1]
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs_test.yml"
+      - "uv.lock"
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


georgeh0 (commented):
Thanks for the improvement for CI!

Please revert changes for `id.py` (let's discuss that in #2006). I'll merge the change on the workflow part.

PR #2012, opened 2026-05-24
fix(serde): register _frombuffer under numpy 1.x and 2.x paths

### LINKED ISSUE #2011: [BUG] Restricted unpickler rejects numpy 1.x pickled ndarrays — incremental reindex fails on Intel Mac
**Describe the bug**

`cocoindex._internal.serde._register_builtin_types` registers numpy's `_frombuffer` under only the **numpy 2.x** allowlist key (`("numpy._core.numeric", "_frombuffer")`), and the lookup itself only attempts the numpy 2.x module path. On numpy 1.x:

1. The 2.x lookup fails → nothing is registered at all.
2. Pickled ndarrays under numpy 1.x reference `("numpy.core.numeric", "_frombuffer")` (no leading underscore on `core`) → restricted unpickler rejects them as "forbidden globals" even if the 2.x key WERE registered.

End result: any `ccc index` after the initial one fails on every changed file. Blocks all incremental reindexing for users stuck on numpy 1.x (Intel Mac, where torch 2.2.2 is the last available wheel and refuses to load with numpy ≥ 2).

**To Reproduce**

1. On Intel Mac (or any env where torch needs numpy 1.x), install `cocoindex-code[full]` with `numpy<2`, `transformers<5`, `sentence-transformers<5`.
2. `ccc init && ccc index` on any project — succeeds, embeddings stored.
3. Edit any indexed file.
4. Run `ccc index` again.

Observed:

```
Indexing: 759 files listed | 0 added, 0 deleted, 0 reprocessed, 755 unchanged, error: 4
```

Daemon log:

```
_pickle.UnpicklingError: Forbidden global during unpickling: numpy.core.numeric._frombuffer
cocoindex._internal.serde.DeserializationError: Failed to deserialize pickle payload
  (type_hint=numpy.ndarray[typing.Any, numpy.dtype[numpy.float32]],
   source=return type of cocoindex.ops.sentence_transformers.SentenceTransformerEmbedder.embed())
```

Minimal reproducer (no indexing needed):

```python
# In the cocoindex env with numpy 1.x installed
from cocoindex._internal import serde
keys = [k for k in serde._UNPICKLE_SAFE_GLOBALS if "_frombuffer" in k[1]]
print(keys)
# Actual:   []
# Expected: at least ('numpy.core.numeric', '_frombuffer'), and ideally also ('numpy._core.numeric', '_frombuffer')
```

**Expected behavior**

Incremental `ccc index` reads back previously-stored embeddings without error, regardless of which numpy major version pickled them. The serde allowlist should cover both the numpy 1.x and numpy 2.x module paths for `_frombuffer`.

**Screenshots**

N/A — text-only failure.

**CocoIndex Version**

- `cocoindex==1.0.6`
- `cocoindex-code==0.2.33`

**Additional context**

Environment:

- macOS Tahoe 26.4.1, Intel x86_64 (MacBookPro16,1, i7-9750H)
- Python 3.12.12 (uv-managed `cpython-3.12.12-macos-x86_64-none`)
- `torch==2.2.2` — last Intel Mac wheel; newer torch only ships arm64/Linux
- `numpy==1.26.4` — pinned because torch 2.2.2 fails to init with numpy 2.x (`RuntimeError: Numpy is not available`)
- `transformers==4.57.6`, `sentence-transformers==4.1.0`
- Embedding model: `nomic-ai/CodeRankEmbed`, `device: cpu`

Buggy code — `python/cocoindex/_internal/serde.py` lines 80–86 (as of 1.0.6):

```python
_core_numeric = getattr(np, "_core", None)
if _core_numeric is not None:
    _frombuffer = getattr(_core_numeric.numeric, "_frombuffer", None)
    if _frombuffer is not None:
        _UNPICKLE_SAFE_GLOBALS[("numpy._core.numeric", "_frombuffer")] = (
            _frombuffer
        )
```

Proposed fix — resolve numpy's `numeric` module under either layout and register the allowlist entry under both possible pickle key paths:

```python
_numeric = None
for _root_name in ("_core", "core"):
    _root = getattr(np, _root_name, None)
    if _root is None:
        continue
    _candidate = getattr(_root, "numeric", None)
    if _candidate is not None:
        _numeric = _candidate
        break

if _numeric is not None:
    _frombuffer = getattr(_numeric, "_frombuffer", None)
    if _frombuffer is not None:
        _UNPICKLE_SAFE_GLOBALS[("numpy.core.numeric", "_frombuffer")] = _frombuffer
        _UNPICKLE_SAFE_GLOBALS[("numpy._core.numeric", "_frombuffer")] = _frombuffer
```

I've verified this patch locally — incremental reindex completes without errors on the same environment. Happy to send a PR if useful.

shaiar:
## Description

  The restricted unpickler only registered `_frombuffer` under the numpy 2.x module path (`numpy._core.numeric`) and only attempted lookup via
  `np._core`. On numpy 1.x: lookup fails entirely (nothing registered), and pickled ndarrays reference `numpy.core.numeric` anyway — so the 2.x key
  wouldn't match even if registered.

  Resolve numpy's `numeric` submodule under either layout and register `_frombuffer` under both pickle key paths so ndarrays unpickle cleanly
  regardless of which numpy major version pickled them.

  ## Motivation

  Surfaces on environments stuck on numpy 1.x — notably Intel Mac, where `torch==2.2.2` (the last available wheel) pins `numpy<2`. Initial `ccc index`
   succeeds, but every subsequent `ccc index` after a file edit errors with `Forbidden global during unpickling: numpy.core.numeric._frombuffer` when
  reading back memoized embeddings. Verified locally on `cocoindex==1.0.6` / `numpy==1.26.4` / `torch==2.2.2` — incremental reindex completes cleanly
  with the patch.

  ## Breaking change

  No. Pure additive change to the unpickler allowlist; the existing 2.x key is still registered.

  ## Related issue

  Fixes #2011

### DIFF
--- a/python/cocoindex/_internal/serde.py
+++ b/python/cocoindex/_internal/serde.py
@@ -77,10 +77,21 @@ def _register_builtin_types() -> None:
             _UNPICKLE_SAFE_GLOBALS[(_dtype_sub.__module__, _dtype_sub.__qualname__)] = (
                 _dtype_sub
             )
-        _core_numeric = getattr(np, "_core", None)
-        if _core_numeric is not None:
-            _frombuffer = getattr(_core_numeric.numeric, "_frombuffer", None)
+        _numeric = None
+        for _root_name in ("_core", "core"):
+            _root = getattr(np, _root_name, None)
+            if _root is None:
+                continue
+            _candidate = getattr(_root, "numeric", None)
+            if _candidate is not None:
+                _numeric = _candidate
+                break
+        if _numeric is not None:
+            _frombuffer = getattr(_numeric, "_frombuffer", None)
             if _frombuffer is not None:
+                _UNPICKLE_SAFE_GLOBALS[("numpy.core.numeric", "_frombuffer")] = (
+                    _frombuffer
+                )
                 _UNPICKLE_SAFE_GLOBALS[("numpy._core.numeric", "_frombuffer")] = (
                     _frombuffer
                 )


georgeh0:
Thanks for the fix!

Issue #2015, opened 2026-05-24
[FEATURE] Support states for components


georgeh0:
Users can declare states for components, e.g.

```python
my_state = coco.use_state("key", initial_value)  # initial value defaults to `None`
```

Set / get the state by

```python
my_state.value = ...
my_state.value
```

Notes:
- State key is unique within the component
- We can start from supporting states declared at top level of a component (i.e. currently on the component's own path without another sub path, and not within another layer of memoized function) and raise an exception otherwise. Later we can see if we want to relax the requirement based on real usages.
- For a component, if it no longer declares a state during reprocessing, the state should be deleted.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

hakuuww:
Can I work on this issue? 👀

badmonster0:
assigned, thanks for contribution Anthony!

hakuuww:
From George through Discord 
```
The initial_value is only kicked in when the same state isn't declared
in the component during last time's processing. It's ignored if it's
already declared last time.

Once declared, the value can be accessed by state.value, 
or set by state.value = ... any time. e.g. you can set it, then access it later, 
in which case you'll get the new value.  

It's only committed to the database during commit for the component.
```

badmonster0:
@hakuuww  thanks for the contribution!

PR #2034, opened 2026-05-28
feat: introduce coco.use_state() for persistent per-component user defined state

hakuuww:
## Description
The commits ordering and implementation follow a bottom up approach starting from the state backend schema to the user api level.

`coco.use_state(key, initial_value?)` is a React-style hook for storing persistent state scoped to a processing component. The returned `StateHandle` exposes `.value` for reading and writing; the written value is serialized to LMDB and restored the next time the same component runs.

```python
my_state = coco.use_state("key", initial_value)  # initial_value defaults to None

# Read the current value
print(my_state.value)

# Update the value — persisted to DB at component commit
my_state.value = new_value
```

References https://github.com/cocoindex-io/cocoindex/issues/2015

## Semantics of `coco.use_state("key", initial_value)`

The state key must be unique within the component. On the first run of the component, `initial_value` is the value returned by the state handler; on subsequent runs, the value stored from the previous run is returned (`initial_value` is ignored once a value has been stored).

Calling `use_state` with no `initial_value` is allowed:
```python
coco.use_state("key")  # initial_value not provided
```
The first run returns `None`, and `None` is flushed to DB. The user can later read back `None` from this user state like any other value.

The value can be read or written at any point during the component run via `state.value` / `state.value = ...`. Writes are in-memory until the component commits/submits; only the final value at commit time is persisted.

`use_state` raises `RuntimeError` if called in the following situations. These errors surface as component build failures — logged by default but not propagated to `app.update()` unless a custom exception handler re-raises them.

```python
@coco.fn
def component_entry_fn():
    # ✓ This is fine: use_state at component top level
    state1 = coco.use_state("counter", initial_value=0)
    state1.value += 1

    some_other_fn()

    with coco.component_subpath(...):
        # ✗ Not supported: use_state inside component_subpath block.
        # State is keyed to the component's own stable path; a shifted
        # path would silently scope it incorrectly.
        # state2 = coco.use_state("nested", initial_value=None)
        pass


@coco.fn(memo=True)
def some_memoized_fn():
    # ✗ Not supported: use_state in memoized function.
    # On a memo cache hit the function body is skipped entirely, so
    # use_state would never run — the key would appear undeclared and
    # be garbage-collected on the next commit.
    # state = coco.use_state("memoized", initial_value=0)

    some_helper()  # ✗ Also not supported: the guard propagates transitively.
                   # If some_helper calls use_state, it also raises —
                   # on a cache hit the whole call stack is skipped.


@coco.fn
def some_helper():
    # ✗ Not supported when called from within a memoized function body.
    # state = coco.use_state("indirect", initial_value=0)
    pass
```

If a component no longer declares a state key during reprocessing through `coco.use_state`, that key is deleted (after this component commits/submits).
If the component itself is removed entirely, all its state is cleaned up (along with all other component data).

## Implementation

This change follows a bottom-up approach, from the LMDB schema up to the Python API.

### LMDB Schema (`state/db_schema.rs`)

Adds `UserStatePrefix` (byte `0xc0`) and `UserState(StableKey)` entry key variants to `StablePathEntryKey`, with symmetric encode/decode implementations. Entries are stored under the owning component's stable path, so all state for a given component is contiguous — enabling fast prefix scans via `list_user_states` and preserving locality with the rest of that component's data (fn memos, target state).

### AppStore CRUD (`state_store/app_store.rs`)

Four methods following the same read/write split as the existing fn-memo layer:

- **`list_user_states`**: standalone read txn, called during prefetch outside any write transaction.

- **`write_user_state`**, **`delete_user_state`**, **`delete_all_user_states`**: accept `&mut WriteTxn` so they batch atomically with the rest of the component-commit writes.

### UserStateCache (`engine/context.rs`)

Per-component in-memory cache (analogous to `FnMemoCache`). Populated by `prefetch_user_states` at build start; `use_state` / `update_declared` accumulate changes during the build; `flush_to_db` commits them at component commit time.

The cache uses a single `HashMap<StableKey, UserStateEntry>` where `UserStateEntry` is an enum encoding the per-key lifecycle:

- **`Loaded(Vec<u8>)`**: prefetched from DB at build start; not yet claimed by `use_state()` this build. Deleted at flush if it stays in this state (key dropped by the component).
- **`Declared(Vec<u8>)`**: claimed by `use_state()` and optionally overwritten by `update_declared()`. Written to DB at flush time.

The enum keeps both fields for the same key co-located in one map entry and makes invalid states impossible by construction. `flush_to_db` is a single pass: `Loaded` entries are deleted (set-reduction), `Declared` entries are written. On full-reprocess, `delete_all_user_states` is called first, then all `Declared` entries are written — covering GC when a component path disappears.

### Python API (`python/cocoindex/_internal/api.py`)

`StateHandle[T]` wraps the current in-memory value and the `ComponentProcessorContext` reference. Assigning to `.value` serializes the new value via the existing serde layer and calls `update_user_state` on the Rust side immediately, so subsequent reads in the same run see the updated value. The final persisted value is whatever was last written before the component commits.

### feat(state_store): add UserState entry key to LMDB schema

Adds `UserStatePrefix` (`0xc0`) and `UserState(StableKey)` (`0xc0 | key`) variants to `StablePathEntryKey`, with symmetric `Encode`/`Decode` impls. These will back per-component user state storage for `coco.use_state()`.

### feat(state_store): add AppStore CRUD methods for user state

Adds four methods on `AppStore` for managing per-component user state, following the same read/write split used by the existing fn-memo methods:

- **`list_user_states`**: standalone read txn (mirrors `list_fn_memos`). Called during component prefetch, outside any write transaction.

- **`write_user_state`**, **`delete_user_state`**, **`delete_all_user_states`**: accept `&mut WriteTxn` so the caller can batch them atomically with the rest of the component-commit writes (tracking info, fn memos, target state). `delete_all_user_states` covers GC when a component path is removed.

**Tests**: Includes six cases, all against a real in-process LMDB env via `tempfile`: empty list on fresh path; write then list; write overwrites existing; write + delete atomically; delete_all + write; writes to two different `StablePaths` don't bleed.

### feat(engine): add UserStateCache for per-component persistent state

Introduces `UserStateCache` (analogous to `FnMemoCache`). The cache uses a single `HashMap<StableKey, UserStateEntry>` with a two-variant enum (`Loaded` / `Declared`) encoding the per-key lifecycle, so `use_state()` resolves in one `Entry` lookup, `flush_to_db` is a single pass with no cross-map set-difference loop, and invalid states are impossible by construction. Adds `user_states: UserStateCache` to `ComponentBuildingState`.

### feat(engine): wire UserStateCache into the component lifecycle

Adds `prefetch_user_states` / `use_state` / `update_user_state` to `ComponentProcessorContext`. Calls `prefetch_user_states` at component build start (after `prefetch_fn_memos`). Adds `user_states: UserStateCache` parameter to `commit_in_txn` and `commit`; flushes after fn memos. The delete path passes `UserStateCache::new()` (empty, unloaded) so `flush_to_db` calls `delete_all_user_states` — GC for free.

### feat(py): expose use_state and StateHandle to Python

Adds `use_state` and `update_user_state` to `PyComponentProcessorContext` (PyO3), with type stubs in `core.pyi`. Adds `StateHandle[T]` and `coco.use_state(key, initial_value?)` to the Python API. `use_state` serializes the initial value, calls the Rust core, deserializes the stored bytes, and returns a `StateHandle`. Assigning to `StateHandle.value` serializes the new value and calls `update_user_state`. Includes integration tests covering: initial value on first run, persistence across runs, per-component independence, and GC after component deletion.

### feat(py): guard use_state against prohibited call sites

Raises `RuntimeError` in three cases where `use_state` cannot work correctly:

1. **Inside a `with coco.component_subpath()` block**: state is keyed to the component's own stable path, so a shifted path would silently read/write state under the wrong identity. Check: `ctx._core_path != ctx._core_processor_ctx.stable_path`.

2. **Inside an inline memoized function call** (not mounted as a component): on a memo cache hit the body is skipped entirely, so `use_state` would never run and the key would appear undeclared, causing set-reduction to GC it on the next commit. Check: `ctx._in_memo_fn` (new bool field on `ComponentContext`, set `True` by `SyncFunction`/`AsyncFunction` when executing a memo function body inline).

3. **Transitively inside a memoized call stack**: `_in_memo_fn=True` propagates through non-memoized `@coco.fn` calls made from within a memoized body. If a helper function calls `use_state` and is called from a memoized function, it also raises — the GC risk is identical regardless of call depth, since on a cache hit the entire call stack is skipped.

Also normalizes the Rust-level duplicate-key `ValueError` to `RuntimeError` so all `use_state` usage errors have a consistent type. Adds tests for the async memo guard path, duplicate key case, and indirect (transitive) memo guard for both sync and async call paths. Documents all error cases with explanations in the `Raises` section of the docstring.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)


python/tests/core/test_use_state.py
```diff
@@ -0,0 +1,200 @@
...
+
+    _raised: dict[str, bool] = {}
+
+    @coco.fn(memo=True)
+    def _memoized_helper(item: str) -> None:
+        try:
+            coco.use_state("counter", 0)
+            _raised[item] = False
+        except RuntimeError:
+            _raised[item] = True
```
georgeh0, line 154: Can we add another test to guard against  `use_state()` calls within memoized functions indirectly?

e.g. 

```python
@coco.fn
def _inner_helper(item: str) -> None:
  coco.use_state("counter", 0)  # expect exception here  

@coco.fn(memo=True)
def _memoized_helper(item: str) -> None:
  _inner_helper(item)

...
```

rust/core/src/engine/component.rs
```diff
@@ -795,22 +795,13 @@ impl<Prof: EngineProfile> Component<Prof> {
...
-                        eager_existence_upsert(processor_context).await?;
-                    }
-
                     // Eagerly load all function-memo entries for this component
                     // into the per-build cache, so every subsequent fn-call probe
                     // serves from memory. Skipped under `full_reprocess` and in
                     // delete mode (no `ComponentBuildingState`); see the cache
                     // flush logic for how those cases are handled at commit time.
                     processor_context.prefetch_fn_memos().await?;
+                    processor_context.prefetch_user_states().await?;
```
georgeh0, line 956: I'm thinking about combining these two and prefetch both fn_memos and user_states in a single transaction.

rust/core/src/engine/context.rs
```diff
@@ -451,10 +437,112 @@ pub(crate) fn decode_stored_entry<Prof: EngineProfile>(
...
+/// Mirrors the `FnMemoCache` pattern: `populate` fills it from a
+/// standalone read at build start, then `use_state` / `update_declared`
+/// accumulate the declared states, and `flush_to_db` applies the
+/// set-reduction diff (write declared, delete loaded-but-not-declared)
+/// inside `commit_in_txn`.
+pub(crate) struct UserStateCache {
+    /// Values loaded from LMDB at the start of this build.
+    loaded: HashMap<StableKey, Vec<u8>>,
+    /// Keys declared (and their current values) during this build.
+    declared: HashMap<StableKey, Vec<u8>>,
```
georgeh0: Consider merging these into a single map of struct?

In general "list of struct" is better than "multiple lists" since multiple fields for the same key is usually accessed together.

Also consider storing the python value rather than raw bytes (similar to memo), otherwise each time a deserialization is needed (OK to follow up in another PR if you prefer).

rust/core/src/engine/context.rs
```diff
@@ -451,10 +437,112 @@ pub(crate) fn decode_stored_entry<Prof: EngineProfile>(
...
+            client_bail!(
+                "coco.use_state() key {:?} declared more than once in the same component run",
+                key
+            );
+        }
+        // The initial_value is considered when the same state isn't declared via
+        // use_state() in this component and committed during the last component run.
+        //
+        // It's ignored if it's already declared and committed last time.
+        let value = self.loaded.get(&key).cloned().unwrap_or(initial_value);
```
georgeh0: Cloning a `Vec<u8>` can have unnecessary cost depends on the size of the state object.

python/tests/core/test_use_state.py
hakuuww: Yep working on it

rust/core/src/engine/component.rs
hakuuww: Ok, I think to keep it clean, I can do it in a followup PR #2041

rust/core/src/engine/context.rs
hakuuww: Good idea, I think instead of "list of struct" we can do list of enum, this guarantees that during runtime at any given time, a single key can only map to either "loaded" state or "declared" state. This solves the locality and also simplifies flush to db logic.

python/tests/core/test_use_state.py
hakuuww: Just finished, it is in the last commit.

rust/core/src/engine/context.rs
hakuuww: > Also consider storing the python value rather than raw bytes (similar to memo), otherwise each time a deserialization is needed (OK to follow up in another PR if you prefer).

Will do in a follow-up PR.

hakuuww:
Updated the `UserStateCache` implementation to be a hashMap of key -> enum instead of two seperate hashMaps.
This simplifies implementation logic and improves access locality. 

Added additional guards to raise exception when the user calls use_state inside a memoized function.

Added TODO for #2039.

Added more documentation in code. 

## Please do commit-by-commit review

rust/core/src/engine/context.rs
hakuuww: > Consider merging these into a single map of struct?
> 
> In general "list of struct" is better than "multiple lists" since multiple fields for the same key is usually accessed together.

Updated the commit

python/tests/core/test_use_state.py
georgeh0: The test for `use_state` in *indirectly* memoized function seems not there yet?

rust/core/src/engine/context.rs
```diff
@@ -451,10 +437,141 @@ pub(crate) fn decode_stored_entry<Prof: EngineProfile>(
...
+                    );
+                }
+                UserStateEntry::Loaded(v) => {
+                    let value = v.clone();
+                    e.insert(UserStateEntry::Declared(value.clone()));
+                    Ok(value)
+                }
+            },
+            Entry::Vacant(e) => {
+                e.insert(UserStateEntry::Declared(initial_value.clone()));
```
georgeh0, line 535: We should avoid cloning `Vec<u8>` (which can be arbitrarily long).

We may use the similar data type used for memo entries `Prof::FunctionData`.

georgeh0 (commented):
Please also merge with the latest `main` and resolve conflicts. Thanks!

python/tests/core/test_use_state.py
hakuuww: > The test for `use_state` in _indirectly_ memoized function seems not there yet?

Just added it.

rust/core/src/engine/context.rs
hakuuww: Is it ok to do this in a followup PR? #2039

georgeh0: OK, sounds good.

Please merge with latest main and resolve conflicts. I'll merge after that. Thanks!

hakuuww: Ok I think conflict resolution may take a while, its a bit complex.

georgeh0:
Thanks for resolving the conflicts! Merging.

PR #2047, opened 2026-06-01
feat(resources): support negated exclude patterns

### LINKED ISSUE #1778: Support negation in excluded_patterns to allow exceptions

sdhilip200:
## Summary

- Support leading `!` rules in `PatternFilePathMatcher.excluded_patterns` to re-include paths after broader exclude rules.
- Preserve directory traversal when a negated descendant pattern can re-include files inside an otherwise excluded directory.
- Add Rust and Python tests for file re-inclusion and the `.github/workflows` exception use case.
- Document the ordered, last-match-wins behavior for excluded patterns.

## Motivation and context

Users currently cannot express common filtering rules such as excluding all dot-directories while keeping `.github/workflows` for CI/CD analysis. This change makes `excluded_patterns` behave more like `.gitignore` by allowing later negated rules to override earlier broader excludes.

## Breaking changes

None.

## Related issues

Closes #1778.

## Test Plan

- `cargo fmt --check`
- `cargo test -p cocoindex_ops_text pattern_matcher`
- `cargo test -p cocoindex_ops_text`
- `/usr/bin/env -u CONDA_PREFIX uv run pytest python/tests/resources/test_file_path_matcher.py -q`
- `/usr/bin/env -u CONDA_PREFIX uv run pytest python/tests/resources -q`
- `/usr/bin/env -u CONDA_PREFIX uv run ruff check python/cocoindex/resources/file.py python/tests/resources/test_file_path_matcher.py`
- `npm run build` from `docs`

### DIFF
--- a/docs/src/content/docs/common_resources/data_types.mdx
+++ b/docs/src/content/docs/common_resources/data_types.mdx
@@ -127,7 +127,7 @@ matcher = PatternFilePathMatcher(
 **Parameters:**
 
 - `included_patterns` — Glob patterns ([globset](https://docs.rs/globset) syntax) for files to include. Use `**/*.ext` to match at any depth. If `None`, all files are included.
-- `excluded_patterns` — Glob patterns ([globset](https://docs.rs/globset) syntax) for files/directories to exclude. Excluded directories are not traversed.
+- `excluded_patterns` — Glob patterns ([globset](https://docs.rs/globset) syntax) for files/directories to exclude. Excluded directories are not traversed. Prefix a pattern with `!` to re-include matching paths after a broader exclude pattern; rules are evaluated in order, and the last matching exclude rule wins.
 
 :::note
 Patterns use [globset](https://docs.rs/globset) semantics: `*.py` matches only in the root directory; use `**/*.py` to match at any depth.
--- a/docs/src/content/docs/connectors/localfs.mdx
+++ b/docs/src/content/docs/connectors/localfs.mdx
@@ -127,6 +127,16 @@ async for file in localfs.walk_dir("/path/to/project", recursive=True, path_matc
     await process(file)
 ```
 
+`excluded_patterns` also supports a leading `!` to keep exceptions after a broader exclude. Rules are evaluated in order, so the last matching exclude rule wins:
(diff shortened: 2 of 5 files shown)


georgeh0:
Closing since #1778 is already resolved.

PR #2076, opened 2026-06-07
refactor(engine): prefetch fn-memos and user-states in one read txn

hakuuww:
## What

Combine the two separate prefetch calls (`prefetch_fn_memos` + `prefetch_user_states`) into a single `prefetch_states` that loads both the function-memo and user-state ranges for a component under **one** LMDB read transaction, via a new `AppStore::prefetch_fn_processing_states`.

The two caches were always loaded back-to-back at the start of build mode, each opening its own read snapshot. Combining them into one read transaction removes a redundant read-txn begin per component.

## When this helps

Under `MDB_NOTLS`, every read-txn begin takes the reader-table mutex. The benefit shows up under `mount_each` fan-out, where many child components prefetch concurrently — this halves reader-table-mutex acquisitions and reader-slot occupancy against LMDB's `MDB_READERS_FULL` limit, allowing for greater concurrency during component building.

No benchmark is included for that reason — the justification is mechanistic (2 read-txn begins → 1 read-txn begin per component prefetch).

## Test consolidation

With the prefetch unified, `AppStore::list_user_states` has no production caller left, it is only used in tests. This PR migrates every user-state test read-back onto the production path via a `read_regular_states` helper that calls `prefetch_fn_processing_states(p).1`, so tests exercise the same read code the engine runs instead of a parallel, test-only reader.

The `user_states_isolated_by_kind` test keeps its Live-state assertions via `read_user_state` point-reads — the unified prefetch loads only `Regular` states (`Live` is point-read in production), so no test-only bulk reader is needed. With the last caller gone, `list_user_states` is deleted.

## Behavior

No behavior change to production reads: the two caches are still populated together, and the `full_reprocess` / already-loaded / delete-mode skip paths are preserved.

### DIFF
--- a/rust/core/src/engine/component.rs
+++ b/rust/core/src/engine/component.rs
@@ -947,13 +947,13 @@ impl<Prof: EngineProfile> Component<Prof> {
                         eager_existence_upsert(processor_context).await?;
                     }
 
-                    // Eagerly load all function-memo entries for this component
-                    // into the per-build cache, so every subsequent fn-call probe
-                    // serves from memory. Skipped under `full_reprocess` and in
-                    // delete mode (no `ComponentBuildingState`); see the cache
-                    // flush logic for how those cases are handled at commit time.
-                    processor_context.prefetch_fn_memos().await?;
-                    processor_context.prefetch_user_states().await?;
+                    // Eagerly load all function-memo and user-state entries for
+                    // this component into the per-build cache (one read txn), so
+                    // every subsequent fn-call probe and `use_state` serves from
+                    // memory. Skipped under `full_reprocess` and in delete mode
+                    // (no `ComponentBuildingState`); see the cache flush logic
+                    // for how those cases are handled at commit time.
+                    processor_context.prefetch_states().await?;
 
                     if memo_fp_to_store.is_some() {
                         *self.inner.last_memo_fp.lock().unwrap() = memo_fp_to_store;
--- a/rust/core/src/engine/context.rs
+++ b/rust/core/src/engine/context.rs
@@ -261,7 +261,7 @@ impl<Prof: EngineProfile> FnMemoCache<Prof> {
 
     /// Insert prefetched rows from the database into the cache and mark it
     /// fully loaded. The async I/O lives at the context layer
-    /// ([`ComponentProcessorContext::prefetch_fn_memos`]); this is the sync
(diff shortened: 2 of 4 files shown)


hakuuww:
~~## Note~~

~~The old `list_user_state`method is still kept, with no caller in production code, but called in test code, which tests the correctness of the `user_state_cache`.~~

georgeh0 (approved):
Thanks!

georgeh0:
> ## Note
> The old `list_user_state`method is still kept, with no caller in production code, but called in test code, which tests the correctness of the `user_state_cache`.

Ideally the code to be tested should be the same as the code running in production, and we should avoid code only used for testing. I think these tests should be migrated to test the new method instead.

hakuuww:
> Ideally the code to be tested should be the same as the code running in production, and we should avoid code only used for testing. I think these tests should be migrated to test the new method instead.

I'm thinking, maybe it is better to just pass a shared `RoTxn` instead of combining the two methods for fn_memo and user_states. 

The same pattern is already used for write transactions. 

This also means there is no need to do any test migrations. 

The only downside is that the caller must manage the life cycle of the `RoTxn`.

georgeh0:
> > Ideally the code to be tested should be the same as the code running in production, and we should avoid code only used for testing. I think these tests should be migrated to test the new method instead.
> 
> I'm thinking, maybe it is better to just pass a shared `RoTxn` instead of combining the two methods for fn_memo and user_states.
> 
> The same pattern is already used for write transactions.
> 
> This also means there is no need to do any test migrations.
> 
> The only downside is that the caller must manage the life cycle of the `RoTxn`.

I think the main rationale behind merging them is: they're the similar types of information for a component's processing: both are states that are needed and accumulated during running the processing fn, so they're always fetched together. So we don't have to make the fetching function over fine granularity. Using a single function will allow the implementation to fetch them in the most efficient way (not make a big difference for LMDB, but while the backend is Postgres, this may make a real difference).

BTW, for the merged function, we may consider call it something like `prefetch_fn_processing_states`, to make the intention clear (so in the future if there're a 3rd type of states needed during processing added there, it's clear they'll belong here)

hakuuww:
Updated the function naming, and refactored to remove `list_user_states`, all tests now wires through `prefetch_fn_processing_states`

georgeh0 (approved):
Thanks!

Issue #2017, opened 2026-05-25
[FEATURE] Leverage component states for live components to skip full processing on rerun


georgeh0:
Currently our [live components](https://cocoindex.io/docs/advanced_topics/live_component/) usually trigger a `update_all()` ([example](https://github.com/cocoindex-io/cocoindex/blob/a2faeb2d18306fd61dc1d25045d3b3749eae8ee1/python/cocoindex/connectors/oci_object_storage/_source.py#L393)) at the beginning of the process starts. Even if sub components are usually memoized so a full update is cheap, but there're still extra metadata scanning and it's usually not necessary if:

- A bootstrap already done in the previous run (after #2015 is done, we can leverage it to indicate that).
- And the incremental update stream is durable (e.g. Kafka).

Besides, we also need to make sure once there're [logic / context changes](https://cocoindex.io/docs/programming_guide/function/#change-detection), a full update will still be needed.

After the core has the support, please update the [OCI Object Storage](https://github.com/cocoindex-io/cocoindex/blob/a2faeb2d18306fd61dc1d25045d3b3749eae8ee1/python/cocoindex/connectors/oci_object_storage/_source.py#L393) source connector to leverage the support.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

pyjuan91:
This looks like a great one, could I take it? Now that [#2015](https://github.com/cocoindex-io/cocoindex/issues/2015) is in, I'll build on the component state to gate the startup update_all() on the OCI source. Will keep it scoped to the connector.

badmonster0:
Assigned, thanks!

pyjuan91:
I've dug into the dependency ([#2015](https://github.com/pyjuan91/cocoindex/issues/2015) / merged [#2034](https://github.com/pyjuan91/cocoindex/issues/2034)) and the current live-component + change-detection code. A couple of design questions before I start, to make sure I build the right thing:

1. Scope of the "logic/context changed" guard.
[...]

PR #2078, opened 2026-06-08
feat(core): add read_committed_state — process_live-callable reader for use_state

pyjuan91:
Part 1/2 of #2017. 

Adds `read_committed_state(key)`, a read-only, `process_live`-callable reader for the per-component state written by `coco.use_state` inside `process()`. Surfaced on `LiveComponentOperator` and `LiveMapSubscriber`.

This is the framework primitive a durable live source needs to skip its startup full scan: in `process()` it persists a bootstrap flag + logic version via `use_state`; on a later run it reads them back here (before `update_full`) and skips the scan when bootstrapped + durable + logic-unchanged. (PR 2 will wire this into the OCI source).

**Design note for @georgeh0:**
During implementation, I found a live component's stored `logic_deps` closure captures only its own body's logic, not its mounted child components' (children drain their `logic_deps` before propagating; the OCI source's user transform runs in child components). 

So automatic framework-level logic-change detection ("did any logic in my live subtree change?") would need subtree-closure aggregation — a substantially larger change. This PR takes the smaller path: the connector declares its own logic version (compared via this read primitive). Happy to pursue the automatic approach as a follow-up if you'd prefer it. 

**Open questions:**
1. **Naming:** `read_committed_state` vs. something that pairs more explicitly with `use_state`.
2. **D2 vs D1:** Connector-declared logic version (this PR) vs. framework auto subtree-logic detection (bigger follow-up).
3. **Placement:** On `LiveComponentOperator` + `LiveMapSubscriber` as done here.

Refs #2017 (reserving Closes for PR 2).


rust/core/src/engine/live_component.rs
```diff
@@ -485,6 +485,27 @@ impl<Prof: EngineProfile> LiveComponentController<Prof> {
...
+    /// so it is safe in the non-committing `process_live` context.
+    pub async fn read_committed_state(&self, key: &StableKey) -> Result<Option<Vec<u8>>> {
+        let rows = self
+            .component
+            .app_ctx()
+            .app_store()
+            .list_user_states(self.component.stable_path())
+            .await?;
+        Ok(rows.into_iter().find_map(|(k, v)| (k == *key).then_some(v)))
+    }
```
georgeh0, line 529: Since in the underlying LMDB, each key is stored in separate entries. So we can directly read the specific entry instead of reading all then filtering out.

However, I realize a bigger issue which we didn't discussed earlier.  Consider a live component can also have its processing logic (which is executed during `update_full`) that declares its own states. Mixing these two type of states will complicate things, e.g. during processing for a regular component, we're doing a prefix read for all state entries (which includes the one created by live components), but at the end it'll prune all states that exist before but no longer declared during this time of processing. So we'll likely lose the states declared by live components here.

The solution in my mind is:

Extend the [`UserState(StableKey)`](https://github.com/cocoindex-io/cocoindex/blob/532a070d6465291b6fa61c449af8498d7d48a072/rust/core/src/state/db_schema.rs#L43). Add another key part to represent the state kind, e.g. `UserState(StateKind, StableKey)`. `StateKind` is an enum, initially with two variants, e.g. `Regular` (for regular processing logic) and `Live` (for live components specific states). Now two kinds of states  are under different key prefix and we only read the specific kind for each specific case.

Let me know if you have any questions and other thoughts. Thanks!

pyjuan91: Both make sense, thanks for catching the prune interaction, that's a real bug.

**(1) Point-get:** agreed, I'll switch `read_committed_state` to a single keyed `get` instead of `list_user_states` + filter.

**(2) `StateKind` split:** agreed. I'll extend `UserState(StableKey)` → `UserState(StateKind, StableKey)` with `StateKind { Regular, Live }`, keeping the `0xc0` + kind-byte + key layout so the per-kind prefix scan invariant holds. The regular `use_state` path (prefetch + set-reduction prune in `into_flush_plan`, plus the commit `clear_all_first`) stays entirely on `Regular`, behavior unchanged, and `read_committed_state` reads `Live`. Whole-component GC is untouched since teardown range-deletes the full `StablePath` prefix, which covers both kinds.

One consequence worth confirming: since the regular prune owns `Regular`, the live component's own bootstrap state can't be a plain `use_state` write anymore (that'd be `Regular` and get pruned). I'll route it through a `Live`-kind write primitive paired with `read_committed_state`, and keep user-facing `coco.use_state` on `Regular`. So this PR would grow a small `Live` writer alongside the reader; the OCI connector wires into it in PR 2.

rust/core/src/state/db_schema.rs
```diff
@@ -17,6 +17,45 @@ use crate::state::{
...
+pub enum StateKind {
+    Regular,
+    Live,
+}
+
+impl storekey::Encode for StateKind {
+    fn encode<W: Write>(&self, e: &mut storekey::Writer<W>) -> Result<(), storekey::EncodeError> {
+        match self {
+            StateKind::Regular => e.write_u8(0x00),
+            StateKind::Live => e.write_u8(0x01),
```
georgeh0: We usually avoid `0x00` and `0x01`, since they'll be encoded as two bytes by `storekey` (`0x00` are reserved for delimiter, so it's escaped with a preceding `0x01`, so `0x01` is also escaped with a preceding `0x01`).

We can use `0x02` and `0x03` for them.

rust/py/src/live_component.rs
```diff
@@ -117,6 +118,46 @@ impl PyLiveComponentController {
...
     }
+
+    /// Read previously-committed user state for `key` (written via
+    /// `coco.use_state` inside `process()`), as a fresh standalone read.
+    /// Returns `None` if absent. Callable from `process_live` to gate a
+    /// durable startup-scan skip.
+    pub fn read_committed_state_async<'py>(
+        &self,
+        py: Python<'py>,
+        key: String,
```
georgeh0: Consider accepting a `StableKey` here to be broader.

rust/core/src/state/db_schema.rs
pyjuan91: TIL, thanks for explaining! Hadn't realized 0x00/0x01 get escaped by storekey. Updated to 0x02/0x03.

georgeh0 (approved):
Thanks for the PR!

PR #2082, opened 2026-06-09
test(resources): add unit tests for chunk and schema modules

nuthalapativarun:
## Summary
- Add `test_chunk.py` with unit tests for `TextPosition` and `Chunk` frozen dataclasses (construction, immutability, equality)
- Add `test_schema.py` with unit tests for `VectorSchema`, `MultiVectorSchema`, and the `get_vector_schema`/`get_multi_vector_schema` helper functions
- Pure Python tests; no external dependencies or CocoIndex app/environment required

## Test Plan
- [ ] `uv run pytest python/tests/resources/` passes

### DIFF
--- a/python/tests/resources/test_chunk.py
+++ b/python/tests/resources/test_chunk.py
@@ -1,0 +1,85 @@
+"""Tests for Chunk and TextPosition dataclasses."""
+
+import pytest
+
+from cocoindex.resources.chunk import Chunk, TextPosition
+
+
+class TestTextPosition:
+    def test_construction(self) -> None:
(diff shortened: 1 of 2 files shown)


georgeh0 (commented):
Thanks for creating these tests!

However, I think these tests are not verifying any non-trivial logic, e.g. [`chunk.py`](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/resources/chunk.py) only has data class definitions without any non-trivial logic. The test itself is even more complex than the code to be tested. I prefer not to create and maintain these tests. They need to be maintained but they provide no / very little value.

PR #2084, opened 2026-06-09
`feat`: Add (initial) `zvec` target integration

### LINKED ISSUE #1701: [FEATURE] Add `zvec` as a target connector
**What is the use case?**

[`zvec`](https://github.com/alibaba/zvec) is an embedded vector database from Alibaba, built on their production [`Proxima`](https://zvec.org/en/blog/introduction/) engine. No server, no daemon; just `pip install zvec` and point it at a local path. It supports dense and sparse vectors, HNSW indexing w/ int8 quantization, filtered hybrid search, and _built-in_ multi-vector retrieval + reranking.

CocoIndex already has [`LanceDB`](https://cocoindex.io/docs/targets/lancedb) as an embedded vector DB target; `zvec` fills a similar niche but brings a few things additionally: native sparse vector support, multi-vector retrieval w/ built-in reranking, and fine-grained resource governance (memory limits, CPU thread caps, mmap mode for datasets exceeding RAM).

Since both `zvec` and `LanceDB` are embedded, path-based vector DBs, the [LanceDB target connector](https://cocoindex.io/docs/targets/lancedb) is a natural implementation template? The mapping seems to be straightforward: `path` replaces `db_uri`, `collection_name` replaces `table_name`, `zvec.open(path)` serves as the shared handle, and `collection.optimize()` mirrors `table.optimize()` for periodic index compaction after incremental writes.

A `Zvec` target spec would need `path`, `collection_name`, and _optionally_ `num_transactions_before_optimize` (same pattern as LanceDB) plus `enable_mmap` for large-dataset scenarios. For mutations, `zvec` supports `insert`, `delete` by ID, and `delete_by_filter`; there's no native `upsert()`, so it'd need a _delete-then-insert_ pattern (similar to how some other connectors handle this). CocoIndex's `Vector[Float32, N]` maps directly to `zvec`'s `VECTOR_FP32` type; scalar metadata goes into `Doc.scalars`.

This would be **target/sink only**. `zvec` has no CDC, no change streaming, and no way to enumerate all documents incrementally; so it doesn't make sense as a source connector.

_CC: @iaojnh @Cuiyus, @feihongxu0824: Hope you don't mind the ping; CC'ing you to keep you in the loop re: this integration._ 

---

Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`** or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

Haleshot:
Closes #1701

First pass at a `zvec` target connector. zvec runs in-process (just a pip install, no server), so it slotted in cleanly as a target-only connector built on a local `ManagedConnection` over a base directory. The shape ended up close to Qdrant's collection/document target, with the typed `from_class` schema borrowed from SQLite/LanceDB. Each CocoIndex collection is a zvec collection directory, and each row is a `Doc` keyed by a single string id.

A few things I pinned down against the real `zvec==0.4.0` wheel: every collection needs at least one vector field, the primary key has to be a single column since it maps to the string `Doc.id` (so composite keys are rejected), and collection names need 3+ characters, and dense vectors have to be float32 or float16. zvec's index rejects anything wider, so for smaller storage you set `quantize` on the vector (e.g. `"int8"`) instead of reaching for a narrower element type.

Left out `MultiVectorSchemaProvider` for now. zvec's "multi-vector retrieval with reranking" is a query-time thing over named vector fields, not a per-doc _ColBERT-style_ storage type, so there's nothing to map on the write side yet.

Tests, mypy, and `ruff format` are green. `ruff check` only grumbles about the `from ._target import *` line, which every existing connector already does too.

Opening as a draft for now...

CC: @feihongxu0824, @egolearner


badmonster0:
hey @Haleshot thanks a lot for the PR, what kind of e2e test did you do?

Haleshot:
Thanks for the quick reply, @badmonster0!

No full e2e example yet (although I played around with some things locally).

So far I’ve only run the connector tests in `python/tests/connectors/ test_zvec_target.py`: they create a real zvec on disk, write/update/delete through CocoIndex, and read back with zvec, so the connector path is tested without mocks.

I’m keeping this as draft until I add a proper example-based (e2e) test, since I remember that being brought upin the ChromaDB PR too: https://github.com/cocoindex-io/cocoindex/pull/1548#issuecomment-2040792807

Haleshot:
Wow, I'm unsure why this PR was closed. Ugh.

PR #2089, opened 2026-06-10
docs: fix dead msgspec doc links (jcristharif.com -> msgspec.dev)

georgeh0:
## Summary
- Fix two dead documentation links in `serialization.mdx`: the msgspec docs moved from `jcristharif.com/msgspec/*` to `msgspec.dev`, and the old URLs 404 — which fails the `Check links` CI on `main`.
- Both new URLs verified to return 200.

## Test plan
- CI `Check links` (lychee).


badmonster0:
Thanks!

PR #2092, opened 2026-06-10
`feat`: Add (initial) `zvec` target integration

### LINKED ISSUE #1701: [FEATURE] Add `zvec` as a target connector

Haleshot:
Closes #1701

First pass at a `zvec` target connector. zvec runs in-process (just a pip install, no server), so it slotted in cleanly as a target-only connector built on a local `ManagedConnection` over a base directory. The shape ended up close to Qdrant's collection/document target, with the typed `from_class` schema borrowed from SQLite/LanceDB. Each CocoIndex collection is a zvec collection directory, and each row is a `Doc` keyed by a single string id.

A few things I pinned down against the real `zvec==0.4.0` wheel: every collection needs at least one vector field, the primary key has to be a single column since it maps to the string `Doc.id` (so composite keys are rejected), and collection names need 3+ characters, and dense vectors have to be float32 or float16. zvec's index rejects anything wider, so for smaller storage you set `quantize` on the vector (e.g. `"int8"`) instead of reaching for a narrower element type.

Left out `MultiVectorSchemaProvider` for now. zvec's "multi-vector retrieval with reranking" is a query-time thing over named vector fields, not a per-doc _ColBERT-style_ storage type, so there's nothing to map on the write side yet.

Tests, mypy, and `ruff format` are green. `ruff check` only grumbles about the `from ._target import *` line, which every existing connector already does too.

~Opening as a draft for now...~ Marking ready for review.

(Re-opening from a proper feature branch; supersedes #2084, which I accidentally based on my fork's `main`.)

CC: @feihongxu0824, @egolearner

_Edit:_ Updated the document handler to match the newer resolve-at-apply convention from #2120/#2121. It now stores`db_key + collection_name` and resolves `ManagedConnection` from context during apply, instead of carrying the resolved connection from the parent.

### DIFF
--- a/docs/src/content/docs/connectors/zvec.mdx
+++ b/docs/src/content/docs/connectors/zvec.mdx
@@ -1,0 +1,273 @@
+---
+title: "*zvec* connector"
+toc_max_heading_level: 4
+description: >
+  Write documents to zvec, an embedded in-process vector database. Covers
+  collection setup, declaring documents with dense and sparse vectors plus
+  scalar fields, schema-from-class type mapping, and lifecycle management.
+---
+The `zvec` connector writes documents to [zvec](https://zvec.org), an embedded, in-process vector database. zvec runs inside your application — no server or daemon — and stores each collection in a directory on disk.
+
+```python
+from cocoindex.connectors import zvec
+```
+
+:::note[Installation]
+zvec is an optional dependency:
+
+```bash
+pip install cocoindex[zvec]
+```
+:::
+
+## Connection setup
+
+### connect
+
+`connect()` creates a `ManagedConnection` rooted at a base directory. Each collection lives in a subdirectory under it.
+
+```python
+def connect(base_path: str | Path, *, enable_mmap: bool = True) -> ManagedConnection
+```
+
+**Parameters:**
+
+- `base_path` — Directory under which collections are stored. Created if missing.
+- `enable_mmap` — Whether zvec uses memory-mapped I/O for data files.
+
+### ManagedConnection
+
+A handle to the base directory. zvec takes an exclusive write lock per open collection, so `ManagedConnection` caches open handles by collection name and reuses them.
+
+**Methods:**
+
+- `collection_path(name)` — Path to a collection's directory.
+- `close()` — Release all open collection handles (drops their write locks).
+
+For a lifespan, use `managed_connection()`, which closes handles on exit:
+
+```python
+def managed_connection(
+    base_path: str | Path, *, enable_mmap: bool = True
+) -> Iterator[ManagedConnection]
+```
+
+## As target
+
+The `zvec` connector tracks which documents should exist in a collection and automatically handles upserts and deletions. zvec's native upsert is used directly, and documents are removed by id when they are no longer declared.
+
+### Declaring target states
+
+#### Setting up a connection
+
+Create a `ContextKey[zvec.ManagedConnection]` to identify your connection, then provide it in your lifespan:
+
+:::note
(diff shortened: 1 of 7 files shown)


Haleshot:
Hey @badmonster0!

W.r.t https://github.com/cocoindex-io/cocoindex/pull/2084#issuecomment-4661805000: so on the tests front, I ran the connector through a full CocoIndex pipeline against a real on-disk zvec db. I declared rows via `App.update()`, then reopened the collection and read them back with zvec's fetch/query to confirm insert, upsert and delete all sync.


I'd love to put together a proper usecase for this and propose it and later in the `blog/` if you're up for it!

badmonster0:
cool thanks a lot! pls take a look at this https://github.com/cocoindex-io/cocoindex/pull/2120 and make sure it is up to date with latest convetion, appreciate your patience!

Haleshot:
Thanks, I took a look at #2120/#2121. The doc handler was still carrying the resolved connection from the parent, following the _older_ Python connector shape.

Fixed that now: `_DocHandler` stores `db_key + collection_name` and resolves `ManagedConnection` from context in its apply step too.

Kept `from_fn` since zvec is sync/in-process.

Haleshot:
Also, on the blog/use-case idea I mentioned [here](https://github.com/cocoindex-io/cocoindex/pull/2092#issuecomment-4702370821): happy to do that once this is merged.

`zvec` has a nice local-first angle, especially around embedded vector search without running a separate service, so I think there’s something there.

badmonster0:
sg, thanks for the contribution!

PR #2096, opened 2026-06-11
feat(py): enforce single-subscriber on live feeds

georgeh0:
## Summary
- Live-source feeds (`LiveStream` / `LiveMapFeed` / `LiveMapView`, consumed by `mount_each`) are **single-subscriber**: `watch()` owns an exclusive resource (a broker consumer subscription, an OS file watch, an in-memory change channel) that can't be safely fanned out to two concurrent callers. Previously only `LiveMap` guarded against a second `watch()`; localfs, OCI, Kafka, and Iggy would silently misbehave (e.g. racing one consumer's offset commits).
- Add `SingleWatcherGuard` to `connectorkits` (public connector-author helper): a flag + context manager that raises `RuntimeError` on a second concurrent `watch()` and resets on every exit, incl. cancellation. Applied at each exclusive-resource owner via a thin `watch()` → `_watch()` wrapper — Kafka/Iggy `TopicStream` (also covers their `topic_as_map` / `payloads` views), localfs `_LiveDirItems`, OCI `_LiveOCIItems`. `LiveMap` keeps its existing `_watcher_queue` guard.
- Document the contract and helper in `advanced_topics/live_component.mdx`; fix the now-false "not runtime-checked" notes in the Kafka and LiveMap docs.

## Test plan
- New `SingleWatcherGuard` unit tests (reentry raises; resets after normal exit and after exception).
- New Kafka and localfs behavioral single-watcher tests (a second concurrent `watch()` raises).
- CI: full `pytest python/`, mypy, ruff-format.

### DIFF
--- a/docs/src/content/docs/advanced_topics/live_component.mdx
+++ b/docs/src/content/docs/advanced_topics/live_component.mdx
@@ -250,6 +250,24 @@ To add live support to a source connector, make your source return an object tha
 - **`LiveMapView` example**: The [`localfs`](../connectors/localfs) connector's `DirWalker` — `walk_dir(..., live=True).items()` returns a `LiveMapView` backed by `watchfiles`.
 - **`LiveMapFeed` example**: The [`kafka`](../connectors/kafka) connector — `topic_as_map()` returns a `LiveMapFeed` that consumes messages from Kafka topics.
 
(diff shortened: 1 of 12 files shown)


badmonster0:
super cool!

PR #2116, opened 2026-06-15
feat(oci): skip startup full scan on unchanged logic

### LINKED ISSUE #2017: [FEATURE] Leverage component states for live components to skip full processing on rerun

pyjuan91:
Part 2 of #2017. Wires the committed-state primitives from #2078 into the OCI live source.

Adds an opt-in `logic_version` to `oci_object_storage.list_objects()`. When set, the live
source records it via `write_committed_state` after a successful initial scan; on a later
run it reads the value back (before `update_full`) and skips the startup full scan when the
version matches. In skip mode the wall-clock event-time cutoff is disabled, so the durable
stream, resumed from its committed offset, replays the downtime backlog instead of having
it dropped. A changed version, or no prior bootstrap, falls back to a full scan.

`logic_version` defaults to `None` (always scan) and is keyword-only, so existing callers
are unaffected and nothing is persisted unless opted in.

**Scope:** this is the D2 path from #2078 — a *user-declared* logic version
(the user bumps it on pipeline changes and provides a durable stream). The framework-level
logic-change detection you mentioned (auto-detecting changes across the live subtree) remains
the intended follow-up that would close #2017. Flagging so we can decide whether D2 is
acceptable as an interim step, or whether you'd rather go straight to the framework approach.

Refs #2017 (reserving Closes for the framework-detection follow-up).

### DIFF
--- a/docs/src/content/docs/connectors/oci_object_storage.mdx
+++ b/docs/src/content/docs/connectors/oci_object_storage.mdx
@@ -58,6 +58,7 @@ def list_objects(
     path_matcher: FilePathMatcher | None = None,
     max_file_size: int | None = None,
     live_stream: LiveStream[bytes] | None = None,
+    logic_version: str | None = None,
 ) -> OCIWalker
 ```
 
@@ -70,6 +71,7 @@ def list_objects(
 - `path_matcher` — Optional filter for files. Patterns are matched against the relative path (after prefix stripping). See [PatternFilePathMatcher](../common_resources/data_types#patternfilepathmatcher).
 - `max_file_size` — Skip objects larger than this size in bytes.
 - `live_stream` — Optional `LiveStream[bytes]` of OCI Object Storage event payloads. When provided, `OCIWalker.items()` returns a [`LiveMapView`](../advanced_topics/live_component#livemapfeed-and-livemapview) that performs an initial scan and continues watching for changes via the supplied stream. See [Live mode](#live-bucket-watching).
+- `logic_version` — Opt into skipping the startup full scan on reruns. See [Skipping the startup scan](#skipping-the-startup-scan).
 
 **Returns:** An `OCIWalker` that can be used with `async for` loops.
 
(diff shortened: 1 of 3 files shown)


docs/src/content/docs/connectors/oci_object_storage.mdx
```diff
@@ -242,6 +244,27 @@ await coco.mount_each(process_file, walker.items(), target)
...
+)
+```
+
+After a successful scan, the connector records `logic_version`. On a later run, if the recorded version still matches, the scan is skipped and the durable stream is replayed from its committed cursor to catch up on changes that happened while the app was down.
+
+This is safe only when two conditions hold, and **you are responsible for both**:
+
+- **Durable stream.** The stream must resume exactly where it left off — e.g. a Kafka consumer with a stable `group.id` and committed offsets (as in the setup above). Without durable resumption, skipping the scan loses changes that occurred during downtime.
+- **Bump `logic_version` whenever your processing logic changes.** The recorded version is the only signal the connector has that a rescan is needed. If you change your pipeline (e.g. how `process_file` transforms content) without bumping it, the run reuses stale results instead of reprocessing. Leave `logic_version` unset (the default) to always scan on startup.
+
```
georgeh0: I feel this is exposing some implementation details unnecessarily to a user-facing doc.

Since the additional `logic_version` param is only a step stone, and it'll be automatic in the next PR, maybe we don't even need to mention it here.

pyjuan91: Agreed. I'll remove it from the user-facing doc.

PR #2119, opened 2026-06-16
fix(lancedb): escape single quotes in delete filter predicates

Sujit-1509:
## Summary

Fix a SQL injection vulnerability in the LanceDB connector's delete path. String primary-key values containing single quotes (e.g. `"O'Brien"`) were interpolated directly into DataFusion SQL filter predicates without escaping, producing malformed SQL or enabling injection.

## Root Cause

In `_RowHandler._execute_deletes`, string PK values are embedded into a SQL filter string via f-string interpolation:

```python
conditions.append(f"{pk_col} = '{pk_value}'")
```

When `pk_value` contains a single quote, the resulting predicate is broken. For example, the PK `"O'Brien"` produces `id = 'O'Brien'` — invalid SQL. A crafted value like `"x' OR 1=1 --"` would delete every row in the table.

LanceDB's `table.delete(filter)` API accepts raw SQL and has no parameterized variant ([lancedb#1368](https://github.com/lancedb/lancedb/issues/1368)), so escaping is the only defense.

## Fix

Add a 4-line private helper that doubles single quotes per the standard SQL / DataFusion string literal convention:

```python
def _escape_sql_string(value: str) -> str:
    """Escape a string for safe embedding in a DataFusion SQL string literal."""
    return value.replace("'", "''")
```

Applied only in the string branch of `_execute_deletes`. No changes to non-string type handling (int, float, bool, etc.).

## Tests

- **`test_escape_sql_string_doubles_single_quotes`** — Unit test verifying the helper on normal strings, quotes, empty strings, and backslashes.
- **`test_execute_deletes_escapes_string_pk`** — Unit test with a mock table that captures the filter string, asserting `"O'Brien"` produces `id = 'O''Brien'`.
- **`test_execute_deletes_with_real_lancedb`** — Integration test that inserts a row with PK `"O'Brien"` into a real LanceDB table, deletes it via `_execute_deletes`, and verifies the row is removed while other rows remain.

### DIFF
--- a/python/cocoindex/connectors/lancedb/_target.py
+++ b/python/cocoindex/connectors/lancedb/_target.py
@@ -319,6 +319,11 @@ class TableSchema(Generic[RowT]):
         return columns
 
 
+def _escape_sql_string(value: str) -> str:
+    """Escape a string for safe embedding in a DataFusion SQL string literal."""
+    return value.replace("'", "''")
+
+
 class _RowAction(NamedTuple):
     """Action to perform on a row."""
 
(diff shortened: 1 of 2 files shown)


georgeh0:
Thanks for the fix!

Issue #2039, opened 2026-05-29
perf(engine): minimize serialization related to UserStateCache


hakuuww:
## Overview

`UserStateCache` (introduced in #2034) currently stores state values as raw `Vec<u8>`. This causes two unnecessary costs per build:

**1. Wasted serialization of `initial_value`**

Every `coco.use_state()` call serializes `initial_value` before passing it to Rust — even on runs 2+ where a stored value already exists and `initial_value` is immediately ignored.

**2. Expensive clones**

When transitioning a `Loaded` entry to `Declared`, two owned `Vec<u8>` copies are needed (one for the map, one to return to the caller). Each is a heap allocation + full buffer copy — O(n) in the size of the serialized value.

## Proposed Design

Keep `Loaded(Vec<u8>)` as raw bytes — entries come straight from LMDB and don't need to be deserialized until actually accessed. Change `Declared` to hold `Prof::FunctionData` (= `PyStoredValue` in the Python profile):

```rust
enum UserStateEntry<Prof: EngineProfile> {
    Loaded(Vec<u8>),                 // raw bytes from DB; deserialized lazily on first access
    Declared(Prof::FunctionData),    // live value; serialized once at flush
}
```

`PyStoredValue` is the right abstraction here — it already holds both representations lazily (`bytes` and `object`) behind an `Arc<Mutex<...>>`, so:

- Cloning is always a cheap `Arc` refcount bump regardless of which representation is held
- Deserialization from bytes happens once on first access and is cached
- Serialization to bytes happens once at flush time and is cached
- All clones of the same entry share the same cached deserialized object

This mirrors how `FnCallMemo<Prof>` uses `Prof::FunctionData` for memoized return values and state.

**Lifecycle with this change:**

- When `use_state()` claims a `Loaded` key: wrap the bytes in `PyStoredValue::from_bytes(...)`, transition to `Declared`. The Python object is deserialized lazily on first `.get()` call.
- When `use_state()` is called for a new key: wrap `initial_value` in `PyStoredValue::new(py_obj)`, store as `Declared` — no serialization.
- `handle.value = x`: store `x` as `PyStoredValue::new(py_obj)` — no serialization until flush.
- `flush_to_db()`: call `.to_bytes()` on each `Declared` entry — serializes once and caches.

## Reference

TODO comment left in [`engine/context.rs`](https://github.com/cocoindex-io/cocoindex/pull/2034).

badmonster0:
[...]

PR #2127, opened 2026-06-17
perf(engine): minimize serialization related to UserStateCache

hakuuww:
Implements #2039, follows #2034, part of #2015
Please review commit-by-commit.

## Background: the waste

`UserStateCache` (added in #2034) stored each state value as a serialized `Vec<u8>`. But the serialized form is only needed when persisting to LMDB — storing it in the in-memory cache forced (de)serialization at the Python↔Rust boundary on every interaction. The cache has a two-state per-key lifecycle:

- `Loaded` — value from the previous component run (bytes read from LMDB).
- `Declared` — value claimed/written this run, which only needs to be persisted at commit.

Holding both as raw `Vec<u8>` caused four unnecessary costs per build:

1. **Wasted `initial_value` serialization** — every `use_state()` serialized `initial_value` eagerly, even on reload runs where a stored value already exists and the `initial_value` is immediately discarded.
2. **Wasted per-write serialization** — every `handle.value = x` serialized `x`, even though only the final value at commit is persisted. A hot loop that repeatedly updates a state would pay a full serialize per iteration:

   ```python
   counter = coco.use_state("counter", 0)
   for row in rows:               # e.g. tens of thousands of rows
       counter.value += 1         # before: serialize() on EVERY iteration
                                  # after:  no serialization until commit (only once)
   ```

   The cost scaled with `iterations × value size`, so it hurt most exactly where state is used most — large accumulators (dicts/lists/ndarrays) rebuilt in a loop were re-encoded in full on every assignment.
3. **First-run round-trip** — the initial was serialized, returned as bytes, then immediately deserialized back into the handle as python object.
4. **Expensive clones** — transitioning `Loaded → Declared` deep-copies the `Vec<u8>` twice (one for the map, one to return), each an O(n) heap copy.

## The idea

Store the value as `PyStoredValue` (= `Prof::FunctionData` in the Python profile) — the same lazy dual-representation cell already used for fn-memo values. It holds `bytes` and/or a Python `object` behind `Arc<Mutex<…>>`:

- clone = cheap `Arc` refcount bump (no buffer copy);
- deserialize bytes→object once, on first read, cached;
- serialize object→bytes once, at commit, cached;
- all clones of an entry share the same cached representations.

So reusing a state during a run touches no serialization at all, and bytes only ever exist for what actually goes to LMDB.

## Commits

1. **refactor(engine): store user state as `Prof::FunctionData` instead of raw bytes** — behavior-preserving plumbing. `UserStateCache<D>` / `UserStateEntry<D>` become generic over the stored value type (instantiated as `Prof::FunctionData`); `populate` decodes via `D::from_bytes`, and `into_flush_plan` serializes each surviving `Declared` via `D::to_bytes` once, at commit, keeping `CommitPlan`/`submit_session` byte-based and untouched. The PyO3 boundary still serializes eagerly — the value just travels internally as a bytes-backed cell. (Both `Loaded` and `Declared` hold `D`; a bytes-backed `PyStoredValue` for `Loaded` keeps deserialization lazy without needing a separate `Vec<u8>` variant.)

2. **perf(py): defer user-state serialization on `.value` writes** — `update_user_state` takes the Python object and wraps it in an object-backed cell without serializing; the `StateHandle` setter drops the eager serialize. N writes in one run cost a single encode of the final value, at commit. (Fixes cost **2**.)

3. **perf(py): defer user-state read serialization (lazy initial + no round-trip)** — `use_state` takes the initial as an object (dropped unserialized when a value is already stored), and both methods now return the `StoredValue` cell; `StateHandle` reads through it via `.get(deserialize_fn)`. First-run reads hand back the original object (no round-trip); a reloaded value is deserialized at most once, on first access, and cached. (Fixes costs **1, 3, 4**.)

## Design notes

- `UserStateCache` is generic over just the value type `D: Persist + Clone`, not the full `EngineProfile` — the cache only needs to serialize/clone the value, so this keeps the bound minimal and lets the unit tests use a tiny `TestData(Vec<u8>)` Persist newtype instead of a full engine profile.
- Serialization is now deferred to commit, which shifts two behaviors (both consistent with the existing fn-return / memo path): a non-serializable value surfaces its error at commit rather than at the `handle.value = ...` line (the error names the offending state key), and the persisted value reflects in-place mutations made after assignment. `into_flush_plan` attaches the key to the serialization error, and a test asserts a non-serializable value fails at commit (via the exception handler, not silently) with the key in the message.

## Testing

- **Rust**: the existing `UserStateCache` unit tests are retargeted onto the `TestData` Persist newtype (flush-plan assertions still compare raw bytes).
- **Python** (`test_use_state.py`), new behavior-level guards:
  - 5 `.value` writes in one run → serialized exactly once;
  - an `initial_value` passed on a reload run → not serialized;
  - first run → handle returns the same object, with zero (de)serialization during the run and one serialize at commit;
  - a reloaded value → deserialized exactly once across repeated reads.

## Scope: which profile benefits

This optimization targets the **Python profile** (`coco.use_state()`), where `Prof::FunctionData = PyStoredValue` — a lazy dual-representation (object/bytes) cell. Only that profile gets the deferred/lazy behavior: no per-write serialize, no first-run round-trip, lazy reload deserialize.

The **Rust SDK** (`rust/sdk/cocoindex`) uses `FunctionData = Value`, a thin newtype over MessagePack bytes with no object form, so it serializes eagerly at the boundary exactly as it did before — it does **not** benefit from this optimization. The core change only makes `use_state`/`update_user_state` generic over `Prof::FunctionData`, which is what lets each profile pick its representation. After merging main, the SDK callers were adapted to pass/return `Value` via `Value::from_serializable`/`Value::deserialize`; this is byte-for-byte equivalent to the prior `encode_state`/`decode_state` (`rmp_serde`), i.e. **no serialization was added or removed for the SDK**. Giving the Rust SDK the same lazy behavior would require a dual-representation `Value` and is out of scope for this PR (#2039 is about the Python user-state cache).


hakuuww:
WIP

hakuuww:
One caveat of the current change is that a "non serializable exception" can only be raised during the commit time of the related component(s). 

Previously, it was raised immediately when assigning a non serialized value to the user state, with the line number that caused the exception for debug context.

`state.value = non_serializable_value`

With this change, the exception is raised during commit time of the component like in the following error message, with the key name being provided, but not the line number. 

```
----- TYPE -----
     (<class 'RuntimeError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
     ----- STR -----

     Context:
       1: failed to serialize user state for key Str("bad_key")
     Error calling Python function:
     NotImplementedError: Cannot serialize _Unserializable

     ----- REPR -----
     ----- REPR -----
     RuntimeError('\nContext:\n  1: failed to serialize user state for key Str("bad_key")\nError calling Python function:\nNotImplementedError: Cannot serialize _Unserializable\n')
     
```

python/tests/core/test_use_state.py
```diff
@@ -136,6 +160,190 @@ async def _root_no_initial(items: list[str]) -> None:
...
+        items=_source_items,
+    )
+    _source_items[:] = ["a"]
+
+    app.update_blocking()  # first run: stores the value
+    app.update_blocking()  # second run: reload, read .value three times
+    assert _info["a"] == 1  # deserialized exactly once despite three reads
+
+
+def test_use_state_unserializable_value_errors_at_commit_with_key() -> None:
```
hakuuww, line 310: Assert for unserializable value at commit tim with key.

hakuuww:
Ready for review

badmonster0:
thanks @hakuuww , @georgeh0 can help take a look!

python/cocoindex/_internal/api.py
```diff
@@ -92,8 +92,7 @@
 from .serde import (
     unpickle_safe,
     serialize_by_pickle,
-    serialize as _serialize,
-    deserialize as _deserialize,
+    make_deserialize_fn as _make_deserialize_fn,
```
georgeh0: nit: we don't need to give a private name, since the module is already under `_internal`.

georgeh0 (approved):
Thanks! Overall looks quite good except a minor style comment.

python/cocoindex/_internal/api.py
hakuuww: Thanks! Amended the fix into the last commit.

hakuuww:
Fixed merge conflicts related to the upstream Rust SDK changes. 

Note: This PR targets the Python profile, see the last section of the PR description for details.

georgeh0:
Thanks a lot! Merged!

PR #2115, opened 2026-06-14
feat: LanceDB vector and FTS indexing support

Sujit-1509:
Implementation of #1352 and #1051: Adds Vector and FTS Indexing Support to the LanceDB Connector

### Overview
This PR implements declarative index definitions for LanceDB tables, aligning the LanceDB connector with the core \VectorIndexSpec\ and \FtsIndexSpec\ patterns used in other target connectors.

### Changes Included
* **Vector Index Support**: Implemented \_VectorIndexHandler\ supporting IVF-PQ and HNSW-PQ algorithms. Configured proper distance/metric handling adapted for the updated LanceDB 0.33 \IndexConfig\ API.
* **FTS Index Support**: Implemented \_FtsIndexHandler\ allowing simple text column definitions with language detection and positioning.
* **Target Table Integration**: Exposed \declare_vector_index()\ and \declare_fts_index()\ on the \lancedb.TableTarget\ component to allow inline, declarative index specification from user-side applications.
* **Testing Updates**: Extended LanceDB tests to properly isolate \_VectorIndexHandler\ and \_FtsIndexHandler\ using handler-level direct invocation to verify LanceDB interactions. Handled \IndexConfig\ API updates where objects are now returned instead of dictionaries.

Fixes #1352
Fixes #1051

PR #2129, opened 2026-06-19
fix(lancedb): physically drop indices on declarative removal

Sujit-1509:
## Summary

Completes the declarative lifecycle for LanceDB vector and FTS indices introduced in #2115.

Currently, removing a declared index from the target configuration produces a delete action during reconciliation, but both `_VectorIndexHandler` and `_FtsIndexHandler` treat index deletion as a no-op. As a result, indices persist indefinitely even after being removed from the desired state.

## Root Cause

The handlers were implemented under the assumption that LanceDB did not expose an index deletion API on `AsyncTable`.

However, `AsyncTable.drop_index()` and `create_index(..., name=...)` are available across the entire LanceDB version range supported by CocoIndex:

| Tag | `async def drop_index`? | `create_index(name=)`? |
|---|---|---|
| v0.24.1 | ✅ | ✅ |
| v0.25.0-beta.0 | ✅ | ✅ |
| v0.26.0 (effective floor) | ✅ | ✅ |
| v0.33.0 | ✅ | ✅ |

`pyproject.toml` pins `lancedb>=0.25.0`; since no `v0.25.0` stable tag exists, the constraint resolves to **v0.26.0** as the lowest satisfying release.

## Changes

- Assign explicit physical LanceDB index names when creating vector and FTS indices.
- Namespace physical index names by attachment type (`coco_idx_vector_` / `coco_idx_fts_`) to avoid cross-type collisions — `vector_index` and `fts_index` are separate CocoIndex attachment providers but share the same per-table LanceDB index namespace.
- Replace the existing deletion no-op with `drop_index(...)`, tolerating a missing index so reconciliation stays idempotent.
- Add lifecycle tests covering create → delete behavior for both vector and FTS indices, a coexistence test for same-named vector + FTS indices, and delete-on-absent tests.

## Behavior Change

Indices are now physically removed when their corresponding declarative definitions are removed from the target configuration, restoring the expected convergence behavior of the attachment system.

Physical LanceDB index names are now explicitly managed by CocoIndex (`coco_idx_vector_{name}` / `coco_idx_fts_{name}`) instead of relying on LanceDB-generated defaults (`{column}_idx`).

### Notes for reviewers

- No existing code or tests depend on the old auto-generated index names (tests use substring matching).
- No persisted tracking records, migrations, or LMDB keys reference physical LanceDB index names — the fingerprints are derived from the index specs (`column`, `metric`, `index_type`, ...), not the physical name.
- Per the `drop_index` docs, dropped indices are reclaimed from disk by a subsequent `table.optimize()`, which the connector already runs periodically.

### DIFF
--- a/python/cocoindex/connectors/lancedb/_target.py
+++ b/python/cocoindex/connectors/lancedb/_target.py
@@ -64,6 +64,12 @@ _RowValue = dict[str, Any]  # Column name -> value
 _RowFingerprint = bytes
 ValueEncoder = Callable[[Any], Any]
 
+# Prefix for physical LanceDB index names to avoid cross-type collisions
+# (vector_index and fts_index are separate CocoIndex attachment providers
+# sharing the same per-table LanceDB index namespace).
+_VECTOR_INDEX_NAME_PREFIX = "coco_idx_vector_"
+_FTS_INDEX_NAME_PREFIX = "coco_idx_fts_"
+
 
 class LanceType(NamedTuple):
     """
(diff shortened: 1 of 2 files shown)


badmonster0:
hey @Sujit-1509 thanks a lot for the contribution! pls refer to the official impl from lancedb maintainer - https://github.com/cocoindex-io/cocoindex/pull/2132 @prrao87  - hope this make things easier! please let us know if there's any questions

prrao87:
Hi @Sujit-1509 many thanks for this - I was anyway fixing a few other things and coincidentally fixed the same issue. We can consider this a duplicate of https://github.com/cocoindex-io/cocoindex/pull/2132. I think this can be closed @badmonster0.

badmonster0:
thanks both!! @Sujit-1509 @prrao87

PR #2130, opened 2026-06-19
docs(lancedb): fix callout syntax, document vector/FTS indexes, tighten prose

prrao87:
## Summary

Documentation-only improvements to the LanceDB connector page (`docs/src/content/docs/connectors/lancedb.mdx`). No code or dependency changes.

### What changed

1. **Callout box syntax** — the Dependencies callout used `::::note Dependencies` (four colons + space-separated title, which dropped the title). Corrected to `:::note[Dependencies]`, matching every sibling connector page.

2. **Documented vector and FTS indexes** — `declare_vector_index` (IVF-PQ / HNSW-PQ params + metrics) and `declare_fts_index` (language, `with_position`) were entirely undocumented. Added them as `(attachment)` subsections, mirroring the postgres/surrealdb/neo4j/falkordb convention, with a cross-link to the custom-target-connector attachment guide. Also wired a vector-index declaration into the end-to-end example.

3. **Tightened prose** — expanded the intro to accurately describe the connector (schema inference, index declaration, lifecycle management) instead of just "writing rows"; removed the redundant "target state APIs" overlap between the intro and the "As target" section; and de-duplicated the VectorSchemaProvider links.

### Note on index removal behavior

The page documents that **removing an index declaration is currently a no-op** — the index persists until the table is rebuilt. This is accurate to the connector's current behavior ([the delete branch is a no-op](https://github.com/cocoindex-io/cocoindex/blob/main/python/cocoindex/connectors/lancedb/_target.py)). A follow-up PR will implement real index dropping via LanceDB's `AsyncTable.drop_index` (available in newer LanceDB) and bump the `lancedb` dependency floor accordingly; the docs note will be updated to match at that time.

### DIFF
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -12,14 +12,14 @@ The `lancedb` connector provides target state APIs for writing rows to LanceDB t
 from cocoindex.connectors import lancedb
 ```
 
-::::note Dependencies
+:::note[Dependencies]
 This connector requires additional dependencies. Install with:
 
 ```bash
 pip install cocoindex[lancedb]
 ```
 
-::::
+:::
 
 ## Connection setup
 
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -110,9 +110,9 @@ def TableTarget.declare_row(
 
 - `row` — A row object (dict, dataclass, NamedTuple, or Pydantic model). Must include all primary key columns.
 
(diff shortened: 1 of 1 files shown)


prrao87:
@badmonster0 From [the custom connector guide](https://cocoindex.io/docs/advanced_topics/custom_target_connector/), I understood attachments as auxiliary target states a child handler manages alongside its regular children (like indexes on a table), each tracked independently. So that's why I assigned this category to the vector/FTS indexes rather than being nested child rows of the table.

Hope that makes sense, feel free to suggest corrections if not!

badmonster0:
ty for the contribution!

PR #2132, opened 2026-06-19
feat(lancedb): implement drop_index; bump lancedb to 0.33.0 and pyarrow to 23.0.0

### LINKED ISSUE #2131: [FEATURE] Implement LanceDB drop_index API
**What is the use case?**
Per https://github.com/cocoindex-io/cocoindex/pull/2130, we're stating in the note that removing an index declaration is currently a no-op because LanceDB 0.27.0 didn't support `drop_index` via its async API.

`drop_index` support has now been added in v0.33.0 (latest version) of LanceDB, so this can benefit CocoIndex users of this connector too, making its behaviour consistent with other similar DBs.

@prrao87 will implement the fix and send a PR.

**Additional context**
This will allow LanceDB's index-dropping behaviour to mirror other solutions, so that users don't get surprised when they change the declaration in their CocoIndex code but the index still gets left behind.

prrao87:
## Summary

Implements real index dropping for the LanceDB connector and bumps the connector's dependencies to support it.

Fixes #2131.

### Dependency bumps
- **lancedb** `>=0.25.0` → `>=0.33.0` — `AsyncTable.drop_index` was added in newer LanceDB (absent in 0.27.0).
- **pyarrow** `>=19.0.0` → `>=23.0.0`.

Both bumped in the `lancedb` and `all` extras.

### Behavior change: index removal now drops the index
Previously, removing a `declare_vector_index` / `declare_fts_index` declaration was a **no-op** — the index lingered until the table was rebuilt. This surprised users whose code no longer declared an index but whose table still carried it. Now the connector drops it, mirroring the postgres/surrealdb connectors.

Implementation details:
- Pass an explicit `name=` to `create_index`, so the LanceDB index name matches the connector's tracking name. (Previously the index was auto-named, so a user-supplied `name=` never actually reached LanceDB.)
- Replace both no-op delete branches with a shared `_drop_index_if_exists` helper that checks `list_indices()` and drops by name — a missing index is a clean no-op rather than an exception (per the repo's "no exceptions for not-found" convention).
- `drop_index` detaches the index immediately; on-disk reclamation happens on the next `table.optimize()`, which the row handler already schedules periodically.

### Tests
Added delete-path tests for both index handlers (`test_vector_index_handler_drops_index_on_delete`, `test_fts_index_handler_drops_index_on_delete`), verifying the index disappears from `list_indices()` after a delete action and that a repeated drop is a safe no-op. Full suite: **28 passed**. mypy + ruff clean; docs build clean.

### ⚠️ Relationship to #2130 (docs PR)
This branch **includes the three docs commits from #2130** (callout fix + vector/FTS index documentation), because it needs that documentation present in order to update the index-reconciliation note from "removal is a no-op" to "removal drops the index."

**Recommended merge order:** merge #2130 first, then rebase this branch — the duplicate docs commits drop out cleanly, leaving only the `feat` commit + the one-line note update. Alternatively, this PR can supersede #2130 if you'd prefer a single PR.

### DIFF
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -12,14 +12,14 @@ The `lancedb` connector provides target state APIs for writing rows to LanceDB t
 from cocoindex.connectors import lancedb
 ```
 
-::::note Dependencies
+:::note[Dependencies]
 This connector requires additional dependencies. Install with:
 
 ```bash
 pip install cocoindex[lancedb]
 ```
 
-::::
+:::
 
 ## Connection setup
 
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -179,7 +179,7 @@ table.declare_fts_index(column="content")
 ```
 
 :::note
-Indexes are reconciled as part of the table's target state: changing a declaration replaces the index in place, and dropping the table removes its indexes. Removing only an index declaration is currently a no-op — the index persists until the table is rebuilt.
+Indexes are reconciled as part of the table's target state: changing a declaration replaces the index in place, removing a declaration drops the index, and dropping the table removes all its indexes.
 :::
 
 ### Table schema: from Python class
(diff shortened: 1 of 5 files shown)


prrao87:
@cocoindex-dev I faced some minor surprises when working with the lancedb connector recently, so this PR with the latest LanceDB release will hopefully benefit other users of this connector too :)

badmonster0:
lg! pls take a look at merge conflict and ready to go

prrao87:
Done, thanks!

Issue #2128, opened 2026-06-18
[FEATURE] API to invalidate memoization for the current call stack (for live components, source read)


georgeh0:
Memoization assumes a function's observable effect is fully captured by its return value (plus its declared target states). But sometimes there're dependencies or effects that live *outside* that:

- Live components: which has live updates over time
- Source read: data read through most sources provided by CocoIndex are supposed to change over time

We want to build an API to invalidate memoization for the entire call stack. The API should be called by the operations above, and will also be used for similar cases in the future.

PR #2160, opened 2026-06-21
feat(core): add API to invalidate memoization for the current call stack

thatditsyboy:
Refs #2128.

## Problem

Memoization assumes a function's observable effect is fully captured by its return value plus its declared target states. That assumption breaks for operations whose result changes over time — the two called out in #2128:

- **Live components** — live updates over time
- **Source reads** — most CocoIndex sources are expected to change over time

There was no way to tell the engine "this call's result isn't stable; don't memoize it (or its callers)."

## Solution

A single, minimal public API:

```python
def invalidate_memoization() -> None: ...
```

Called from within a function, it marks the current call as non-memoizable and **propagates up the entire call stack**, so no caller memoizes this run and the whole stack re-executes next run. This deliberately mirrors the existing `has_child_components` mechanism (a flag on `FnCallContextInner`, OR-merged in `join_child`), so it composes with the established memoization model rather than adding a parallel one.

Crucially, the flag is honored at the **single memo-storage choke point** — `build_fn_call_memo` returns `None` when invalidated, yielding `Ready(None)` — so it covers every memoization path (SDK `cached_by_fingerprint*`, the Python `@coco.fn(memo=True)` path, etc.) without touching each one. Unlike `has_child_components`, it is a non-erroring, intentional opt-out.

## Changes

| Layer | Change |
|---|---|
| `rust/core` | `FnCallContext::invalidate_memo()` + `memo_invalidated` flag on `FnCallContextInner`; propagated in `join_child`; honored in `build_fn_call_memo` |
| `rust/py` | `PyFnCallContext::invalidate_memo()` binding |
| `python` | `coco.invalidate_memoization()` public API; `core.pyi` stub |
| tests | Rust unit test asserting up-the-stack propagation |

Net: **+92 lines, 0 deletions.**

## Usage

```python
@coco.fn(memo=True)
async def read_and_process(src: SomeLiveSource) -> Result:
    data = await src.read()        # changes over time
    coco.invalidate_memoization()  # don't memoize this call or its callers
    return process(data)
```

## Verification

- `cargo test -p cocoindex_core` — new unit test `memo_invalidation_propagates_up_the_call_stack` passes; core builds clean.
- `cargo build -p cocoindex_py` — PyO3 binding compiles (Python 3.12).
- `ruff format --check`, `ruff check`, `mypy` — all clean on the changed Python.

## Scope / follow-up

This PR delivers the **API and the engine mechanism**. Per the "minimize API surface" guideline in AGENTS.md, it does not pre-wire every operation. Wiring the concrete call sites (live components, the source-read paths in the SDK/connectors) to call this API is a natural, focused follow-up — happy to do that here or in a separate PR, whichever the maintainers prefer.


badmonster0:
hey @thatditsyboy please follow https://cocoindex.io/docs/contributing/guide/ guide, we normally don't take pr unless it is assigned and contributor normally take one at a time. this is an ticket that has been assigned internally so i'll close it. thanks!

thatditsyboy:
Understood, and sorry for the unsolicited PR. I'll follow the contributing guide and the assign-first, one-at-a-time flow going forward. Apologies for duplicating work that was already assigned internally.

Issue #2125, opened 2026-06-17
[FEATURE] Support Apache Pulsar source and target connectors


georgeh0:
Similar to [Kafka](https://github.com/cocoindex-io/cocoindex/tree/main/python/cocoindex/connectors/kafka) and [Iggy](https://github.com/cocoindex-io/cocoindex/tree/main/python/cocoindex/connectors/iggy).

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

thatditsyboy:
Hi @badmonster0 — I'd like to take this on. Following the one-at-a-time flow, could you assign me for the **Pulsar target connector** (#2161) first? I'll get it verified end-to-end against a real broker, and once it's in I'll follow up with the source connector (#2162). Thanks!

maohsiah:
Hi @georgeh0 and @thatditsyboy  — is #2125 still available for an extra contributor? Thank you for your time.

badmonster0:
assigned @maohsiah ! thanks for taking up this. we closed https://github.com/cocoindex-io/cocoindex/pull/2162 because @thatditsyboy cannot run pulsar in his environment. please feel free to take from there and let us know if you have any questions!

PR #2161, opened 2026-06-21
feat(connectors): add Apache Pulsar target connector (Python)

thatditsyboy:
Part of #2125 (target connector; source/consumer to follow).

## Summary

Adds an Apache **Pulsar target connector** for the Python SDK, mirroring the existing **Kafka** and **Iggy** target connectors. Declared target states are produced as messages to a user-managed Pulsar topic, with the target-state key used as the Pulsar **partition key**. A message is produced only when a state changes since the last run (tracked by a content fingerprint), and deletes are handled (empty payload by default, or a custom marker via `deletion_value_fn`).

```python
import pulsar
import cocoindex as coco
from cocoindex.connectors import pulsar as pulsar_connector

PULSAR = coco.ContextKey[pulsar.Client]("pulsar")

target = await pulsar_connector.mount_pulsar_topic_target(
    PULSAR, topic="persistent://public/default/derived-events"
)
target.declare_target_state(key="order-123", value=b'{"status":"ready"}')
```

## Design

Same two-level handler model as Kafka/Iggy (`_TopicHandler` → `_MessageHandler`), registered via `register_root_target_states_provider` — no engine/Rust changes. The one adaptation: the `pulsar-client` Python client is **synchronous with per-topic producers**, so the connector takes a single `pulsar.Client` via a `ContextKey`, lazily creates and caches one producer per topic, and sends off the event loop with `asyncio.to_thread`. The `reconcile` change-detection logic (fingerprint upserts, tombstone/`deletion_value_fn` deletes) is identical to Kafka's.

## Changes

- `python/cocoindex/connectors/pulsar/{__init__,_target}.py` — the connector + public API (`pulsar_topic_target`, `declare_pulsar_topic_target`, `mount_pulsar_topic_target`, `PulsarTopicTarget`).
- `pyproject.toml` + `uv.lock` — `cocoindex[pulsar]` optional dependency (`pulsar-client>=3.0.0`), added to the `all` extra and mypy ignore list.
- `docs/src/content/docs/connectors/pulsar.mdx` + sidebar entry.
- `python/tests/connectors/test_pulsar_target.py` — unit tests.

## Verification

- `pytest python/tests/connectors/test_pulsar_target.py` — **17 passed**. Tests mock the Pulsar client (like the Kafka target tests), so **no broker is required** and they're CI-safe.
- `mypy` — clean on the connector and tests.
- `ruff format --check` — clean.

## Scope / follow-ups

This PR is the **target** side only, per an incremental approach. Natural follow-ups: the Pulsar **source** (consumer, mirroring `kafka`/`iggy` `topic_as_map`/`topic_as_stream`) and an optional Rust SDK Pulsar target. Happy to do those next.

### DIFF
--- a/docs/src/content/docs/connectors/pulsar.mdx
+++ b/docs/src/content/docs/connectors/pulsar.mdx
@@ -1,0 +1,81 @@
+---
+title: "*Apache Pulsar* connector"
+sidebar_label: Pulsar
+description: >
+  Produce messages to Apache Pulsar topics as target states.
+---
+
+The `pulsar` connector supports [Apache Pulsar](https://pulsar.apache.org/docs/)
+as a **target**: declared target states are produced as messages to a
+user-managed Pulsar topic, with the target-state key used as the message
+[partition key](https://pulsar.apache.org/docs/concepts-messaging/#routing-modes).
+
+```python
+from cocoindex.connectors import pulsar
+```
+
+Install the optional dependency:
+
+```bash
+pip install cocoindex[pulsar]
+```
+
+The connector expects a `pulsar.Client` that you create and provide through your
+app context. Topics are user-managed: CocoIndex does not create or drop them.
+
+## As Target
+
+The target connector sends bytes or strings to a user-managed Pulsar topic. The
+`pulsar.Client` is provided via a [`ContextKey`](/core/context); the connector
+creates and caches one producer per topic and sends off the event loop.
+
+```python
+import pulsar
+
+PULSAR = coco.ContextKey[pulsar.Client]("pulsar")
+
+target = await pulsar.mount_pulsar_topic_target(
+    PULSAR,
+    topic="persistent://public/default/derived-events",
+)
+
+target.declare_target_state(key="order-123", value=b'{"status":"ready"}')
+```
+
+On each commit, CocoIndex produces a message only when a declared target state
+has changed since the last run (tracked by a content fingerprint).
+
+### Deletions
+
+By default a deleted target state produces a message with an empty payload (and
+the key as the partition key). For key-based compaction, use a
+[compacted topic](https://pulsar.apache.org/docs/concepts-topic-compaction/).
+Pass `deletion_value_fn` to emit a custom delete marker instead:
+
+```python
+target = await pulsar.mount_pulsar_topic_target(
+    PULSAR,
+    topic="persistent://public/default/derived-events",
+    deletion_value_fn=lambda key: f'{{"id":{key!r},"deleted":true}}',
+)
+```
+
+### Target APIs
+
+```python
+def declare_pulsar_topic_target(
+    client: ContextKey[pulsar.Client],
+    topic: str,
+    *,
+    deletion_value_fn: DeletionValueFn | None = None,
+) -> PulsarTopicTarget[PendingS]
+```
+
+```python
+async def mount_pulsar_topic_target(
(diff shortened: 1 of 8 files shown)


badmonster0:
hi @thatditsyboy could you share how did you test the feature?

thatditsyboy:
On how I tested the target connector (also in the PR's Verification section):

- `python/tests/connectors/test_pulsar_target.py` is 17 unit tests that mock `pulsar.Client` / `Producer`, the same approach as the existing Kafka target tests, so they need no live broker and are CI-safe. They cover the reconcile change-detection (fingerprint upsert, no-op when unchanged), the partition key being the target-state key, delete handling (empty tombstone and a custom `deletion_value_fn`), and the `asyncio.to_thread` offload of the synchronous client. `mypy` and `ruff format --check` are clean.
- I did not add a real-broker integration test, so the feature is verified against mocks only, not a live Pulsar broker.

Separately, following the guide note from #2160: I realize #2161 and #2162 are also unassigned, and I opened more than one at once, which goes against the assign-first / one-at-a-time flow. Apologies for that. If the Pulsar work under #2125 is open to outside contributors, I'm happy to be assigned first and take it one PR at a time. Otherwise feel free to close these and I'll step back. Your call.

badmonster0:
hey @thatditsyboy 
- please comment on issue so we can assign it to you 
- code looks good, please verify it works end to end in real env and we can approve it, thanks!

thatditsyboy:
Thanks! I commented on #2125 to get assigned for the target connector first.

On the end-to-end ask: I want to be upfront rather than claim a real-env run I didn't do. My current environment has no Docker, container runtime, or Java, so I can't stand up a real Pulsar broker here to capture genuine e2e output.

What I've done instead is add a live e2e test (27bfac7): `python/tests/connectors/test_pulsar_target_live.py`. It drives the real produce path through `_MessageHandler._apply_actions` against a broker and reads the topic back to assert the payload, the target-state key used as the Pulsar partition key, and the empty-payload tombstone on delete. It's gated with `skipif` on `PULSAR_SERVICE_URL`, so the default broker-less suite is unaffected, and the docstring has the one-liner to run it against a Pulsar standalone container:

```
docker run -d --name pulsar -p 6650:6650 -p 8080:8080 apachepulsar/pulsar:3.3.1 bin/pulsar standalone
PULSAR_SERVICE_URL=pulsar://localhost:6650 pytest python/tests/connectors/test_pulsar_target_live.py -v
```

It mirrors the verified mocked tests' handler-driving pattern, so I'm fairly confident in it, but I have not executed it against a live broker myself. Could you run it (or point CI at a broker)? Happy to fix anything it surfaces. If you'd prefer to verify it on your side before approving, that works too.

badmonster0:
@thatditsyboy thanks! we do expect our contributor to test the code e2e. if this is something you cannot do, please let me know and we will close the PR, thank a lot for your contribution and taking a look at the issue!

thatditsyboy:
Totally fair, and thanks for being clear about the expectation. I can't run a real Pulsar broker from my current environment (no Docker or Java) and I don't have another broker to point at, so I can't give you the e2e verification you reasonably require. Closing this on my end to save you the step.

The gated live test I added is there if it's useful to whoever picks the issue up next. Thanks for the careful review and for taking the time, sorry I couldn't take it all the way.

badmonster0:
@thatditsyboy thanks a lot for giving this issue a shot! Really appreciate it!

PR #2192, opened 2026-06-22
feat: framework logic-change detection for live components

pyjuan91:
The core change is the engine can detect if there is any logic in this live component's subtree changing by itself. This replaces the `logic_version` approach from #2116 (where the user had to manually bump a version string whenever their transform changed, and they might forget the bump and the scan gets silently skipped, keeping stale results.)

On the connector side (OCI, both Python and Rust SDK), the skip-scan check changes from comparing with old version to processing unchanged.

**Why the engine couldn't answer this before**

Live children are built in isolated contexts, and the live drain path (`HandleOutcome`) carries no `logic_deps`. So the deps of the per-item transforms only live in the sub-components, nothing aggregates upward, and the live component has nothing to read back on restart.

**How the aggregation works**

The subtree's deps are collected as a union set on the write path:

- `update_full` **replaces** the set, so fingerprints of edited/deleted code drop out.
- Incremental `update` **unions** into it, a streamed item still exists on the next restart, so its code has to stay tracked.
- `processing_unchanged()` reads the set back and runs `all_contained_with_env()`: if any fingerprint is no longer registered, some code changed → re-scan. It also returns false when there's no stored set or it can't be decoded, so the failure direction is always "scan again," not "silently skip."

One thing this depends on: #2142 (already merged) made stored memos cover their full subtree, so a cache-hit child still contributes its whole closure. Without that, editing a mounted grandchild function would wrongly pass the skip gate.

Following George's suggestion, the aggregation set reuses the component's existing `ComponentMemoizationInfo` DB entry instead of user state (this is framework change detection info, not user state). A live root has no `processor_fp` (it's not a memoized function call) and no `return_value`, so those two fields (serde renames `"F"` / `"R"`) become `Option`, and the regular memo write path now wraps them in `Some(...)`.

I went with `Option` rather than dummy values because the memo reuse check compares `memo_info.processor_fp == Some(processor_fp)`, because a live entry's `None` can never match it, so the two kinds of entries can't be mistaken for each other.

Durability stays user-declared (`durable_stream: bool`) since `LiveStream` exposes no cursor for the framework to detect it.


rust/core/src/engine/live_component.rs
```diff
@@ -1146,6 +1191,79 @@ async fn drain_task_body<Prof: EngineProfile>(
...
+    component
+        .app_ctx()
+        .app_store()
+        .write_user_state_standalone(
+            component.stable_path(),
+            db_schema::StateKind::Live,
+            &logic_deps_state_key(),
+            &encoded,
+        )
+        .await
```
georgeh0: I think user state may not be the right place for logic deps.

What we need to persist for change detection is pretty much the same as [`ComponentMemoizationInfo`](https://github.com/cocoindex-io/cocoindex/blob/4667e562979f6e8d0114c91b69bfbd5bf055f78c/rust/core/src/state/db_schema.rs#L238-L254). And the intention is also similar to its existing usage: they keep necessary information to validate if a regular component's last execution is still valid, and our purpose is actually the same.

So I think probably we can just reuse `ComponentMemoizationInfo` and reuse the same entry in the DB for it.

User states for live components is intended to store states related to specific live-component logic. And live-component logic will decide if persisting the `ComponentMemoizationInfo` and if the last persisted `ComponentMemoizationInfo` is currently valid. But these don't belong to user states.

pyjuan91: Agreed, this is framework change-detection state, not user state. I'll move it onto the component's memoization entry and drop the `sys/live_logic_deps` user-state key.

Two things I'd confirm before wiring it up:

(1) Lifecycle. The live set accumulates rather than being a per-run snapshot,  `update_full` replaces it, incremental `update` unions into it (an item processed earlier still exists on restart, so its subtree fingerprints must stay tracked). Validity check is the same `all_contained_with_env`, and the live controller owns the read-merge-write, which matches your "live logic decides when to persist / whether the last one is valid."

(2) Shape. Live only ever needs `logic_deps`. `ComponentMemoizationInfo`'s `processor_fp` and `return_value` are required (no skip/default, and `MemoizedValue` has no empty variant), so reusing it as-is means storing dummy F/R. Do you prefer that, or factoring `logic_deps` (the shared part) into a leaner shape both paths reuse? Leaning toward the latter, but your call.

badmonster0: hey @pyjuan91, thanks for the comment, what's your thought and take on this?

pyjuan91: Hi @badmonster0, I'll propose to reuse the single `ComponentMemoizationInfo` entry as @georgeh0 suggested, and make `F`/`R` `Option` so a live root stores just `{L: [...]}`.

I'd floated factoring `logic_deps` into a shared struct earlier. But that doesn't really fit "reuse the same entry", a leaner value would need either a different type under the `ComponentMemoization` key or a separate key. And it wouldn't remove the actual blocker anyway: `F`/`R` are required and a live root has neither. So I'd keep the one struct and just relax those two fields.

`Option` over sentinel `F`/`R`: with `processor_fp: None` a live entry can never match a regular skip (which always passes `Some`), so it can't be misread as a real memo, and sentinels would need a fake `R`, since `MemoizedValue` has no empty variant.

Lifecycle as before: `update_full` replaces the set, incremental `update` unions in, controller owns read-merge-write.

Small change either way, happy to store sentinels instead if you'd rather.

georgeh0: Can you clarify what's "F/R"? I don't understand the term. It's never defined anywhere.

pyjuan91: Sorry, "F/R" was my shorthand, I meant the two serde(rename) codes on [ComponentMemoizationInfo](https://github.com/cocoindex-io/cocoindex/blob/9f030317cb9fbc587878847c9ea8e1d352b34093/rust/core/src/state/db_schema.rs#L240-L243)

badmonster0: Thanks Poe for the follow up. we find these summary are overly verbose with ai generated summaries. could you summarize proposal in your own words for us to easily communicate and follow through the changes? Try Surface the most important design aspect that we should discuss.

This also helps us to see how much you are digesting and learning from this PR, thanks!

pyjuan91: Thanks for the feedback! I've rewritten the PR description to make it easy to understand.

Issue #2021, opened 2026-05-25
[FEATURE] Support sliced execution targeted for certain components


georgeh0:
Sometimes we want to only run a slice of the application, e.g. for certain components. It's useful for debugging/inspection purpose, and the "CocoInsight" functionality will be based on that.

- When we run an app, a component selector can be provided, and components not fitting into the selector can be skipped if the result is not used.
- For sources, when using `mount_each()`, it should also be able to push down the selectors, so that for sources that support point read we can read certain keys only, which skips the entire scanning.

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

MrAnayDongre:
Hey @georgeh0, I’d like to take this one.
I’ll work on a concrete plan first and then turn it into a PR.

georgeh0:
Hi @MrAnayDongre, great! Thanks for taking this!

badmonster0:
hey @MrAnayDongre are you still working on this? if not i'll take it for internal assignment just want to make sure that you are not in the middle of it. thanks!

badmonster0:
hey @MrAnayDongre i'll recycle this for now, but in case you have any thoughts, please let me know anytime, thanks for taking a look!

gjj930923:
hey @badmonster0 I can take a look at this issue.

badmonster0:
@gjj930923 assigned, thanks for taking a looK!

PR #2195, opened 2026-06-22
feat: Support sliced execution targeted for certain components (#2021)

gjj930923:
Description

Adds an optional `component_selector` parameter to `App.update()` / `App.update_blocking()` that limits execution to only matching `mount_each` children. Non-matching components are skipped before spawning any Rust tasks — no LMDB reads, no inflight permits, no tokio overhead.

Selector patterns are `fnmatch`-style globs matched against a human-readable path string (e.g. `process_file/ox.jpeg`):

```python
app.update_blocking(component_selector=["process_file/ox.jpeg"])
```

Motivation

When debugging or inspecting a pipeline, re-running only specific components is much faster than a full reprocess. This is also foundational for the "CocoInsight" inspection tool (see #2021).

How?

The selector flows through a module-level variable (`_current_component_selector`) in `component_ctx.py`. Filtering happens at mount time in `mount_each()` for both static and live sources, and in `_MountEachLiveComponent` / `LiveMapSubscriber` for live-mode catch-up and incremental updates.

`use_mount` components and the root always run — they're foreground dependencies, not selectable.

Breaking change?

No. The new parameter defaults to None (run everything), fully backward-compatible.

Related issue

[#2021](https://github.com/cocoindex-io/cocoindex/issues/2021)


python/cocoindex/_internal/app.py
```diff
@@ -277,6 +293,7 @@ def update(
         full_reprocess: bool = False,
         live: bool = False,
         preview: bool = False,
+        component_selector: list[str] | None = None,
```
georgeh0: I think we may want to use `StablePath` as the selector type, because in general the path can be different types, not only string.

python/cocoindex/_internal/component_ctx.py
```diff
@@ -42,6 +42,14 @@ class ExceptionHandlerChain(NamedTuple):
 # ContextVar for the current ComponentContext
 _context_var: ContextVar[ComponentContext] = ContextVar("coco_component_context")
 
+# Active component selector, set by App.update() / App.update_blocking().
+# This is a plain module attribute (not a ContextVar) because the Python
+# function callbacks from Rust tokio tasks may not share the asyncio
+# ContextVar of the event-loop task that called App.update(). However,
+# they DO share the same thread (the asyncio event-loop thread), so a
+# plain module-level variable works across this boundary.
+_current_component_selector: tuple[str, ...] | None = None
```
georgeh0: The selector is passed in as an argument to `update()`, but here it's a global variable. It won't work when multiple `update()` calls happen in parallel.

python/tests/test_component_selector.py
```diff
@@ -0,0 +1,282 @@
...
+
+    app = coco.App(
+        coco.AppConfig(name="test_selector_basic", environment=env),
+        app_main,
+    )
+
+    await app.update(component_selector=["process/a"])
+
+    assert set(GlobalDictTarget.store.data.keys()) == {"process/a"}
+    assert GlobalDictTarget.store.data["process/a"].data == "value_a"
```
georgeh0: I think we want to add one more test to verify: when the selector is provided, unselected paths are not touched.

e.g. we can test it by:
- run `app.update` without selector
- update input item
- run `app.update` with a selector

Under this sequence, we expect result values for unselected paths are not touched

Note: if we simply skip `mount()` call for a stable path, under the current engine logic, since the component for the skipped path is unseen, their output will be deleted (the effect is similar to the related entry from the source is gone), which is unexpected.

python/cocoindex/_internal/api.py
georgeh0, line 1: Besides `mount_each()`, we also want to be able to skip execution for unselected components for `mount()`.

Note that the original issue mentioned `mount_each()` specifically because there's one specific optimization can and should happen there: selector pushdown to source. e.g. if the source supports reading by key, it may expose an additional API to "read" instead of "list + filter" (e.g. if the source is local file system, directly reading a specific file is way more efficient than listing + filter).

With this said, it's also OK to implement the basic functionality without the optimization in the first PR, and leave the optimization to later.

gjj930923: I analyzed the codebase with AI and seems like the change requires to modify the Rust code, so I think we can leave the optimization later.

python/cocoindex/_internal/app.py
gjj930923: OK updated.

python/cocoindex/_internal/component_ctx.py
gjj930923: OK I removed _current_component_selector and add this selector into the ContextVar.

python/tests/test_component_selector.py
gjj930923: I added a new test case for this verification.

python/cocoindex/_internal/stable_path.py
```diff
@@ -55,5 +94,35 @@ def parts(self) -> list[StableKey]:
         """
         return self._core.parts()
 
+    def selector_path(self) -> str:
```
georgeh0: Is it used anywhere now?

python/tests/test_component_selector.py
```diff
@@ -0,0 +1,412 @@
...
+    # First run without selector — all items processed.
+    await app.update()
+    assert set(GlobalDictTarget.store.data.keys()) == {
+        "process/a",
+        "process/b",
+        "process/c",
+    }
+
+    # Change only the selected item's input.
+    items["a"] = "new_value_a"
```
georgeh0, line 78: Can we also change the source value for a non-selected key? Hence we can verify the selector really works.

Similar to other tests.

python/cocoindex/_internal/api.py
georgeh0: Sounds good to support `mount_each()` optimization later.

Note that I think Rust core needs to be updated any way even to support the basic functionality for `mount()` correctly. The current code change doesn't really handle that. I also left a comment in the tests to indicate the missing verifications there.

python/cocoindex/_internal/stable_path.py
gjj930923: Not used now. Let me remove it.

python/tests/test_component_selector.py
gjj930923: Do you mean if values of items 'a' and 'b' are changed to 'new_value_a' and 'new_value_b', but the following update() function only update the 'a' key, then the expected result of 'process/a' should be 'new_value_a' while 'process/b' should be 'value_b'? If true then I think something needs to be updated in the main coding files. Seems like this code change doesn't address this.

georgeh0: Yes, that's what I meant. The whole purpose of the feature request  (sliced execution) is to skip execution for unselected components. Please cover this with test and address this in the changed code. Thanks!

gjj930923: OK I have updated the code and all tests were also updated to verify unselected components. Please review, thanks.

georgeh0: The tests aren't updated to verify the property. e.g. in this test, only `items["a"]` is changed and it's selected in the next `update()` call. Source data for none of unselected component is updated.  Similar for other tests.

I insist this because I really suspect the unselected component will result in a deletion as the current implementation (it simply skips the execution, for the engine the effect is exactly the same as "the entry is gone"). Please really cover that property in test and see related problems. Thanks!

gjj930923: Hi @georgeh0 can you take a look at the last test case in this file, which is "test_unselected_paths_preserved"? I think it has already covered the situation of unselected component. If updating this test case to verify the unselected component is a better practice then I will make the change in the test file.

georgeh0: Thanks for pointing it out! Yes, the `test_unselected_input_changed_not_processed` test covered this the case of "source for an unselected component is updated".

I overlooked that, and the change in `rust/py` can deliver right outcome for the "update" case. But would you please also add another test case to cover the case of "source for an unselected component is deleted"? We expect the output for the key is unchanged too.

I think the right way to implement the feature is to do the core logic in the `rust/core` crate, and any attempt to do it without touching the core seems unable to deliver the right semantics (and also make the feature not reusable by SDKs other than Python, e.g. we also have Rust SDK in alpha now).

gjj930923: Hi @georgeh0 I made a new commit recently to add a test 'test_unselected_item_deleted_output_preserved' at the end of the test file. in order to make it work, I also changed some rust files to make it work.

rust/core/src/engine/execution.rs
```diff
@@ -553,12 +553,30 @@ impl<Prof: EngineProfile> Committer<Prof> {
...
+            let pre_execute_check: Option<Box<dyn FnOnce() -> bool + Send>> =
+                if let Some(ref sel) = component_selector {
+                    if !is_path_selected(&stable_path, Some(sel)) {
+                        Some(Box::new(|| false))
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
```
georgeh0, line 579: Instead of "preserving tombstones", I think the more expected approach is to not to create the tombstone at the first place. The ultimate purpose of the "sliced execution" is to only touch the selected components and skip execution & avoid all side effects for everything else not selected.

Also note that current changes for mount (processing) and deletion scattered at two different layers: one in the `py` crate and the other in the `engine` crate. Ideally we should handle them more uniformly within `engine`.

rust/core/src/state/stable_path.rs
```diff
@@ -2,6 +2,196 @@ use crate::prelude::*;
...
+            continue;
+        }
+        if path_parts
+            .iter()
+            .zip(sel_parts.iter())
+            .all(|(p, s)| selector_part_matches(p, s))
+        {
+            return true;
+        }
+    }
```
georgeh0, line 191: I think "fnmatch" may not be the desirable way to represent selectors. "fnmatch" is usually used to match filenames (which are strings), but component paths can have more types (e.g. integers, booleans, even tuples).

For more context, the most common way to use this sliced execution feature will be: in the observability UI, execute for components shown on a single page (usually 10 to 50 items per page), and there's usually no generic way to write fnmatch patterns for 50 component paths in a page. Instead, we'll likely need to use 50 separate paths. Hence, a more ideal way here is to parse the selector into a tree structure, using a `HashMap` at each level, so at each level we can do a matching with O(1) (rather than evaluating N different selectors here).

badmonster0:
@gjj930923 thanks a lot for working on this issue! we'll close it for now since it seem to be inactive. if you are still working on it, let us know and we are happy to re-open this!

PR #2210, opened 2026-06-24
feat: add type_hint to use_state for typed state deserialization (#2198)

### LINKED ISSUE #2198: [FEATURE] allow users to specify type hints for component states
In #2015 we introduced component states, but since there's nowhere to specify a type hint, we're using the generic deserializer when loading value from the database.

We want to extend the API to allow an explicit type hint passed into `use_state()` call, so the value can be deserialized into a specific type (e.g. a specific dataclass).

---

lichuang:
Allow users to pass a `type_hint` to `coco.use_state()` so stored values are deserialized into a concrete type (dataclass, NamedTuple, msgspec.Struct, pydantic BaseModel) instead of the generic Any returned by the default decoder. This brings type-safe access to persistent component state, with IDE auto-complete and static type checking on `.value`.

Changes:
- `use_state()` gains an optional keyword-only `type_hint` parameter
- `StateHandle` holds a per-instance deserializer; `.value` delegates to it instead of the module-level `_DESERIALIZE_ANY` fallback
- New overload covers `use_state(key, initial_value, *, type_hint)` call pattern
- Deserialization errors include the state key via `source_label` for debugging
- 7 new tests: dataclass / NamedTuple / msgspec.Struct / pydantic round-trips, no-initial edge case, overload resolution, and type-mismatch error path

Closes #2198.

### DIFF
--- a/python/cocoindex/_internal/api.py
+++ b/python/cocoindex/_internal/api.py
@@ -93,6 +93,8 @@ from .serde import (
     unpickle_safe,
     serialize_by_pickle,
     make_deserialize_fn,
+    get_deserialize_fn,
+    DeserializeFn,
 )
 
 from .pending_marker import PendingS, ResolvedS, MaybePendingS
@@ -648,22 +650,24 @@ class StateHandle(Generic[_StateT]):
     assign to `.value` to persist a new value for the next run.
     """
 
-    __slots__ = ("_key", "_stored", "_core_processor_ctx")
+    __slots__ = ("_key", "_stored", "_deserializer", "_core_processor_ctx")
 
     def __init__(
         self,
         key: StableKey,
         stored: core.StoredValue,
+        deserializer: DeserializeFn,
         core_processor_ctx: core.ComponentProcessorContext,
     ) -> None:
         self._key = key
         self._stored = stored
+        self._deserializer = deserializer
         self._core_processor_ctx = core_processor_ctx
 
     @property
     def value(self) -> _StateT:
         # Lazy read: object-backed returns directly; bytes-backed deserializes once, cached.
-        return self._stored.get(_DESERIALIZE_ANY)  # type: ignore[no-any-return]
+        return self._stored.get(self._deserializer)  # type: ignore[no-any-return]
 
     @value.setter
     def value(self, new_value: _StateT) -> None:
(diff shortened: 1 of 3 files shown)


badmonster0:

python/cocoindex/_internal/api.py
```diff
@@ -736,7 +781,14 @@ def use_state(key: StableKey, initial_value: Any = None) -> StateHandle[Any]:
         # Rust client errors surface as ValueError; normalize to RuntimeError so
         # all use_state usage errors have a consistent type for callers.
         raise RuntimeError(str(e)) from None
-    return StateHandle(key, stored, ctx._core_processor_ctx)
+    if type_hint is not None:
+        deserializer = make_deserialize_fn(
+            type_hint,
+            source_label=f"use_state key {key!r}",
+        )
```
georgeh0: We may consider using this one (after renaming), which is reused rather than constructed each time on the fly: https://github.com/cocoindex-io/cocoindex/blob/4667e562979f6e8d0114c91b69bfbd5bf055f78c/python/cocoindex/_internal/serde.py#L461-L463

python/cocoindex/_internal/api.py
```diff
@@ -676,8 +679,27 @@ def value(self, new_value: _StateT) -> None:
...
+@overload
 def use_state(key: StableKey, initial_value: _StateT) -> StateHandle[_StateT]: ...
-def use_state(key: StableKey, initial_value: Any = None) -> StateHandle[Any]:
+@overload
+def use_state(
+    key: StableKey,
+    *,
+    type_hint: type[_StateT],
+    initial_value: _StateT | None = None,
+) -> StateHandle[_StateT]: ...
```
georgeh0, line 688: minor: I think without `type_hint` we may just return `StateHandle[Any]` (there's no guarantee for the returned type). Maybe 2 overload variants are sufficient?

```python
@overload
def use_state(key: StableKey, initial_value: Any = None) -> StateHandle[Any]: ...
@overload
def use_state(
    key: StableKey,
    initial_value: _StateT | None = None,
    *,
    type_hint: type[_StateT],
) -> StateHandle[_StateT]:
  ...
```

georgeh0 (approved):
Thanks!

PR #2211, opened 2026-06-24
Agent experience: fix skill/config drift, add MCP discoverability

badmonster0:
## What this is

A deep review of CocoIndex's **AI-coding-agent surface** and the accuracy/discoverability fixes it surfaced. Scope reviewed: `AGENTS.md`, the bundled `skills/cocoindex/` skill, the `llms.txt` docs surface, `examples/`, and the MCP / Context7 / OpenCode configs.

**Verdict:** the agent-experience setup here is unusually strong. `AGENTS.md` is accurate against the real v1 API, the `llms.txt` `.md`-twin mechanism is genuinely implemented, all examples are on the v1 API with zero dead pre-1.0 leakage, and the hooks/skills are honest thin adapters. No P0s. These are targeted accuracy and discoverability fixes, not a rescue.

## Changes

**Skill (`skills/cocoindex/`)**
- `chunk.content` → `chunk.text` (the real `Chunk` field) in the ID-generation snippets of `SKILL.md` and `api_reference.md`. The skill already used `chunk.text` everywhere else, so this was self-inconsistent and would `AttributeError` if copied.
- Corrected project setup: `cocoindex init` does **not** create a `.env`, and `COCOINDEX_DB` is an optional fallback, not "required" (the scaffold sets `builder.settings.db_path` in the lifespan). Kept the accurate "CLI auto-loads `.env`" note.
- Documented the **Neo4j** and **FalkorDB** knowledge-graph target connectors (the skill advertises KG use cases but never documented them). Signatures ground-truthed against the connector source.
- `VectorSchema(dtype=np.dtype(np.float32))` — the field expects a `np.dtype` instance.

**Configs**
- `opencode.json`: bump pin `cocoindex>=1.0.0a16` → `>=1.0.6` and drop `--prerelease=explicit` (`cocoindex-code` already requires `>=1.0.6`).
- **Added `.mcp.json`** so Claude Code / any `.mcp.json`-aware client also gets the `cocoindex-code` semantic-search server (was OpenCode-only — Claude Code clones never got the tool).
- `context7.json`: dropped `previousVersions: v0.3.39`, a pre-v1 tag that would re-surface the dead `FlowBuilder`/`sources` API the file's own rules forbid.

**`AGENTS.md`**
- Added an "Agent Tooling" section documenting the `cocoindex-code` MCP `search` server + `ccc` CLI (previously undiscoverable from the repo).
- Made the worked example copy-runnable: added the import header and `recursive=True` on `walk_dir`.

## Notes for the reviewer
- Every API symbol in the new connector docs and the skill edits was verified against the real source (`python/cocoindex/connectors/{neo4j,falkordb}/_target.py`, `resources/chunk.py`, `cli.py`, `setting.py`).
- An earlier draft also touched ~25 docs/example files via an over-eager `cocoindex update main` → `main.py` rewrite and added 6 example READMEs. Both were dropped: `user_app_loader.py` accepts both run forms by design (`examples.ts` keeps `RUN_MAIN`/`RUN_MAIN_PY` on purpose), and `main` already has canonical READMEs for those 6 examples. This PR is scoped to the 8 files above only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -160,6 +160,12 @@ Changes are applied atomically per component. If a source item is deleted (path 
 ### Example
 
 ```python
+import pathlib
+
+import cocoindex as coco
+from cocoindex.connectors import localfs
+from cocoindex.resources.file import FileLike, PatternFilePathMatcher
+
 @coco.fn(memo=True)
 async def process_file(file: FileLike, target: localfs.DirTarget) -> None:
     html = _markdown_it.render(await file.read_text())
@@ -171,7 +177,9 @@ async def app_main(sourcedir: pathlib.Path, outdir: pathlib.Path) -> None:
     target = await coco.use_mount(localfs.declare_dir_target, outdir)
 
     files = localfs.walk_dir(
-        sourcedir, path_matcher=PatternFilePathMatcher(included_patterns=["**/*.md"])
+        sourcedir,
+        recursive=True,
+        path_matcher=PatternFilePathMatcher(included_patterns=["**/*.md"]),
     )
     await coco.mount_each(process_file, files.items(), target)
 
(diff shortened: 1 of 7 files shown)


AGENTS.md
```diff
@@ -293,6 +301,23 @@ The current codebase is for CocoIndex v1, which is a fundamental redesign from C
...
+
+### Semantic code search over this repo (`cocoindex-code` MCP)
+
+This repo ships a Model Context Protocol server for AST-aware semantic search over its own
+codebase. It is registered for both OpenCode (`opencode.json`) and Claude Code / any
+`.mcp.json`-aware client (`.mcp.json`). Both launch it the same way:
+
+```bash
+uvx --with cocoindex>=1.0.6 cocoindex-code@latest
+```
```
georgeh0, line 314: Should we actually encourage using the CLI rather than MCP?

Also the latest recommendation is to use `uv tool install` for this:

https://github.com/cocoindex-io/cocoindex-code#install

PR #2212, opened 2026-06-24
fix(core): reconcile orphaned child-existence rows when Directory becomes Component

Sujit-1509:
Fixes a reconciliation bug in the core state store where transitioning a target state from a `Directory` to a `Component` leaves orphaned child-existence (`__cex`) rows in the database and fails to emit tombstones for descendants.

When a node changes from `Directory` → `Component`, the existing implementation updates the node type in the parent child-existence table but never queues the old directory subtree for recursive cleanup. As a result:

* Nested child-existence rows remain orphaned in the state store.
* Tombstones are not written for descendant components.
* Stale state can persist indefinitely after pipeline structure changes.

## Root Cause

In `reconcile_child_existence`, the `Ordering::Equal` branch correctly handles:

* Directory → Directory
* Component → Component

However, when a node transitions from `Directory` → `Component`, the old directory subtree is not queued for deletion.

The code updates the node type but skips the recursive cleanup path normally triggered by directory removal.

## Fix

When reconciling an existing `Directory` against a new `Component`, enqueue the path with:

```rust
Level {
    path: level.path.concat_part(curr_key.clone()),
    child_path_set: None,
}
```

This reuses the existing recursive deletion logic and ensures:

* Descendant `__cex` rows are removed.
* Tombstones are emitted for descendant components.
* Nested directories are recursively cleaned up.
* No duplicate traversal or duplicate tombstones are generated.

## Tests

Added a regression test:

```python
test_directory_to_component_transition()
```

The test:

1. Creates a directory target state containing descendants.
2. Updates the application so the same path becomes a leaf component.
3. Verifies descendant state is removed.
4. Verifies tombstones are generated.
5. Verifies the new component remains intact.

## Why This Is Safe

* Uses the existing recursive deletion mechanism.
* No schema changes.
* No API changes.
* No behavioral changes outside the affected reconciliation path.
* Deletion and tombstone writes remain idempotent.

## Reproduction

Before:

```text
Directory D
└── child C
```

After pipeline update:

```text
Component D
```

Expected:

* C removed
* tombstone for C written

Actual:

* C remains in child-existence state store
* no tombstone written

This PR makes the reconciliation behavior consistent with the rest of the state cleanup system.

### DIFF
--- a/python/tests/core/test_component_target_states.py
+++ b/python/tests/core/test_component_target_states.py
@@ -1187,3 +1187,60 @@ def test_preview_rejects_child_target_providers() -> None:
     _source_data["D1"] = {"a": 1}
     with pytest.raises(Exception, match="child target providers"):
         app.update_blocking(preview=True)
+
+
+##################################################################################
+# Test: Directory -> Component transition child existence reconciliation
+##################################################################################
+
+_transition_to_component_mode = False
+
+
+@coco.fn
+async def _dummy_leaf_component() -> None:
+    pass
+
+
+async def _declare_transition_to_component() -> None:
+    with coco.component_subpath("transition_test"):
+        if not _transition_to_component_mode:
+            single_dict_provider = await coco.mount_target(DictsTarget.dict_target("D1"))
+            for key, value in _mount_target_source_data.get("D1", {}).items():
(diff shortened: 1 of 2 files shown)


georgeh0:
Thanks a lot for identifying the bug and fixing this!

PR #2221, opened 2026-06-25
fix(core): batch leaf target actions

### LINKED ISSUE #2219: [FEATURE]: Support batching for target state sinks from different components
## Describe the bug
Target row actions are only batched within a single processing component submit, not across sibling components created by `mount_each`.

In a `mount_each` pipeline where many child components write rows to the same target sink, CocoIndex invokes the row sink once per child component, often with actions containing a single row. This defeats the existing batch-shaped target sink API and causes backends such as LanceDB to perform many singleton row commits instead of one or a few batched writes.

This is especially visible with LanceDB: a 50-row ingest can result in roughly 50 `merge_insert` commits/table versions, even though the LanceDB connector can already build one Arrow batch and execute one batched `merge_insert` when it receives multiple actions.

### Steps to reproduce the behavior
1. Create a CocoIndex app that uses `mount_each` over many source items.
2. In each child component, declare one row into the same LanceDB table target.
3. Run a fresh ingest or global rewrite with many changed rows.
4. Instrument `cocoindex.connectors.lancedb._target._RowHandler._apply_actions`. Then, `print len(actions)` and observe that the sink is called many times with `len(actions) == 1`.
6. Inspect LanceDB table metadata/version history and observe one row-write commit per changed row, plus any optimize-related versions.
```bash
Observed example for 50 rows:
BATCH_PROBE actions=1
BATCH_PROBE actions=1
...
```
Final LanceDB metadata showed table versions roughly proportional to row count.

## Expected behavior
CocoIndex should coalesce target actions by sink across concurrently completing sibling components, within bounded batch limits.

For a `mount_each` workload where many child components write rows to the same LanceDB table, the LanceDB row sink should receive larger action batches, for example tens or hundreds of row actions per call, rather than one action per child component.

Tracking records should still only advance after the corresponding sink batch succeeds, and child-provider ordering should remain correct for sinks that return child handlers.

## How to reproduce

```py
# repro_lancedb_mount_each_batching.py
from __future__ import annotations

import pathlib
import tempfile
from typing import AsyncIterator

import cocoindex as coco
from cocoindex.connectors import lancedb
from cocoindex.connectors.lancedb import ColumnDef, TableSchema

import pyarrow as pa


DB = coco.ContextKey[lancedb.LanceAsyncConnection]("lancedb_conn")


@dataclass
class Row:
    id: int
    text: str


schema = TableSchema[Row](
    columns={
        "id": ColumnDef(pa.int64(), nullable=False),
        "text": ColumnDef(pa.string(), nullable=False),
    },
    primary_key=["id"],
    row_type=Row,
)


@coco.lifespan
async def lifespan(builder: coco.EnvironmentBuilder) -> AsyncIterator[None]:
    with tempfile.TemporaryDirectory() as tmpdir:
        conn = await lancedb.connect_async(str(pathlib.Path(tmpdir) / "lance"))
        builder.provide(DB, conn)
        yield


@coco.fn
async def write_one(i: int, table: lancedb.TableTarget[Row]) -> None:
    table.declare_row(row=Row(id=i, text=f"row {i}"))


@coco.fn
async def app_main() -> None:
    table = await lancedb.mount_table_target(
        DB,
        "rows",
        schema,
        # Avoid optimize noise in the repro; the batching problem exists either way.
        num_transactions_before_optimize=10_000,
    )

    await coco.mount_each(write_one, [(i, i) for i in range(50)], table)


def install_probe() -> None:
    orig = lancedb._target._RowHandler._apply_actions  # type: ignore[attr-defined]

    async def patched(self, context_provider, actions):
        print(f"BATCH_PROBE actions={len(actions)}")
        return await orig(self, context_provider, actions)

    lancedb._target._RowHandler._apply_actions = patched  # type: ignore[attr-defined]


if __name__ == "__main__":
    install_probe()
    app = coco.App(
        coco.AppConfig(name="repro_lancedb_mount_each_batching", environment=lifespan),
        app_main,
    )
    app.update_blocking()
```

```bash
uv run repro_lancedb_mount_each_batching.py
```
```
Expected current output shape:
BATCH_PROBE actions=1
BATCH_PROBE actions=1
...
```

There should be about 50 singleton calls. This demonstrates the issue: `mount_each` creates one child component per item, and target actions are applied per component submit instead of coalesced across sibling components writing to the same LanceDB row sink.

## CocoIndex Version
Current main branch / CocoIndex v1 source as of June 25, 2026.

## Additional context
This seems like it is not strictly LanceDB-specific. LanceDB exposes the underlying problem more clearly because singleton `merge_insert` calls create many table commits/versions and can conflict with background `table.optimize()` work. Other connectors such as Qdrant, Turbopuffer, Postgres, and Doris also have batch-capable sink implementations, so they can lose throughput when core passes singleton actions from many `mount_each` child components.

The issue appears to be a core runtime batching limitation: batching is scoped to one component submit, while `mount_each` intentionally creates many independent child components. `mount_each` is an idiomatic CocoIndex pattern, so users should not have to collapse work into one component to get efficient target writes.

prrao87:
## Summary

This fixes target sink batching for row-level writes emitted by sibling processing components, especially the common `mount_each(..., table_target)` shape reported in #2219.

Before this change, each component submitted its own leaf target actions directly to the sink after precommit. That meant connectors such as LanceDB could receive many separate `len(actions) == 1` calls even though their sink API already accepts a batch. The fix adds a small app-scoped core batcher keyed by target action sink, so concurrent sibling component commits that target the same sink are coalesced before the connector apply call.

Container targets still use the direct apply path because they return per-action child handlers. The batched path is only used when there are no child providers to fulfill; it still tolerates the orphan-delete case where a container sink returns an all-`None` handler list.

For LanceDB specifically, this also exposes a Python-level `write_batch_size` option on `table_target`, `declare_table_target`, and `mount_table_target`. The core fix determines whether the connector receives meaningful batches at all; the LanceDB option lets users split those coalesced batches into smaller LanceDB write calls when they need to tune memory, transaction size, or backend behavior.

Fixes #2219.

## Validation

I locally tested the core batching path and LanceDB connector behavior, including the new regression coverage for `mount_each` batching. I also checked the affected connector paths against Qdrant, Turbopuffer, Doris, SQLite, and the broader Rust/Python core suites to verify this did not regress other targets.

### DIFF
--- a/python/cocoindex/_internal/target_state.py
+++ b/python/cocoindex/_internal/target_state.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import (
     Collection,
+    Callable,
     Generic,
     Literal,
     NamedTuple,
@@ -11,7 +12,6 @@ from typing import (
     TypeAlias,
     overload,
 )
-import threading
 import weakref
 from typing_extensions import TypeVar
 
@@ -141,7 +141,7 @@ class AsyncTargetActionSinkFn(Protocol[ActionT_contra, OptChildHandlerT_co]):
 
 
 class TargetActionSink(Generic[ActionT_contra, OptChildHandlerT_co]):
-    __slots__ = ("_core",)
+    __slots__ = ("_core", "__weakref__")
     _core: core.TargetActionSink
 
     def __init__(self, core_action_sink: core.TargetActionSink):
@@ -151,38 +151,82 @@ class TargetActionSink(Generic[ActionT_contra, OptChildHandlerT_co]):
     def from_fn(
         fn: TargetActionSinkFn[ActionT_contra, OptChildHandlerT_co],
     ) -> "TargetActionSink[ActionT_contra, OptChildHandlerT_co]":
-        canonical = _SYNC_FN_DEDUPER.get_canonical(fn)
-        return TargetActionSink(core.TargetActionSink.new_sync(canonical))
+        return _get_cached_sink(fn, "sync", core.TargetActionSink.new_sync)
 
     @staticmethod
     def from_async_fn(
         fn: AsyncTargetActionSinkFn[ActionT_contra, OptChildHandlerT_co],
     ) -> "TargetActionSink[ActionT_contra, OptChildHandlerT_co]":
-        canonical = _ASYNC_FN_DEDUPER.get_canonical(fn)
-        return TargetActionSink(core.TargetActionSink.new_async(canonical))
+        return _get_cached_sink(fn, "async", core.TargetActionSink.new_async)
 
 
-class _ObjectDeduper:
-    __slots__ = ("_lock", "_map")
-    _lock: threading.Lock
-    _map: weakref.WeakValueDictionary[Any, Any]
-
-    def __init__(self) -> None:
-        self._lock = threading.Lock()
-        self._map = weakref.WeakValueDictionary()
-
-    def get_canonical(self, obj: Any) -> Any:
-        with self._lock:
-            value = self._map.get(obj)
-            if value is not None:
-                return value
-
-            self._map[obj] = obj
-            return obj
+_SINK_CACHE_ATTR = "_cocoindex_target_action_sinks"
 
 
-_SYNC_FN_DEDUPER = _ObjectDeduper()
-_ASYNC_FN_DEDUPER = _ObjectDeduper()
+def _get_cached_sink(
+    fn: Any,
+    mode: Literal["sync", "async"],
+    new_core: Callable[[Any], core.TargetActionSink],
+) -> TargetActionSink[Any, Any]:
+    """Reuse one core sink per callable owner so the Rust-side batcher is shared."""
+    owner = getattr(fn, "__self__", None)
+    func = getattr(fn, "__func__", None)
+    cache_key = (mode, func) if owner is not None and func is not None else (mode, None)
+    if owner is None:
+        owner = fn
+        func = None
+
+    try:
+        owner_ref = weakref.ref(owner)
+    except TypeError:
+        return TargetActionSink(new_core(fn))
+
+    cache = getattr(owner, _SINK_CACHE_ATTR, None)
+    if cache is None:
+        cache = {}
+        try:
+            setattr(owner, _SINK_CACHE_ATTR, cache)
+        except (AttributeError, TypeError):
+            return TargetActionSink(new_core(fn))
+
+    value = cache.get(cache_key)
+    if value is None:
+        callback = _make_weak_sink_callback(owner_ref, func, mode)
+        value = TargetActionSink(new_core(callback))
+        cache[cache_key] = value
+    return value
+
+
+def _make_weak_sink_callback(
+    owner_ref: weakref.ReferenceType[Any],
+    func: Any,
+    mode: Literal["sync", "async"],
+) -> Any:
+    if mode == "sync":
+
+        def sync_callback(
+            context_provider: ContextProvider, actions: Sequence[Any]
+        ) -> Any:
+            owner = owner_ref()
+            if owner is None:
+                raise RuntimeError("target action sink owner has been released")
+            if func is None:
+                return owner(context_provider, actions)
+            return func(owner, context_provider, actions)
+
+        return sync_callback
+
+    async def async_callback(
+        context_provider: ContextProvider, actions: Sequence[Any]
+    ) -> Any:
+        owner = owner_ref()
+        if owner is None:
+            raise RuntimeError("target action sink owner has been released")
+        if func is None:
+            return await owner(context_provider, actions)
+        return await func(owner, context_provider, actions)
+
+    return async_callback
 
 
 class TargetReconcileOutput(
--- a/python/cocoindex/_internal/target_state.py
+++ b/python/cocoindex/_internal/target_state.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import (
     Collection,
-    Callable,
     Generic,
     Literal,
     NamedTuple,
@@ -12,6 +11,7 @@ from typing import (
     TypeAlias,
     overload,
 )
+import threading
 import weakref
 from typing_extensions import TypeVar
 
@@ -151,82 +151,38 @@ class TargetActionSink(Generic[ActionT_contra, OptChildHandlerT_co]):
     def from_fn(
         fn: TargetActionSinkFn[ActionT_contra, OptChildHandlerT_co],
     ) -> "TargetActionSink[ActionT_contra, OptChildHandlerT_co]":
-        return _get_cached_sink(fn, "sync", core.TargetActionSink.new_sync)
+        canonical = _SYNC_FN_DEDUPER.get_canonical(fn)
+        return TargetActionSink(core.TargetActionSink.new_sync(canonical))
 
     @staticmethod
     def from_async_fn(
         fn: AsyncTargetActionSinkFn[ActionT_contra, OptChildHandlerT_co],
     ) -> "TargetActionSink[ActionT_contra, OptChildHandlerT_co]":
-        return _get_cached_sink(fn, "async", core.TargetActionSink.new_async)
+        canonical = _ASYNC_FN_DEDUPER.get_canonical(fn)
+        return TargetActionSink(core.TargetActionSink.new_async(canonical))
 
 
-_SINK_CACHE_ATTR = "_cocoindex_target_action_sinks"
+class _ObjectDeduper:
+    __slots__ = ("_lock", "_map")
+    _lock: threading.Lock
+    _map: weakref.WeakValueDictionary[Any, Any]
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._map = weakref.WeakValueDictionary()
+
+    def get_canonical(self, obj: Any) -> Any:
+        with self._lock:
+            value = self._map.get(obj)
+            if value is not None:
+                return value
+
+            self._map[obj] = obj
+            return obj
 
 
-def _get_cached_sink(
-    fn: Any,
-    mode: Literal["sync", "async"],
-    new_core: Callable[[Any], core.TargetActionSink],
-) -> TargetActionSink[Any, Any]:
-    """Reuse one core sink per callable owner so the Rust-side batcher is shared."""
-    owner = getattr(fn, "__self__", None)
-    func = getattr(fn, "__func__", None)
-    cache_key = (mode, func) if owner is not None and func is not None else (mode, None)
-    if owner is None:
-        owner = fn
-        func = None
-
-    try:
-        owner_ref = weakref.ref(owner)
-    except TypeError:
-        return TargetActionSink(new_core(fn))
-
-    cache = getattr(owner, _SINK_CACHE_ATTR, None)
-    if cache is None:
-        cache = {}
-        try:
-            setattr(owner, _SINK_CACHE_ATTR, cache)
-        except (AttributeError, TypeError):
-            return TargetActionSink(new_core(fn))
-
-    value = cache.get(cache_key)
-    if value is None:
-        callback = _make_weak_sink_callback(owner_ref, func, mode)
-        value = TargetActionSink(new_core(callback))
-        cache[cache_key] = value
-    return value
-
-
-def _make_weak_sink_callback(
-    owner_ref: weakref.ReferenceType[Any],
-    func: Any,
-    mode: Literal["sync", "async"],
-) -> Any:
-    if mode == "sync":
-
-        def sync_callback(
-            context_provider: ContextProvider, actions: Sequence[Any]
-        ) -> Any:
-            owner = owner_ref()
-            if owner is None:
-                raise RuntimeError("target action sink owner has been released")
-            if func is None:
-                return owner(context_provider, actions)
-            return func(owner, context_provider, actions)
-
-        return sync_callback
-
-    async def async_callback(
-        context_provider: ContextProvider, actions: Sequence[Any]
-    ) -> Any:
-        owner = owner_ref()
-        if owner is None:
-            raise RuntimeError("target action sink owner has been released")
-        if func is None:
-            return await owner(context_provider, actions)
-        return await func(owner, context_provider, actions)
-
-    return async_callback
+_SYNC_FN_DEDUPER = _ObjectDeduper()
+_ASYNC_FN_DEDUPER = _ObjectDeduper()
 
 
 class TargetReconcileOutput(
--- a/python/cocoindex/_internal/target_state.py
+++ b/python/cocoindex/_internal/target_state.py
@@ -141,7 +141,7 @@ class AsyncTargetActionSinkFn(Protocol[ActionT_contra, OptChildHandlerT_co]):
 
 
 class TargetActionSink(Generic[ActionT_contra, OptChildHandlerT_co]):
-    __slots__ = ("_core", "__weakref__")
+    __slots__ = ("_core",)
     _core: core.TargetActionSink
 
     def __init__(self, core_action_sink: core.TargetActionSink):
--- a/rust/core/src/engine/context.rs
+++ b/rust/core/src/engine/context.rs
@@ -329,12 +329,25 @@ impl<Prof: EngineProfile> TargetActionBatchState<Prof> {
             return;
         }
 
+        let actions_len = actions.len();
         let result = sink
             .apply(host_runtime_ctx, host_ctx, actions)
             .await
             .and_then(|handlers| {
-                if handlers.is_some() {
-                    client_bail!("leaf target action sink unexpectedly returned child handlers");
+                if let Some(handlers) = handlers {
+                    if handlers.len() != actions_len {
+                        client_bail!(
+                            "expect child providers returned by Sink to be the same length as the actions ({}), got {}",
+                            actions_len,
+                            handlers.len(),
+                        );
+                    }
+                    // Orphan deletes for container targets have no child
+                    // provider to fulfill, but their sink still returns an
+                    // all-None handler list.
+                    if handlers.into_iter().any(|handler| handler.is_some()) {
+                        client_bail!("target action sink returned child handlers without child providers");
+                    }
                 }
                 Ok(())
             })
--- a/rust/core/src/engine/context.rs
+++ b/rust/core/src/engine/context.rs
@@ -1,12 +1,16 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::mem;
 
+use cocoindex_utils::error::SharedResultExt;
 use cocoindex_utils::fingerprint::Fingerprint;
 
 use crate::engine::component::{Component, ComponentBgChildReadiness, StatsGroup};
 use crate::engine::id_sequencer::IdSequencerManager;
 use crate::engine::profile::EngineProfile;
 use crate::engine::stats::ProcessingStats;
-use crate::engine::target_state::{TargetStateProvider, TargetStateProviderRegistry};
+use crate::engine::target_state::{
+    TargetActionSink as _, TargetStateProvider, TargetStateProviderRegistry,
+};
 use crate::prelude::*;
 
 use crate::state::stable_path::StableKey;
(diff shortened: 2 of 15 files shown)


rust/core/src/engine/context.rs
```diff
@@ -166,6 +184,208 @@ impl<Prof: EngineProfile> AppContext<Prof> {
     }
 }
 
+/// App-level coalescer for leaf target actions that target the same sink.
+struct TargetActionBatcher<Prof: EngineProfile> {
```
georgeh0: CocoIndex already have a builtin batching util:

https://github.com/cocoindex-io/cocoindex/blob/main/rust/utils/src/batching.rs

Can we use this?

rust/core/src/engine/context.rs
```diff
@@ -166,6 +184,208 @@ impl<Prof: EngineProfile> AppContext<Prof> {
...
+    states:
+        Arc<tokio::sync::Mutex<HashMap<Prof::TargetActionSink, Arc<TargetActionBatchState<Prof>>>>>,
```
georgeh0: We need to have a clear way to GC orphaned sinks. We may consider using a `Weak` reference as key rather than `Arc`, and also GC the entry when the sink is GCed (to do this, may need RAII for existing `Prof::TargetActionSink`).

georgeh0 (commented):
Thanks for implementing the feature!

rust/core/src/engine/context.rs
prrao87: Ahh, missed that! Fixed. The custom queue code is now removed, and leaf target actions now use `cocoindex_utils::batching::{Batcher, BatchQueue, BatchingOptions, Runner}` in `context.rs`.

rust/core/src/engine/context.rs
prrao87: Makes sense. The app context now holds only `HashMap<usize, Weak<Batcher<…>>>` keyed by the sink's existing identity key; it keeps no strong reference to a sink or its closure. A batcher (which captures the sink) lives only while there are in-flight apply calls holding its `Arc`. When a wave drains, the Arc drops, the sink is released, and the Weak goes dead. The create path overwrites any dead entry for the same key, so the map doesn't accumulate. Hope that works!

prrao87: One trade-off to confirm: this fix recreates the batcher per concurrent wave rather than persisting it for the sink's whole lifetime. From my code analysis, coalescing only matters within a wave, so that's why I think this fix made sense.

prrao87:
Thanks @georgeh0 for the review! This fix addresses both comments and removes the LanceDB `write_batch_size` (upon second thought, it doesn't achieve anything useful in this PR):

1. **Reuse Batcher**: Good catch; the leaf-action coalescer is now a thin Runner over the existing `cocoindex_utils::batching::Batcher` instead of a custom queue/debounce. This means less code, is shared with the LMDB path, and no fixed debounce latency for single writes.
2. **Explicit sink GC**: app context holds `HashMap<usize, Weak<Batcher>>` keyed by sink identity; no strong references to closure sinks, batchers drop with the wave.
3. **Dropped the LanceDB `write_batch_size` option**. I thought long and hard about this, it doesn't warrant adding to the API surface. The core fix now already gives LanceDB batched writes (one `merge_insert` per coalesced batch), and the option only re-splits that into more commits — this is the opposite of what this PR needs 😅. It was also a no-op on the delete path. In the future, If we ever need to bound batch size, the right place is a connector-agnostic core `max_batch_size`, not a LanceDB kwarg — so I've left that as a documented follow-up and kept this PR to the core fix.

Container targets are unchanged (still direct apply). Core `max_batch_size` is intentionally deferred, because the current behavior is already unbounded per component, and the existing `Batcher` makes adding a cap a one-liner if a real need shows up.

rust/py/src/target_state.rs
```diff
@@ -64,6 +64,10 @@ fn get_core_field(py: Python<'_>, obj: Py<PyAny>) -> PyResult<Py<PyAny>> {
 
 #[async_trait]
 impl TargetActionSink<PyEngineProfile> for PyTargetActionSink {
+    fn batch_key(&self) -> usize {
+        self.key
+    }
+
```
georgeh0: For the method, `reclaimable_key` will be a better name than `batch_key` here.
The uniqueness of `self.key` is  only guaranteed when `self.callback` is not GCed.
`reclaimable_key` makes this clear, but `batch_key` doesn't (which can be misused).

rust/core/src/engine/context.rs
```diff
@@ -166,6 +184,107 @@ impl<Prof: EngineProfile> AppContext<Prof> {
     }
 }
 
+/// App-level coalescer for leaf target actions that target the same sink.
+struct TargetActionBatcher<Prof: EngineProfile> {
+    batchers: std::sync::Mutex<HashMap<usize, Weak<Batcher<TargetActionRunner<Prof>>>>>,
```
georgeh0: I think `Weak` should be around the underlying `TargetActionSink` rather than the batcher.

- Otherwise, whenever there's no active requests, the existing batcher will be dropped, and new batcher needs to be created. There'll be a lot of churn. And the map will grow forever.
- The map key here is actually reclaimable (see comments below).    In case the underlying sink is already GCed and the address is reused by another sink, but the batcher isn't dropped yet, we'll have trouble (maybe it cannot happen under current implementation, but it's fragile, since this relies on implementations of two crates).

Besides, note that we also need a way to recycle the map entry after the underlying sink is gone, otherwise the map will still grow forever sometimes.

---

In my mind, a potential solution could be something like:

- Engine side provides a `TargetActionSinkKeeper`, which owns a `EngineProfile::TargetActionSink`.
- `Arc<TargetActionSinkKeeper>` will be the major way to reference the `TargetActionSink`. i.e. the object we return to Python side should also be wrapped around it.
- `TargetActionBatcher` owns a `Weak<TargetActionSinkKeeper>`
- Implement `Drop` for  `TargetActionSinkKeeper` to remove the entry from the hash map when it's dropped.

prrao87: Great feedback, thanks! Will fix.

prrao87: @georgeh0 One cleanup detail I want to clarify before I rework it:

Do you want the batcher map entry removed immediately when the underlying sink is dropped, e.g. via RAII/`Drop`, or is it acceptable for dead entries to be recycled on the next batcher lookup by pruning `Weak` entries that can no longer be upgraded?

The second approach prevents unbounded growth in active workloads, but a dead entry can remain while the app is idle. The RAII approach gives immediate cleanup, but means the sink-lifetime object needs to know which app-level batcher registry entry to unregister from.

georgeh0: Yes, we want it to be cleaned up immediately.

As I said, there's no guarantee there's only limited number of sink fn within the same process. The second approach will potentially leak infinite map entries, which is not acceptable in my mind.

prrao87: @georgeh0 Reworked per your feedback! The sink lifetime now drives the cleanup:

- **Sink lifetime is explicit in core.** Added a `TargetActionSinkKeeper` that owns the `TargetActionSink`. `Arc<Keeper>` is the main handle (the Python `TargetActionSink` class and the Rust SDK `TargetActionSink<A>` both wrap it), so connector APIs are unchanged.
- **`Weak` is now around the sink, not the batcher.** The batcher map holds `Weak<Keeper>` + the batcher, so the batcher is reused across waves instead of being recreated each time (no churn).
- **Immediate RAII cleanup.** `Drop` for `TargetActionSinkKeeper` removes the map entry the moment the sink is gone, so an unbounded number of sink functions can't leak entries. Removal is identity-checked against the keeper allocation, so a reused reclaimable key can't evict a newer sink's entry.
- **Renamed `batch_key` → `reclaimable_key`** to make clear the key is only valid while the underlying callback is alive.

rust/core/src/engine/context.rs
```diff
@@ -1,12 +1,16 @@
...
+use cocoindex_utils::batching::{BatchQueue, Batcher, BatchingOptions, Runner};
 use cocoindex_utils::fingerprint::Fingerprint;
 
 use crate::engine::component::{Component, ComponentBgChildReadiness, StatsGroup};
 use crate::engine::id_sequencer::IdSequencerManager;
 use crate::engine::profile::EngineProfile;
 use crate::engine::stats::ProcessingStats;
-use crate::engine::target_state::{TargetStateProvider, TargetStateProviderRegistry};
+use crate::engine::target_state::{
+    TargetActionSink as _, TargetActionSinkKeeper, TargetStateProvider, TargetStateProviderRegistry,
```
georgeh0: What is this for: `TargetActionSink as _`

rust/core/src/engine/context.rs
```diff
@@ -166,6 +184,149 @@ impl<Prof: EngineProfile> AppContext<Prof> {
...
+                    TargetActionBatcherEntry {
+                        keeper_id,
+                        sink: Arc::downgrade(&sink),
+                        batcher: Arc::clone(&batcher),
+                    },
+                );
+
+                let weak_inner = Arc::downgrade(&self.inner);
+                sink.register_cleanup(move || {
+                    if let Some(inner) = weak_inner.upgrade() {
```
georgeh0: We don't need an upgrade here. We can directly remove `key` from the map, which is simpler.

The `keeper_id` field also seems unnecessary.

rust/core/src/engine/target_state.rs
```diff
@@ -29,6 +35,62 @@ pub trait TargetActionSink<Prof: EngineProfile>: Send + Sync + Eq + Hash + 'stat
     ) -> Result<Option<Vec<Option<ChildTargetDef<Prof>>>>>;
 }
 
+type TargetActionSinkCleanup = Box<dyn Fn() + Send + Sync + 'static>;
+
+/// Owns a target action sink and any cleanup hooks tied to its lifetime.
+pub struct TargetActionSinkKeeper<Prof: EngineProfile> {
+    sink: Prof::TargetActionSink,
+    reclaimable_key: usize,
+    cleanups: parking_lot::Mutex<Vec<TargetActionSinkCleanup>>,
```
georgeh0: I think can directly use a `OnceLock<TargetActionBatcher>`  here for simplicity. There's exactly one thing to cleanup, so don't need a `Vec` protected by a `Mutex`, and also don't need the `Box<dyn Fn()>` thing. `OnceLock` is much lighter.

rust/core/src/engine/execution.rs
```diff
@@ -1606,10 +1606,14 @@ pub(crate) async fn submit<Prof: EngineProfile>(
     let sink_result: Result<()> = async {
         let host_runtime_ctx = comp_ctx.app_ctx().env().host_runtime_ctx();
         for (sink, input) in actions_by_sinks {
-            let handlers = sink
-                .apply(host_runtime_ctx, Arc::clone(comp_ctx.host_ctx()), input.actions)
-                .await?;
             if let Some(child_providers) = input.child_providers {
+                // Container targets must return child handlers in the same
+                // order as their actions, so they cannot be coalesced with
+                // independent component submits.
```
georgeh0: Wondering where does this limitation come from?

Batching preserves ordering (elements of output vector is corresponding to each element of the input vector) otherwise a lot more will be broken.

prrao87: Good point. The container targets can be batched too, so the batcher runner is updated to return per-input child handler vectors. It's split from the flattened sink response using the original action counts, and routed container target applies through the same batcher path. Does that make sense?

rust/core/src/engine/context.rs
prrao87: Agreed, simplified this. The cleanup now directly removes the map entry by `reclaimable_key` when the keeper drops, and `keeper_id` is removed.

rust/core/src/engine/context.rs
prrao87: Removed this. It was only bringing the trait into scope for method-call syntax, but it was unclear (uses a clearer direct path).

rust/core/src/engine/target_state.rs
prrao87: Makes sense! This is now a `OnceLock<TargetActionBatcher<_>>` instead of a mutex-protected vector of boxed callbacks.

rust/core/src/engine/context.rs
```diff
@@ -166,6 +184,149 @@ impl<Prof: EngineProfile> AppContext<Prof> {
     }
 }
 
+/// App-level coalescer for leaf target actions that target the same sink.
+pub(crate) struct TargetActionBatcher<Prof: EngineProfile> {
+    inner: Arc<TargetActionBatcherInner<Prof>>,
+}
+
+struct TargetActionBatcherInner<Prof: EngineProfile> {
+    batchers: std::sync::Mutex<HashMap<usize, TargetActionBatcherEntry<Prof>>>,
```
georgeh0: Actually things can be even further simplified, a lot.

Since we already have a `TargetActionSinkKeeper`, we can make the `TargetActionSinkKeeper` own a `Batcher<TargetActionRunner<Prof>>`: they're strictly 1-1 mapping, so this is the cleanest way. We won't need another map, and won't need to play with `Weak` or control `drop`.

BTW, since we no longer need a `Weak<TargetActionSinkKeeper>`, define it as a `TargetActionSinkKeeperInner` and hide it within `TargetActionSinkKeeper`'s implementation. We'll pass `TargetActionSinkKeeper` around everywhere (which is cloneable ).

prrao87: I love it! Yes, this is a lot simpler. Removed a lot of unnecessary code as a result, and tests pass.

Just one added change on the Python side because we removed the sink identity key/registry: `target_state.py` now caches `TargetActionSink.from_fn(self._sink)` wrappers on the callable owner. Seems like there's no way around that change otherwise the python tests didn't pass.

prrao87:
There seems to be a flaky github runner issue for that type check fail, but tests are passing on my end.

rust/core/src/engine/context.rs
```diff
@@ -134,6 +136,16 @@ impl<Prof: EngineProfile> AppContext<Prof> {
...
+    pub async fn apply_target_actions(
+        &self,
+        host_ctx: Arc<Prof::HostCtx>,
+        sink: TargetActionSinkKeeper<Prof>,
+        actions: Vec<Prof::TargetAction>,
+    ) -> Result<Option<Vec<Option<ChildTargetDef<Prof>>>>> {
+        sink.apply(self.env().host_runtime_ctx(), host_ctx, actions)
+            .await
+    }
+
```
georgeh0: This layer has no value now. We can directly inline the call for `sink.apply` in the caller.

rust/core/src/engine/target_state.rs
```diff
@@ -29,6 +33,144 @@ pub trait TargetActionSink<Prof: EngineProfile>: Send + Sync + Eq + Hash + 'stat
...
+    batcher: Batcher<TargetActionRunner<Prof>>,
+}
+
+impl<Prof: EngineProfile> Clone for TargetActionSinkKeeper<Prof> {
+    fn clone(&self) -> Self {
+        Self {
+            inner: Arc::clone(&self.inner),
+        }
+    }
+}
```
georgeh0: We don't need to manually implement this. This can be derived.

rust/core/src/engine/context.rs
prrao87: Fixed

rust/core/src/engine/target_state.rs
prrao87: Good catch, moved to a `#[derive(Clone)]`

python/cocoindex/_internal/target_state.py
```diff
@@ -151,38 +151,82 @@ def __init__(self, core_action_sink: core.TargetActionSink):
...
+        context_provider: ContextProvider, actions: Sequence[Any]
+    ) -> Any:
+        owner = owner_ref()
+        if owner is None:
+            raise RuntimeError("target action sink owner has been released")
+        if func is None:
+            return await owner(context_provider, actions)
+        return await func(owner, context_provider, actions)
+
+    return async_callback
```
georgeh0: I don't get why we need this change? it's intrusive and has more assumption on what the function is. Why isn't the original `_ObjectDeduper` sufficient?

Existing target implementations already only call `TargetActionSink.from_async_fn()` exactly once per instance (e.g. [lancedb code](https://github.com/cocoindex-io/cocoindex/blob/36eca13e0d84c128ae72a9874757c9e93736af0d/python/cocoindex/connectors/lancedb/_target.py#L378), calling it once per table).

(Actually an even better way is to have a shared sink per database, so changes across different tables can also be applied within the same transaction. But this should be achieved by changing the connector code about how to construct the sink. The change here isn't helpful for this)

rust/core/src/engine/target_state.rs
```diff
@@ -29,6 +33,137 @@ pub trait TargetActionSink<Prof: EngineProfile>: Send + Sync + Eq + Hash + 'stat
...
+    ) -> Result<impl ExactSizeIterator<Item = Self::Output>> {
+        let num_inputs = inputs.len();
+        let mut input_iter = inputs.into_iter();
+        let Some(first_input) = input_iter.next() else {
+            return Ok(Vec::new().into_iter());
+        };
+        let TargetActionRunnerInput {
+            keeper,
+            host_runtime_ctx,
+            host_ctx,
```
georgeh0: For all these 3 fields (`keeper`, `host_runtime_ctx` and `host_ctx`), only the 1st element is used, and others are discarded. This is suspicious.

- We don't need `keeper` as a field for the input. Instead, I think `sink` should be a member of `Runner`. Move that to `Runner` when constructing it.
- For `host_runtime_ctx` and `host_ctx`, ideally we want to create a batcher for each pair of `(host_runtime_ctx, host_ctx)` so they can be held by the Runner too. To unblock this PR, it also works if we group them here locally by `(host_runtime_ctx, host_ctx)` and call `apply()` once for each group. Either way, we need to provide `Eq` and `Hash` for them.

georgeh0 (approved):

python/cocoindex/_internal/target_state.py
prrao87: Thanks for your patience! The mistake was that Codex ended up optimizing for the existing construction pattern tests that already existed, which led it down the wrong path. The original `_ObjectDeduper` is sufficient, and the generic wrapper cache was unnecessary. I'll be careful to limit the blast radius of changes to the bare minimum (Python-side target changes should not be necessary).

I've been testing all these fixes after building the Python SDK on my real use case and it's working beautifully so far, so that's also giving me confidence that we're getting there, slowly but surely! 😄

rust/core/src/engine/target_state.rs
prrao87: Fixed. Removed `keeper` from each batch input (the runner now owns the sink directly) and added identity-based `Eq`/`Hash` where needed for context grouping, including `PyAsyncContext`.

python/cocoindex/_internal/target_state.py
prrao87: Fixed, restored the original `_ObjectDeduper` behavior, so the core Python target-state API is back to matching main. 🎉

prrao87: > Actually an even better way is to have a shared sink per database, so changes across different tables can also be applied within the same transaction. But this should be achieved by changing the connector code about how to construct the sink.

My understanding is that this would be a larger refactor that could make this design cleaner for a future PR? If this PR works for the interim, maybe we could create an issue for that refactor that would be done by you in a future PR?

rust/core/src/engine/target_state.rs
```diff
@@ -29,6 +33,169 @@ pub trait TargetActionSink<Prof: EngineProfile>: Send + Sync + Eq + Hash + 'stat
...
 }
 
+/// Cloneable handle to a target action sink and its per-sink batcher.
+#[derive(Clone)]
+pub struct TargetActionSinkKeeper<Prof: EngineProfile> {
+    inner: Arc<TargetActionSinkKeeperInner<Prof>>,
+}
+
+struct TargetActionSinkKeeperInner<Prof: EngineProfile> {
+    sink: Arc<Prof::TargetActionSink>,
```
georgeh0: We don't really need the `sink` field here. The only way to use it is through the `batcher`. We won't need a clone for it here.

(there're one existing call for `sink()` but it's for testing only and should be updated to call the `apply()` method directly which uses the batcher)

prrao87: Should be fixed now! I also checked for other such duplicates like it.

georgeh0 (approved):

PR #2224, opened 2026-06-26
feat: support multiple GPUs and fractional GPU allocations via GPUPool (#2217)

### LINKED ISSUE #2217: [FEATURE] Support multiple GPU and fractional GPU allocations
Currently we provides a [GPU runner](https://cocoindex.io/docs/programming_guide/function/#runner) for scheduling GPU workload. The semantic is: a function annotated with `runner=coco.GPU` will be considered as requiring 1 GPU exclusively during running. So we queue up all invocations for cocoindex functions running with `runner=coco.GPU`.

We want to extend the support for the following situations:
- Multiple GPUs. In this case, multiple functions with `runner=coco.GPU` can run in parallel, and they're aware of the GPU assigned to them.
- Fractional GPU allocations, e.g. a cocoindex function may only require 0.5 GPU.

---

lichuang:
Closes #2217.

## Summary

Extends the GPU runner to support **multiple GPUs** (parallel execution across cards) and **fractional GPU allocations** (e.g. `coco.GPU(0.5)` to share a card).

## Motivation

Previously, `runner=coco.GPU` meant "exclusive use of 1 GPU" — all GPU functions shared a singleton `GPURunner` with a single `BatchQueue` that serialized every call. Even on multi-GPU machines, only one GPU function could run at a time.

### New: `GPUPool` (`runner.py`)

Fractional capacity tracker. Each GPU starts at capacity `1.0`. `acquire(fraction)` blocks until a GPU with enough remaining capacity is available, then returns its id. Uses a most-free scheduling strategy to balance load across GPUs. `asyncio.Condition` is lazily created and auto-rebound on event-loop switches (fixes pytest-asyncio compatibility).

### Refactored: `GPURunner` (`runner.py`)

- **Removed singleton** (`__new__` + `_instance`) — `GPURunner` is now a regular class
- **Added `fraction` parameter** and `__call__` factory — `coco.GPU` is `GPURunner(fraction=1.0)` (backward compatible), `coco.GPU(0.5)` creates a new runner requesting half a GPU
- **Pool-based concurrency** — `run`/`run_sync_fn` now `acquire` from `GPUPool` before execution and `release` after, replacing the old `ThreadPoolExecutor(max_workers=1)` serialization
- **GPU thread pool preserved** (`thread_name_prefix="gpu"`) but without `max_workers=1` — the pool gates concurrency, not the thread pool

### New: `current_gpu()` (`runner.py`)

`ContextVar`-based GPU identity propagation. Set by `GPURunner` before calling the user function, readable via `coco.current_gpu()` inside the function body. For sync functions running in the thread pool, `_run_with_gpu_context` sets the `ContextVar` inside the executor thread.

### Fix: `_create_batcher` queue selection (`function.py`)

**Critical fix for multi-GPU parallelism.** Runner-only mode (`runner=coco.GPU` without batching) now uses a per-function `BatchQueue` instead of the runner's shared queue. The shared queue's `ongoing_count` serialized all GPU calls regardless of pool capacity. With per-function queues, `GPUPool` is the sole concurrency gate.

Batching mode (`batching=True, runner=coco.GPU`) still uses the shared queue for batch aggregation — multi-GPU for batching is a known limitation (documented in code comments).

### Config

- `COCOINDEX_NUM_GPUS=N` env var (default: 1) — sets the default pool size
- `coco.configure_gpu_pool(N)` — programmatic override
- `COCOINDEX_RUN_GPU_IN_SUBPROCESS=1` + multi-GPU emits a `UserWarning` (subprocess mode doesn't yet support per-GPU `CUDA_VISIBLE_DEVICES`)

### Tests (`test_gpu_pool.py`)

17 new tests across three layers:
- **GPUPool unit tests** (6): acquire/release, blocking, fractional sharing, multi-GPU parallel, validation
- **GPURunner direct-call tests** (6): sync/async ContextVar propagation, `__call__` factory, fraction validation, parallel/serial behavior
- **`@coco.fn` full-path tests** (5): multi-GPU parallel (sync + async), single-GPU serialization, fractional sharing, capacity limiting — all through the real `@coco.fn` → Batcher → BatchQueue → Runner → Pool path

## Backward Compatibility

With the default pool (`num_gpus=1`), `runner=coco.GPU` behaves identically to before — `pool.acquire(1.0)` is equivalent to the old mutex. All 32 existing GPU/batching tests pass unchanged.

## Known Limitations

- `batching=True, runner=coco.GPU` still serializes batches through the shared queue (batching vs parallelism trade-off)
- Subprocess mode (`COCOINDEX_RUN_GPU_IN_SUBPROCESS=1`) doesn't support multi-GPU (requires per-GPU `ProcessPoolExecutor` with `CUDA_VISIBLE_DEVICES` at spawn time)

### DIFF
--- a/python/cocoindex/_internal/runner.py
+++ b/python/cocoindex/_internal/runner.py
@@ -1,25 +1,28 @@
 """
 Runner base class and GPU runner implementation.
 
-Runners execute functions in specific contexts. Each runner owns a BatchQueue
-that serializes execution.
+Runners execute functions in specific contexts. Each runner owns a BatchQueue.
 
-The GPU runner runs in-process by default with an async lock for serialization.
-Set COCOINDEX_RUN_GPU_IN_SUBPROCESS=1 for subprocess isolation.
+The GPU runner supports multiple GPUs and fractional GPU allocations.
+Configure the GPU pool size via the COCOINDEX_NUM_GPUS environment variable
+(default: 1). Set COCOINDEX_RUN_GPU_IN_SUBPROCESS=1 for subprocess isolation.
 """
 
 from __future__ import annotations
 
 import asyncio
 import functools
+import os
 import pickle
+import threading
+import multiprocessing as mp
+import warnings
 from abc import ABC, abstractmethod
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from concurrent.futures.process import BrokenProcessPool
+from contextvars import ContextVar
 from typing import Any, Callable, Coroutine, TypeVar, ParamSpec
-import threading
-import os
-import multiprocessing as mp
+
 from . import core
 
 P = ParamSpec("P")
@@ -34,8 +37,9 @@ _in_subprocess: bool = False
 class Runner(ABC):
     """Base class for runners that execute functions.
 
-    Each runner owns a BatchQueue that serializes execution of all functions
-    using this runner. The queue is created lazily on first use.
+    Each runner owns a BatchQueue, created lazily on first use. The queue is
+    shared with functions using this runner for batch aggregation; concurrency
+    control is handled by the runner's run/run_sync_fn implementations.
 
     Subclasses must implement:
     - run(): Execute an async function
@@ -52,8 +56,7 @@ class Runner(ABC):
     def get_queue(self) -> core.BatchQueue:
         """Get or create the BatchQueue for this runner.
 
-        All functions using this runner share this queue, ensuring
-        serial execution of workloads.
+        All functions using this runner share this queue for batch aggregation.
         """
         if self._queue is None:
             with self._queue_lock:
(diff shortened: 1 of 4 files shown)


python/cocoindex/_internal/runner.py
```diff
@@ -204,34 +207,148 @@ def in_subprocess() -> bool:
...
+
+_default_gpu_pool: GPUPool | None = None
+_default_gpu_pool_lock = threading.Lock()
+
+
+def _get_default_gpu_pool() -> GPUPool:
+    global _default_gpu_pool
+    with _default_gpu_pool_lock:
+        if _default_gpu_pool is None:
+            num_gpus = int(os.environ.get("COCOINDEX_NUM_GPUS", "1"))
```
georgeh0: I'm thinking maybe there're better default value than `1`.

I searched a little bit. A reasonable way may be checking `CUDA_VISIBLE_DEVICES`, and then calling `nvidia-smi` (it's what the framework like Ray is doing). When both missing, use the default value `1`.

python/cocoindex/_internal/runner.py
```diff
@@ -204,34 +207,148 @@ def in_subprocess() -> bool:
...
+# ============================================================================
+# GPU identity propagation
+# ============================================================================
+
+_current_gpu: ContextVar[int | None] = ContextVar("coco_current_gpu", default=None)
+
+
+def current_gpu() -> int | None:
+    """Return the physical GPU id assigned to the current call, or None."""
+    return _current_gpu.get()
```
georgeh0: We may return a list. There can also be more than 1 GPUs allocated for some tasks (even if not supported in the same PR, will likely be supported later, and we don't want a breaking API change then).

Besides, apart from this API, we may need to provide another API that returns the GPU amount allocated for the current context.

i.e. to the end, we may support 2 scenarios:
- fractional GPU < 1 (e.g. 0.3, 0.5)
- GPU in integer amount >=1 (e.g. 1, 2, 3)

With both APIs (an API to return allocated GPU list, and an API to return allocated GPU amount), the user code will have all needed information.

python/cocoindex/_internal/runner.py
georgeh0, line 1: I'm thinking if it's good to put the most intensive logic (e.g. the GPU resource allocation) to Rust side? I think the logic will also grow further. Putting it on Rust side will enable maximal reuse (e.g. by the Rust SDK in the future).

georgeh0 (commented):
Thanks for the PR!

python/cocoindex/_internal/runner.py
lichuang: That's a good long-term direction. Moving GPU allocation to the Rust side would indeed make it reusable for a future Rust SDK and keep intensive logic in one place.

For this PR, I'd prefer to keep the implementation on the Python side because:
1. The scope is already focused on exposing multi-GPU / fractional allocation to Python users.
2. Moving it to Rust would require new PyO3 bindings and cross-language state synchronization, which is a larger refactor.
3. The current Python implementation is self-contained (GPUPool + GPURunner), so it can be migrated later without breaking the public API.

Happy to open a follow-up issue to track moving GPU resource allocation to Rust once the Rust SDK becomes a priority.

lichuang:
The failure in test_dicts_in_sub_components_delete_dict appears unrelated to this PR — it's a RuntimeError: This event loop is already running from environment.py under Python 3.14 freethreaded. Could a maintainer rerun the failed job? https://github.com/cocoindex-io/cocoindex/actions/runs/28347102597/job/83972592715?pr=2224

python/cocoindex/_internal/runner.py
georgeh0: Sounds good to address as a follow-up issue. Thanks!

georgeh0:
> The failure in test_dicts_in_sub_components_delete_dict appears unrelated to this PR — it's a RuntimeError: This event loop is already running from environment.py under Python 3.14 freethreaded. Could a maintainer rerun the failed job? https://github.com/cocoindex-io/cocoindex/actions/runs/28347102597/job/83972592715?pr=2224

It's a flakiness introduced recently and just fixed in https://github.com/cocoindex-io/cocoindex/commit/32834a4656821c81c968ed3e8ec9e61d370a7596. The PR is good to merge. Thanks.

lichuang:
Can this PR be merged?

georgeh0 (approved):
Thanks for the PR! Merging.

PR #2228, opened 2026-06-27
fix(postgres): preserve data and NOT NULL constraints during column type evolution

Sujit-1509:
## Summary

Fixes an issue in the PostgreSQL connector where evolving the type of a `NOT NULL` column could unnecessarily trigger the destructive fallback path.

The existing implementation generated invalid SQL when applying a type change together with a `NOT NULL` constraint. PostgreSQL rejected the statement, causing the connector to assume the type conversion had failed and fall back to dropping and recreating the column.

As a result:

* Compatible type changes could unnecessarily destroy existing column data.
* Recreated columns silently lost their `NOT NULL` constraint.

## Changes

* Generate valid PostgreSQL `ALTER TABLE` syntax by separating the type change from the nullability modification.
* Preserve the desired `NOT NULL` constraint when the fallback `DROP COLUMN` + `ADD COLUMN` path is genuinely required.
* Emit a warning before performing the destructive fallback so users are aware that existing column data will be lost.
* Extend the PostgreSQL connector integration tests to cover schema evolution scenarios and ensure data and constraints are preserved for compatible type changes.

## Root Cause

The connector attempted to execute SQL in the form:

```sql
ALTER TABLE ...
ALTER COLUMN "column" TYPE <type> NOT NULL
```

PostgreSQL requires these to be separate `ALTER COLUMN` actions within the same `ALTER TABLE` statement. The syntax error caused every compatible `NOT NULL` type migration to enter the fallback path intended only for incompatible casts.

## Testing

Added regression coverage for PostgreSQL schema evolution, including:

* compatible type changes
* `NOT NULL` → `NOT NULL`
* nullable → nullable
* nullable → `NOT NULL`
* `NOT NULL` → nullable
* destructive fallback preserving the target nullability

These tests ensure compatible schema evolution preserves both existing data and column constraints.

### DIFF
--- a/python/cocoindex/connectors/postgres/_target.py
+++ b/python/cocoindex/connectors/postgres/_target.py
@@ -1158,13 +1158,12 @@ class _TableHandler(coco.TargetHandler[_TableSpec, _TableTrackingRecord, _RowHan
                         f"Recreating column. Existing data will be lost."
                     )
 
-                    nullable = "" if desired_col.nullable else " NOT NULL"
                     await conn.execute(
                         f'ALTER TABLE {qualified_name} DROP COLUMN IF EXISTS "{col_name}"'
                     )
                     await conn.execute(
                         f"ALTER TABLE {qualified_name} "
-                        f'ADD COLUMN "{col_name}" {desired_col.type}{nullable}'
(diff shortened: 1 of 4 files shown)


georgeh0 (approved):
Thanks a lot for the fix!

PR #2232, opened 2026-06-28
fix(localfs): add periodic rescan + watcher recreation for live mode

### LINKED ISSUE #2227: [BUG] localfs live watcher silently stops receiving FSEvents after days on macOS
### Problem

When using `localfs.walk_dir(..., live=True)` on macOS, the file watcher silently stops detecting filesystem changes after running continuously for several days. The process remains alive and logs show "Watching for changes," but no new file events are processed. Restarting the process immediately resolves the issue.

### Environment

- **macOS**: Sequoia (Darwin 24.x, Apple Silicon)
- **Python**: 3.14
- **watchdog**: 6.0.0
- **cocoindex**: 1.0.11
- **Observer backend**: `FSEventsObserver` (default on macOS)

### Reproduction

We run four CocoIndex apps in `live` mode via `launchd` (KeepAlive, always-on). Three use `localfs.walk_dir(..., live=True, recursive=True)` to watch directories of markdown and source files. After ~4 days of continuous operation, the docs watcher stopped processing new or modified files entirely. The process was healthy (no crash, no error logs), but no new files were ingested since the watcher went stale. A service restart immediately resumed normal operation.

This is difficult to reproduce on demand because the underlying FSEvents failure is non-deterministic and timing-dependent.

### Root cause analysis

The `_LiveDirItems._watch()` method in `connectors/localfs/_source.py` enters an unbounded `while True` loop:

```python
while True:
    event = await events_queue.get()
    # ... process event ...
```

This blocks indefinitely waiting for the next event from watchdog's `FSEventsObserver`. If the macOS FSEvents stream silently stops delivering events, no event is ever enqueued, and the loop hangs forever with no recovery path.

FSEvents on macOS has several known failure modes where the stream silently stops:

- **`fseventsd` daemon enters a bad state** when too many stream clients are active or internal limits are hit ([parcel-bundler/watcher#190](https://github.com/parcel-bundler/watcher/issues/190), [Facebook Watchman troubleshooting docs](https://github.com/facebook/watchman/blob/main/website/docs/troubleshooting.md))
- **`FSEventsObserver` emitter thread hangs** after concurrent file operations ([gorakhargosh/watchdog#736](https://github.com/gorakhargosh/watchdog/issues/736))
- **FSEvents silently fails on newer macOS versions** due to SDK/ABI changes ([fsevents/fsevents#406](https://github.com/fsevents/fsevents/issues/406))
- **`CFRunLoop` stops processing** in long-running background threads (Apple Developer docs, various SO reports)

These are OS-level issues outside CocoIndex's control, but the watcher can defend against them.

### Proposed fix

Replace the unbounded `await events_queue.get()` with a timeout-based poll that triggers a periodic full rescan as a safety net:

```python
while True:
    try:
        event = await asyncio.wait_for(events_queue.get(), timeout=300.0)
    except asyncio.TimeoutError:
        # Periodic safety-net rescan: catches any events the OS
        # watcher silently dropped.
        await subscriber.update_all()
        continue
    # ... rest of event processing unchanged ...
```

This approach:

- **Reuses existing infrastructure**: `subscriber.update_all()` is already called during the initial scan and is idempotent — it only processes files that actually changed
- **Is backward-compatible**: No API changes, no new dependencies
- **Is the industry-standard pattern**: Facebook Watchman, VS Code, and chokidar all recommend overlaying periodic polling on native filesystem watchers for exactly this class of failure
- **Has minimal performance impact**: A 5-minute interval means at most one extra directory walk per 5 minutes, which is negligible for typical watched directories

The timeout interval could optionally be made configurable (e.g., a `rescan_interval` parameter on `walk_dir` or `DirWalker`).

### Additional context

We currently mitigate this with a daily `launchd` job that restarts the CocoIndex process, but that's a blunt workaround that loses in-flight state and doesn't protect against the watcher going stale mid-day.

Happy to submit a PR for this if the approach looks good. Also open to discussing alternative approaches (e.g., a separate health-check thread, configurable interval, etc.).

jordigilh:
## Summary

Implements the periodic rescan + watcher recreation approach discussed in #2227 with @georgeh0.

On macOS (and potentially other platforms), FSEvents can silently stop delivering events after days of continuous operation. The live file watcher had no recovery path — it blocked indefinitely on `events_queue.get()`.

This adds a configurable `rescan_interval` parameter to `walk_dir()` (default: 3600s / 1 hour) that periodically:

1. **Stops** the current watchdog Observer (releases the FSEvents stream)
2. **Drains** stale events from the queue
3. **Creates** a fresh Observer (new FSEvents/inotify stream from the OS)
4. **Runs** a full rescan via `subscriber.update_all()`

### Design decisions

- **Fixed-cadence rescan, decoupled from event flow** — as @georgeh0 suggested, the rescan fires on a timer regardless of whether events are being received. This catches both failure modes: watcher completely dead (no events) and watcher dropping individual events while the stream is alive.
- **Watcher recreation** — each cycle tears down the old Observer and creates a new one, getting a fresh OS-level watch stream. This is the definitive fix for `CFRunLoop`/`fseventsd` staleness.
- **`rescan_interval=None` disables** — for callers who don't want periodic rescanning (e.g. short-lived processes).
- **Backward compatible** — the default 1-hour interval is enabled automatically; existing callers get the safety net without code changes.

### Changes

- `python/cocoindex/connectors/localfs/_source.py` — core implementation
- `python/tests/connectors/test_localfs_source.py` — unit tests for `rescan_interval` parameter
- `python/tests/connectors/test_localfs_live.py` — e2e tests validating rescan + disable behavior

### Test plan

- [x] All existing unit tests pass (`test_localfs_source.py` — 37 tests)
- [x] All existing e2e live tests pass (`test_localfs_live.py` — 4 tests)
- [x] New `test_localfs_live_rescan_interval` — validates files are detected with a short rescan interval, including after a rescan cycle
- [x] New `test_localfs_live_rescan_none_disables` — validates `rescan_interval=None` disables rescanning while events still work
- [x] New `TestRescanInterval` unit tests — validates parameter defaults, custom values, None, and pass-through from `walk_dir()`
- [x] `ruff check` and `ruff format` pass

Closes #2227

Made with [Cursor](https://cursor.com)

### DIFF
--- a/python/cocoindex/connectors/localfs/_source.py
+++ b/python/cocoindex/connectors/localfs/_source.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import datetime as _datetime
 import os
 from datetime import datetime
 from pathlib import Path
@@ -66,7 +67,7 @@ class File(_file.FileLike[pathlib.Path]):
         return await asyncio.to_thread(self._read_sync, size)
 
 
-_DEFAULT_RESCAN_INTERVAL: float = 3600.0
+_DEFAULT_RESCAN_INTERVAL: _datetime.timedelta = _datetime.timedelta(hours=1)
 
 
 class DirWalker:
@@ -85,7 +86,7 @@ class DirWalker:
     _recursive: bool
     _path_matcher: FilePathMatcher
     _live: bool
-    _rescan_interval: float | None
+    _rescan_interval: _datetime.timedelta | None
 
     def __init__(
         self,
@@ -94,7 +95,7 @@ class DirWalker:
         live: bool = False,
         recursive: bool = False,
         path_matcher: FilePathMatcher | None = None,
-        rescan_interval: float | None = _DEFAULT_RESCAN_INTERVAL,
+        rescan_interval: _datetime.timedelta | None = _DEFAULT_RESCAN_INTERVAL,
     ) -> None:
         self._root_path = to_file_path(path)
         self._live = live
@@ -175,10 +176,22 @@ class DirWalker:
             yield file
 
 
-def _stop_observer(observer: object) -> None:
-    """Stop a watchdog observer, suppressing TypeError during interpreter shutdown."""
+def _stop_observer(observer: object, *, join_timeout: float = 5.0) -> None:
+    """Stop and join a watchdog observer.
+
+    All blocking work lives here so callers can offload the entire
+    shutdown to a thread via ``asyncio.to_thread(_stop_observer, obs)``.
+
+    TypeError is suppressed because during interpreter shutdown watchdog's
+    platform observer may already have its C-library globals torn down
+    (e.g. macOS FSEvents calling a function that's been GC'd to None).
+    """
     try:
         observer.stop()  # type: ignore[union-attr]
+    except TypeError:
+        pass
+    try:
+        observer.join(timeout=join_timeout)  # type: ignore[union-attr]
     except TypeError:
         pass
 
@@ -216,7 +229,8 @@ class _LiveDirItems:
 
         root_resolved = self._resolved_root
         loop = asyncio.get_running_loop()
-        rescan_interval = self._walker._rescan_interval
+        td = self._walker._rescan_interval
+        rescan_seconds: float | None = td.total_seconds() if td is not None else None
         # Watchdog dispatches events from a background thread; we forward
         # them into this asyncio.Queue via call_soon_threadsafe.
         events_queue: asyncio.Queue[FileSystemEvent] = asyncio.Queue()
@@ -245,22 +259,15 @@ class _LiveDirItems:
 
             while True:
                 # Compute timeout until next periodic rescan.
-                if rescan_interval is not None:
-                    remaining = rescan_interval - (loop.time() - last_rescan)
+                if rescan_seconds is not None:
+                    remaining = rescan_seconds - (loop.time() - last_rescan)
                     if remaining <= 0:
                         # Periodic rescan cycle: tear down the old
                         # OS-level watcher, drain stale events, create a
                         # fresh watcher (new FSEvents/inotify stream),
                         # then do a full rescan to catch anything the old
                         # watcher may have silently dropped.
-                        _stop_observer(observer)
-                        try:
-                            await asyncio.to_thread(observer.join, 5.0)
-                        except (RuntimeError, TypeError):
-                            try:
-                                observer.join(timeout=5.0)
-                            except TypeError:
-                                pass
+                        await asyncio.to_thread(_stop_observer, observer)
 
                         while not events_queue.empty():
                             try:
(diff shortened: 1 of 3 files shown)


python/cocoindex/connectors/localfs/_source.py
```diff
@@ -90,11 +94,13 @@ def __init__(
         live: bool = False,
         recursive: bool = False,
         path_matcher: FilePathMatcher | None = None,
+        rescan_interval: float | None = _DEFAULT_RESCAN_INTERVAL,
```
georgeh0: Prefer a `datetime.timedelta` here

python/cocoindex/connectors/localfs/_source.py
```diff
@@ -210,20 +225,63 @@ class _Handler(FileSystemEventHandler):
...
+                        # OS-level watcher, drain stale events, create a
+                        # fresh watcher (new FSEvents/inotify stream),
+                        # then do a full rescan to catch anything the old
+                        # watcher may have silently dropped.
+                        _stop_observer(observer)
+                        try:
+                            await asyncio.to_thread(observer.join, 5.0)
+                        except (RuntimeError, TypeError):
+                            try:
+                                observer.join(timeout=5.0)
```
georgeh0: I think this may block the event loop.

Also how about handling join in `_stop_observer()` too? All blocking logic can be there, and we await on a single call to `asyncio.to_thread(_stop_observer, observer)`

python/cocoindex/connectors/localfs/_source.py
jordigilh: Good call — switched to `datetime.timedelta`, consistent with `auto_refresh()`. Fixed in 20822ef.

python/cocoindex/connectors/localfs/_source.py
jordigilh: You're right — that `observer.join()` was blocking the event loop. Consolidated both `stop()` and `join()` into `_stop_observer()` so the entire shutdown runs in a single `asyncio.to_thread(_stop_observer, observer)` call. This also simplified the `finally` block. Fixed in 20822ef.

python/cocoindex/connectors/localfs/_source.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import datetime as _datetime
 import os
 from datetime import datetime
```
georgeh0: I think we may drop the existing `from datetime import datetime` and directly `import datetime` like other libraries

jordigilh: Done — unified to a single `import datetime`, using `datetime.datetime` and `datetime.timedelta` explicitly. Fixed in dfdaab4.

georgeh0:
Hi @jordigilh , PR looks good. Would you please fix the type check failures? Thanks!

jordigilh:
> Hi @jordigilh , PR looks good. Would you please fix the type check failures? Thanks!

done :) 

/Jordi

georgeh0:
Thanks!

PR #2236, opened 2026-06-29
fix(rust): LanceDB nullable schema-only row rewrites

prrao87:
## Summary

Same fix as #2235, but for the LanceDB connector in the Rust SDK.

- Implement the Rust LanceDB connector fix for additive nullable schema-only columns.
- Track columns that LanceDB null-backfills during `add_columns`, then let row reconciliation advance CocoIndex tracking without opening the mutation path when the only row difference is a newly added nullable column with `null`.
- Preserve normal row mutations for real changes: non-null new values, existing-column changes, missing-row uncertainty, and non-nullable additions still upsert.
- Add focused Rust regressions for the row reconciliation behavior and the schema-only table-version path.

## Tests

- `cargo test -p cocoindex --features lancedb lancedb_nullable_schema_only_add_does_not_upsert_rows`
- `cargo test -p cocoindex --features lancedb row_reconcile`
- `cargo fmt --check`

### DIFF
--- a/rust/sdk/cocoindex/src/lancedb.rs
+++ b/rust/sdk/cocoindex/src/lancedb.rs
@@ -10,7 +10,7 @@
 //!
 //! Built on the native Rust `lancedb` crate (LanceDB's core is Rust) + Arrow.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


badmonster0:
rust sdk too ;) ty!

prrao87:
>rust sdk too ;) ty!

🦀 🫶🏽

PR #2239, opened 2026-06-30
fix: reject rows with missing or null primary keys

MaxRong:
## Summary

Reject SQLite target rows whose primary-key values are missing or `None` before declaring target state.

## Why

Dict rows are projected with `row.get(column)`. That is correct for nullable non-key columns, but it allowed a missing primary-key field to become `None` and enter CocoIndex as the row identity before SQLite validation ran.

Without early validation, CocoIndex can compute a target-state identity from `None` even though SQLite would later reject or mishandle the row. The fix makes the connector reject invalid identities at the user-facing boundary while keeping nullable non-key fields unchanged.

## What changed

- Validate every configured primary-key column in `TableTarget.declare_row()` after `_row_to_dict(row)` and before target-state declaration.
- Raise `ValueError` when a dict row omits a primary-key column.
- Raise `ValueError` when any projected primary-key value is `None`.
- Preserve the existing missing-to-`NULL` behavior for nullable non-primary-key dict fields.
- Clarify SQLite row documentation so primary-key values must be present and non-`None`, while nullable non-key dict omissions still write SQL `NULL`.

## Breaking changes

None for valid rows. Invalid rows that already violated the documented primary-key contract now fail earlier, before target-state declaration.

## Follow-up

A similar missing-or-null primary-key risk appears in the PostgreSQL target path (`row.get(col_name)` projection, then primary keys derived from `row_dict`). This PR stays scoped to SQLite and leaves PostgreSQL as a separate follow-up.

## Tests

- `uv run pytest python/` 
- `uv run ruff format --check python/cocoindex/connectors/sqlite/_target.py python/tests/connectors/test_sqlite_target.py` 
- `uv run mypy`

### DIFF
--- a/docs/src/content/docs/connectors/sqlite.mdx
+++ b/docs/src/content/docs/connectors/sqlite.mdx
@@ -141,7 +141,7 @@ def TableTarget.declare_row(
 
 **Parameters:**
 
-- `row` — A row object (dict, dataclass, NamedTuple, or Pydantic model). Must include all primary key columns.
+- `row` — A row object (dict, dataclass, NamedTuple, or Pydantic model). Must include all primary key columns with non-`None` values. Dict rows may omit nullable non-primary-key columns; omitted values are written as SQL `NULL`.
 
 ### Table schema: from Python class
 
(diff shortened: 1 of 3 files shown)


badmonster0:
thanks @MaxRong , @georgeh0 can help take a look!

georgeh0 (approved):
Thanks for the enhancement!

PR #2240, opened 2026-07-01
feat(postgres): share row sink per database for cross-table atomicity

### LINKED ISSUE #2230: [FEATURE] Use shared target sink instances to achieve stronger atomicity when multiple tables involved within the same database
Currently, for most target connectors, we create a different sink for each table (or similar entities names, depending on terms the specific target), e.g. [postgres code](https://github.com/cocoindex-io/cocoindex/blob/36eca13e0d84c128ae72a9874757c9e93736af0d/python/cocoindex/connectors/postgres/_target.py#L690) (per `_RowHandler`; we create a separate `_RowHandler` per postgres table).

Ideally we should have target sink in larger granularity, depending on the transactional boundary supported by the specific target. e.g. for Postgres it should be per database, so we can apply changes to different tables with the same database within a single transaction.

---

tim901231:
## Description of the changes

Row targets that share the same Postgres database now apply their writes through a single shared connection within one transaction, instead of each row target owning its own sink and connection.

- Introduced a module-level `_ROW_SINK` shared across row handlers.
- Row actions now carry a reference to their owning handler.
- Actions are grouped by connection pool (run concurrently across different databases), then by table, with each database's writes executed sequentially inside a single `async with pool.acquire() as conn: async with conn.transaction()` block.
- Replaced the per-target parallel `TaskGroup` upsert/delete scheduling with sequential execution on the shared connection.
- Added `test_postgres_cross_table_row_atomicity`, verifying that row writes to two tables sharing the same database are atomic. Full `prek run --all-files` passes locally.

## Motivation and context

Previously, each row target had its own `TargetActionSink`, so writes to different tables in the same database ran on separate connections and were not atomic — a failure partway through could leave rows committed in one table but not another. Issue #2230 asks for target sinks at a larger granularity so related writes can be committed together.

## Breaking change

None.

## Related issues

Closes #2230

## Open question / seeking feedback

This is my first contribution here, so I'd appreciate a sanity check on direction: I implemented the shared sink at the **per-database** granularity (all row targets on the same Postgres DB share one connection + transaction). Does this match the granularity you had in mind for #2230, or did you envision a different design to solve this? Happy to adjust.


badmonster0:
thanks @tim901231 , @georgeh0 can help take a look!

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -870,12 +871,54 @@ def reconcile(
...
+
+    async with pool.acquire() as conn:
+        async with conn.transaction():
+            for handler, handler_actions in by_handler.items():
+                await handler._execute_actions(conn, handler_actions)
+
+
+_ROW_SINK: coco.TargetActionSink[_RowAction] = coco.TargetActionSink.from_async_fn(
+    _apply_row_actions
+)
```
georgeh0: I think we can have a sink per pool, so we won't need to do another group-by within `_apply_pool_row_actions()` itself.

Besides, ideally we can even make all actions (both row level and table level) share the same sink, so there'll be ACID across all operations for the same Postgres database.

georgeh0 (commented):
Thanks for the contribution!

python/cocoindex/connectors/postgres/_target.py
tim901231: Hi George,

I now have a better understanding of having the row handlers share one sink per pool, and I've also changed the table handler from using a single global sink to one sink per pool.

Where I'm stuck is the part about sharing a sink across the row handler and the table handler within the same pool. I'm not sure how to evolve the current "return child handler" design into a unified one — could you clarify what you had in mind there? Would love to learn from you.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -971,67 +1007,80 @@ class _TableAction(NamedTuple):
 class _TableHandler(coco.TargetHandler[_TableSpec, _TableTrackingRecord, _RowHandler]):
     """Handler for table-level target states."""
 
-    _sink: coco.TargetActionSink[_TableAction, _RowHandler]
+    # One sink per database, keyed by db_key. Sharing a sink across all tables
+    # of the same database lets the engine batch their DDL into a single
+    # connection/transaction, so schema changes across tables are atomic.
+    _table_sinks: dict[str, coco.TargetActionSink[_TableAction, _RowHandler]]
```
georgeh0: Actually since we already have this per-db dict, we don't need another dict `_pool_to_row_sink`.

Actually we may even be able to remove this `dict`. e.g. something like:

```python
@dataclass(frozen=True)
class DbSink:
  _pool: ContextKey[asyncpg.Pool]

  async def __call__(self, context_provider: ContextProvider, actions: ...):
    ...
```

This guarantees equality on all `DbSink` on the same pool.
Then we can directly use `coco.TargetActionSink.from_async_fn(DbSink(db_key))` at certain entry points.

python/cocoindex/connectors/postgres/_target.py
georgeh0: > Where I'm stuck is the part about sharing a sink across the row handler and the table handler within the same pool. I'm not sure how to evolve the current "return child handler" design into a unified one — could you clarify what you had in mind there? Would love to learn from you.

I think when we use a unified sink, we'll need to distinguish table-level and row-level actions by the action itself (e.g. either use different types, or use a field to distinguish). Based on that, we can return a child handler for table level actions, or `None` for row level actions. In theory this should work.

Let me know if you encountered any specific difficulty. e.g. difficulty in type hints?

python/cocoindex/connectors/postgres/_target.py
tim901231: Hi George,

Thanks for the mentoring! I've created a shared per-database sink and route actions by type inside the apply function. While doing that I realized the create vs. drop ordering needs care, so I group the table and row actions into three phases within the single transaction:

1. table DDL that leaves the table existing (create / alter / replace) — must run before any row DML targeting it;
2. row DML (upserts + deletes);
3. table drops — must run after row DML, since deleting rows from an already-dropped table would fail.

This makes the code a little bit messy, and I'd like to ask your opinion on whether this handling is correct. Would love to learn from you.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1239,18 +1234,93 @@ def reconcile(
...
+# One sink per database, keyed by db_key. Both table (DDL) and row (DML) target
+# states of the same database route through this single sink, so the engine
+# batches them into one connection/transaction when they are applied in the same
+# pass — giving cross-table, and mixed DDL+DML, atomicity within one database.
+_DbAction = _TableAction | _RowAction
+
+_db_sinks: dict[str, coco.TargetActionSink[_DbAction, _RowHandler]] = {}
+
+
+def _get_db_sink(
```
georgeh0, line 1259: Since this is a synchronous function and can be called from different threads, there can be race conditions in read-write on `_db_sinks`.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1239,18 +1234,93 @@ def reconcile(
                 main_action=main_action,
                 column_actions=column_actions,
             ),
-            sink=self._sink,
+            sink=_get_db_sink(key.db_key),
```
georgeh0, line 1237: After a recent change #3404, this can be simplified a lot. e.g. we can wrap the `db_key` in a `NamedTuple` similar to [this](https://github.com/cocoindex-io/cocoindex/blob/8704d3f027b855812fed56744b2d2326797e24ab/python/tests/core/test_target_sink_identity.py#L43-L54).

tim901231: Hi George,

Thanks for your guidance! I’ve removed _db_sinks and replaced it with a callable NamedTuple keyed by db_key, using the engine’s synchronized value-based sink identity. I also added a concurrent regression test confirming that sinks with the same key share an identity and different keys remain separate.

Could you take another look when you have a chance? I appreciate your help!

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
+_table_handler = _TableHandler()
 _table_provider = coco.register_root_target_states_provider(
-    "cocoindex/postgres/table", _TableHandler()
+    "cocoindex/postgres/table", _table_handler
 )
 
 
+# One sink per database, keyed by db_key. Both table (DDL) and row (DML) target
+# states of the same database route through this single sink, so the engine
+# batches them into one connection/transaction when they are applied in the same
```
georgeh0, line 1257: Note that actions from one component are always one transaction; cross-component merging is opportunistic.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -720,16 +730,15 @@ async def _apply_actions(
             else:
                 upserts.append(action)
 
-        async with asyncio.TaskGroup() as tg:
-            self._schedule_upserts(tg, upserts)
-            self._schedule_deletes(tg, deletes)
+        await self._execute_upserts(conn, upserts)
+        await self._execute_deletes(conn, deletes)
```
georgeh0, line 734: Now let's run deletes before upserts.

python/cocoindex/connectors/postgres/_target.py
georgeh0, line 519: This is still on its own connections. Consider routing through the per-database sink.

georgeh0, line 637: (similar to the comment above for L519) This is still on its own connections. Consider routing through the per-database sink.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -685,29 +688,36 @@ class _RowHandler(coco.TargetHandler[_RowValue, _RowFingerprint]):
     """Handler for row-level target states within a table."""
 
     _pool: asyncpg.Pool
+    _db_key: str
```
georgeh0, line 691: I think we can resolve the sink once in _apply_table_action (L986) and pass `sink` into `__init__`. This uses a more specific type instead of a weaker one (like `str`), and avoids global lock acquisitions per row from `reconcile()`.

georgeh0:
Heads-up from #2404 (landing now): `TargetActionSink.from_fn` / `from_async_fn` now reject a callback that does not support weak references with a `TypeError`, and sink identity is keyed by the callback's exact type as well as `==`.

`class _DbSink(NamedTuple)` in `python/cocoindex/connectors/postgres/_target.py` will fail that check the first time `reconcile()` constructs a sink: a NamedTuple instance cannot be weakly referenced (so an idle sink identity could never be released), and it compares equal to any same-shaped tuple (so its identity was not type-safe). Switching it to `@dataclasses.dataclass(frozen=True)`, as already requested in review, satisfies both; the per-database batching behavior is otherwise unchanged.

The rationale and the tightened "Shared Action Sinks" guideline are in #2404 and `dev/agent-skills/target-connector/SKILL.md`.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
             ),
-            sink=self._sink,
+            sink=_get_db_sink(key.db_key),
             tracking_record=tracking_record,
             child_invalidation=child_invalidation,
         )
 
 
 # Register the root target states provider
+_table_handler = _TableHandler()
```
georgeh0, line 1249: `_TableHandler` is stateless now, so we can make `_apply_table_action` (and its DDL helpers) module-level functions as sqlite/doris do, and inline `_TableHandler()` in the provider registration again. Removes the singleton and the cross-object private calls at L1300/L1306.

python/tests/connectors/test_postgres_target.py
```diff
@@ -1119,6 +1119,187 @@ def test_sanitize_nul_preserves_dict() -> None:
...
+async def test_postgres_cross_table_ddl_atomicity(pg_env: _PgEnv) -> None:
+    """Table DDL across two tables sharing the same database must be atomic.
+
+    The table handler uses one sink per database (keyed by db_key), so table
+    actions declared in the same processing component reconcile into a single
+    ``_apply_actions`` batch and run in one connection/transaction. If one
+    table's ``CREATE TABLE`` fails (here: an invalid column type), the other
+    table declared in the same ``app.update()`` must NOT be created either --
+    the shared transaction rolls back both.
+    """
```
georgeh0, line 1231: Two fixes: `_apply_actions` no longer exists (L1227); and the guarantee shown holds only because both tables are declared bare in one component. With use_mount` — as the sibling test at L1123 and all docs do — each table is its own component, and the second table's DDL failure leaves the first committed (verified). Say so in the docstring, or add the mounted variant to demonstrate the difference.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
+        # Partition actions into three ordered phases so the transaction is
+        # correct by construction, independent of the order the engine
+        # batches actions in:
+        #   1. table DDL that leaves the table existing (create / alter /
+        #      replace) — must run before any row DML targeting it;
+        #   2. row DML (upserts + deletes);
+        #   3. table drops — must run after row DML, since deleting rows
+        #      from an already-dropped table would fail.
+        # `replace` drops-then-recreates internally, so it ends with the
+        # table existing and belongs to phase 1.
```
georgeh0, line 1284: Actually DDL and DML for the same table are never batched together. Comments like this may be more accurate:

```python
        # Apply in a fixed order: (1) table DDL that leaves the table existing
        # — create / alter / replace (`replace` drops-then-recreates, so it
        # belongs here); (2) row DML, grouped per table; (3) table drops.
        #
        # A batch mixes table and row actions only when the batcher coalesces
        # submissions from different components: within one component the
        # engine applies a table's DDL (and commits) before that table's rows
        # can be reconciled, and rows of a table being dropped are never
        # reconciled at all. So no reachable batch requires this ordering
        # today; it pins a deterministic, dependency-safe order (creates
        # first, drops last) so correctness doesn't hinge on how the engine
        # happens to concatenate components' actions.
```

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
+            else:
+                rows_by_handler.setdefault(action.handler, []).append(action)
+
+        async with pool.acquire() as conn:
+            async with conn.transaction():
+                for i, action in ensure_table_actions:
+                    outputs[i] = await _table_handler._apply_table_action(
+                        conn, pool, action
+                    )
+                for handler, handler_actions in rows_by_handler.items():
```
georgeh0, line 1303: This iterates in insertion order, i.e. the order actions arrived in the batch, which differs from batch to batch. Since the whole batch is now one transaction, two concurrent updates of the same database from different processes can lock tables in opposite orders and deadlock.

Please make the lock order deterministic so every batch acquires table locks in the same sequence — e.g. iterate `sorted(rows_by_handler.items(), key=lambda kv: (kv[0]._schema_name or "", kv[0]._table_name))`, and sort `ensure_table_actions` / `drop_table_actions` by `action.key` as well. That removes cross-table cycles outright; the retry below covers what's left.

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
+                    await handler._execute_actions(conn, handler_actions)
+                for i, action in drop_table_actions:
+                    outputs[i] = await _table_handler._apply_table_action(
+                        conn, pool, action
+                    )
```
georgeh0, line 1308: Even with a deterministic table order, this transaction can still be aborted transiently: row-level lock cycles against another process updating the same rows in a different order, or another application's transactions, surface as `DeadlockDetectedError` / `SerializationError` — both subclasses of `asyncpg.TransactionRollbackError`. Postgres kills exactly one side and the other proceeds, so the aborted side just needs to run again.

We'll need to retry on `asyncpg.TransactionRollbackError`

python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1246,18 +1239,88 @@ def reconcile(
...
 
 
+# One sink per database, keyed by db_key. Both table (DDL) and row (DML) target
+# states of the same database route through this single sink, so the engine
+# batches them into one connection/transaction when they are applied in the same
+# pass — giving cross-table, and mixed DDL+DML, atomicity within one database.
+_DbAction = _TableAction | _RowAction
+
+
+class _DbSink(NamedTuple):
```
georgeh0, line 1262: Oh, actually I didn't notice there's a pitfall by using `NamedTuple` here. We should use a `@dataclass(frozen=True)` instead, and #2404 contains the clarification with the reasons.

Issue #2123, opened 2026-06-16
[FEATURE] support sparse vector


badmonster0:
**What is the use case?**

https://www.reddit.com/r/Rag/comments/1u7eypx/can_cocoindex_generate_and_store_both_dense/

---
❤️ Contributors, please refer to 📙[Contributing Guide](https://cocoindex.io/docs/about/contributing).
Unless the PR can be sent immediately (e.g. just a few lines of code), we recommend you to leave a comment on the issue like **`I'm working on it`**  or **`Can I work on this issue?`** to avoid duplicating work. Our [Discord server](https://discord.com/invite/zpA9S2DR7s) is always open and friendly.

tomz-alt:
assign me

badmonster0:
@tomz-alt let us know if you plan to do a follow up on this, thanks!

tomz-alt:
yes, my plan is to support postgres and zvec together with supported qdrant do you have some more inputs on here?

badmonster0:
sg, thanks @tomz-alt !

PR #2242, opened 2026-07-04
feat(qdrant): native sparse vector support + eager point-ID validation

tomz-alt:
Adds native sparse vector support to the Qdrant connector, plus eager point-ID validation and corrections to the Qdrant docs.

## Sparse vectors

- New `QdrantSparseVectorDef` and `CollectionSchema.create(..., sparse_vectors={...})` (keyword-only). Sparse vectors are always named, per Qdrant's model; `modifier="idf"` maps to `models.Modifier.IDF`.
- Collections are created with native `sparse_vectors_config` — sparse data lives in `PointStruct.vector`, never in payload. `vectors_config=None` is supported for sparse-only collections.
- The collection tracking record now includes the sparse config, so adding/changing sparse vectors triggers collection replacement (destructive child invalidation, same as dense schema changes).

**Upgrade safety (data-loss guard):** the new field is a trailing `msgspec` field with a default, so tracking records written before this change decode equal to a new dense-only record — existing collections are *not* spuriously drop-and-recreated on upgrade. This is pinned by `TestTrackingRecordUpgradeCompat`, which round-trips a pre-sparse-shaped record through the engine's real serde hooks.

## Point-ID validation (and docs correction)

Qdrant only accepts **unsigned 64-bit integers and UUIDs** as point IDs — [Qdrant: Point IDs](https://qdrant.tech/documentation/manage-data/points/#point-ids). Confirmed empirically against Qdrant 1.18 over both REST and gRPC: arbitrary strings like `"doc-123"` are rejected by the server; UUID strings are accepted in hyphenated, 32-char-hex, and URN forms; ints outside `[0, 2**64)` are rejected.

- `declare_point` now validates eagerly and raises `ValueError` with the rule, the doc link, and a `uuid.uuid5` recipe for string-derived stable IDs.
- **Failure-mode note for reviewers:** this moves the failure from apply time to declare time. Previously an invalid ID surfaced as an opaque transport error during sink apply (after the collection side effects); now it raises inside the component before any side effects. No previously-working ID is affected — invalid IDs never succeeded against a real server.
- Docs: the "Point IDs" section previously claimed arbitrary strings work ("All other types are converted to strings automatically") — rewritten to the real rule with the official link; the two `id="doc-123"` examples fixed; search examples migrated from the removed `client.search` to `query_points` (incl. hybrid RRF example); `face_recognition` drops its dead `client.search` fallback, fixing full `mypy`.

## Tests

New: sparse param resolution; fake-client `create_collection` kwarg forwarding; sparse-vector participation in point fingerprints; the full point-ID accept/reject matrix; tracking-record upgrade compat; and a `QDRANT_URL`-gated live test that creates a dense+sparse collection through CocoIndex, upserts, and queries dense, sparse, and hybrid (RRF fusion) successfully.

**Mutation checks** (each fix verified to be load-bearing):
- Removing `sparse_vectors_config` from `create_collection` → fake-client test fails.
- Moving the sparse vector into payload instead of `PointStruct.vector` → live sparse query returns `[]`, test fails.
- Neutering ID validation → `test_arbitrary_string_rejected` fails.

## Verification

- `pytest python/tests/connectors/test_qdrant_target.py` with live Qdrant (Docker, gRPC): 25 passed
- Full `pytest python/`: 910 passed, 232 skipped
- `uv run mypy`: clean (260 files; includes the `face_recognition` fix)
- `uv run ruff format --check .`: clean; docs build: clean

### DIFF
--- a/python/cocoindex/connectors/qdrant/_target.py
+++ b/python/cocoindex/connectors/qdrant/_target.py
@@ -7,9 +7,10 @@ This module provides a two-level target state system for Qdrant:
 """
 
 from __future__ import annotations
-import cocoindex as coco
 
 import asyncio
+import uuid
+
 import msgspec
 from dataclasses import dataclass
 from typing import (
@@ -68,6 +69,17 @@ class QdrantVectorDef(NamedTuple):
     multivector_comparator: Literal["max_sim"] = "max_sim"
 
 
+class QdrantSparseVectorDef(NamedTuple):
+    """Qdrant sparse vector specification.
+
+    Sparse vectors in Qdrant are always named and use dot-product scoring.
+    ``modifier="idf"`` enables Qdrant's IDF modifier; ``None`` (the default)
+    applies no modifier.
+    """
+
+    modifier: Literal["idf"] | None = None
+
+
 class _ResolvedQdrantVectorDef(msgspec.Struct, frozen=True, tag=True):
     """Resolved single (unnamed) vector specification.
 
@@ -86,6 +98,18 @@ class _ResolvedQdrantNamedVectorsDef(msgspec.Struct, frozen=True, tag=True):
     vectors: dict[str, _ResolvedQdrantVectorDef]
 
 
+class _ResolvedQdrantSparseVectorDef(msgspec.Struct, frozen=True, tag=True):
+    """Resolved sparse vector specification."""
+
+    modifier: Literal["idf"] | None = None
+
+
+class _ResolvedQdrantSparseVectorsDef(msgspec.Struct, frozen=True, tag=True):
+    """Resolved named sparse vectors specification."""
+
+    sparse_vectors: dict[str, _ResolvedQdrantSparseVectorDef]
+
+
 async def _resolve_vector_def(vector_def: QdrantVectorDef) -> _ResolvedQdrantVectorDef:
     resolved_schema: res_schema.VectorSchema | res_schema.MultiVectorSchema
     vs = await res_schema.get_vector_schema(vector_def.schema)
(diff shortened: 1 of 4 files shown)


badmonster0:

python/cocoindex/connectors/qdrant/_target.py
```diff
@@ -132,53 +162,99 @@ class CollectionSchema:
...
+                "Qdrant collection schema must declare dense/multivector or sparse vectors"
+            )
         self._vectors = vectors
+        self._sparse_vectors = sparse_vectors
 
     @classmethod
     async def create(
         cls,
-        vectors: QdrantVectorDef | dict[str, QdrantVectorDef],
+        vectors: QdrantVectorDef | dict[str, QdrantVectorDef] | None = None,
```
georgeh0: What do you think about using 

```python
QdrantVectorDef | dict[str, QdrantVectorDef | QdrantSparseVectorDef] | None
```

instead of a separate `sparse_vectors` here?

Similar changes to `_ResolvedQdrantSparseVectorsDef`: the vector wrapper type for sparse vector can be omitted.

Since dense and sparse vectors share the same namespace and cannot collide on name, this way will better enforce this constraint.

georgeh0 (commented):
Thanks for supporting sparse vector for Qdrant!

Also note that the original motivation for #2123 is to add sparse vector support for CocoIndex targets in general, e.g. pgvector also supports sparse vector. With this said, these can happen on different PRs.

python/cocoindex/connectors/qdrant/_target.py
tomz-alt: agreed and implemented in [862a96c](https://github.com/cocoindex-io/cocoindex/pull/2242/commits/862a96ce3283987d0eedaf0feab968a8bbff23f9)

badmonster0:
thanks @tomz-alt  follow up on @georgeh0 's comment - original motivation for https://github.com/cocoindex-io/cocoindex/issues/2123 is to add sparse vector support for CocoIndex targets in general, e.g. pgvector also supports sparse vector. do you like to create a separate issue on it that i can assign it to you or have it covered in this pr?

tomz-alt:
> thanks @tomz-alt follow up on @georgeh0 's comment - original motivation for #2123 is to add sparse vector support for CocoIndex targets in general, e.g. pgvector also supports sparse vector. do you like to create a separate issue on it that i can assign it to you or have it covered in this pr?

I can do it in a separate PR.

badmonster0:
sg, @georgeh0 can help another look!

georgeh0 (approved):
Thanks!

PR #2244, opened 2026-07-04
fix(postgres): recursively strip NUL (U+0000) from array and composite bindings

Sujit-1509:
## Summary

Fixes a bug where strings containing NUL (`U+0000`) inside Postgres array (`text[]`) or composite values bypass sanitization and cause `asyncpg` to fail with:
```
ValueError: string cannot contain NUL (0x00) characters
```

### Root Cause
In `postgres/_target.py`, `_row_to_dict` only stripped NUL from scalar string values and JSONB column payloads. Strings bound within native array columns (`text[]`, `varchar[]`) and composite types (`tuple`) bypassed sanitization completely.

### Fix
1. **Extend `_sanitize_nul`:** Updated the walker to recursively strip NUL from lists and tuples, ensuring that `tuple` objects return `tuple` (crucial for `asyncpg` composite type adapters) and `list` objects return `list`.
2. **Apply to Containers in `_row_to_dict`:** Updated parameter mapping to apply `_sanitize_nul` to list and tuple structures.
3. **Tests:** Added comprehensive regression tests to verify that `text[]` columns correctly strip NUL from array strings and unit tests to ensure collections retain correct type structures.

Refs #933

### DIFF
--- a/python/cocoindex/connectors/postgres/_target.py
+++ b/python/cocoindex/connectors/postgres/_target.py
@@ -1301,15 +1301,10 @@ class TableTarget(
 
             if value is not None and col.encoder is not None:
                 value = col.encoder(value)
-            # Strip NUL from values bound to Postgres.  Scalar strings hit
-            # the fast path; lists/tuples (array and composite columns) are
-            # walked recursively.  jsonb columns are already handled inside
-            # `_json_encoder`, so the recursive walk is a harmless no-op on
-            # the encoded JSON string.
-            if isinstance(value, str):
-                value = _strip_nul(value)
-            elif isinstance(value, (list, tuple)):
-                value = _sanitize_nul(value)
+            # Strip NUL from values bound to Postgres recursively. jsonb columns
+            # are already handled inside `_json_encoder`, so the recursive walk
+            # is a harmless no-op on the encoded JSON string.
+            value = _sanitize_nul(value)
             out[col_name] = value
         return out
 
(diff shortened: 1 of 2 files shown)


python/cocoindex/connectors/postgres/_target.py
```diff
@@ -1296,14 +1301,15 @@ def _row_to_dict(self, row: RowT) -> dict[str, Any]:
...
+            # Strip NUL from values bound to Postgres.  Scalar strings hit
+            # the fast path; lists/tuples (array and composite columns) are
+            # walked recursively.  jsonb columns are already handled inside
+            # `_json_encoder`, so the recursive walk is a harmless no-op on
+            # the encoded JSON string.
             if isinstance(value, str):
                 value = _strip_nul(value)
+            elif isinstance(value, (list, tuple)):
+                value = _sanitize_nul(value)
             out[col_name] = value
```
georgeh0: `dict` is missed here. Actually I think we can blindly call `_sanitize_nul(value)` here for all types, since `_sanitize_nul()` already checks for types.

Sujit-1509: Good catch, thanks! Blindly calling `_sanitize_nul(value)` simplifies the logic nicely and ensures `dict `bindings are sanitized correctly too.

I've updated `_row_to_dict` to call `_sanitize_nul(value)` directly and added a unit test `test_sanitize_nul_preserves_dict` to verify dict sanitization. Pushed the update.

georgeh0 (approved):
Thanks for the enhancement!

PR #2245, opened 2026-07-04
feat(lancedb): optimize table maintenance from LanceDB stats

prrao87:
## Summary

This PR changes the LanceDB connector's automatic table maintenance from a transaction-count trigger to a stats-driven trigger.

Previously, the connector ran `table.optimize()` after `num_transactions_before_optimize` mutation batches. That was a weak proxy for actual table/index maintenance debt: it's incredibly common to see scenarios where there are many small updates across short runs might never reach the threshold, while one large batch could leave a large unindexed tail until a later run.

### UX improvement
The connector now decides based on LanceDB's own table/index state. The user no longer needs to worry about a `num_transactions_before_optimize` kwarg (this has been removed in this version).

## Why Is This Change Needed?

In LanceDB, table optimization involves physical maintenance of the underlying files: it compacts small fragments, cleans up deletion files, and reindexes new data so the freshest rows are available through declared indexes.

A transaction-count trigger does not reliably represent that physical maintenance debt. Batch counts vary widely depending on runtime scheduling: a tiny update can be one batch, and a large backfill can also arrive as only one or two batches. That means the old design could miss both important ends of the workload spectrum:

- many small, sporadic updates across short runs could build up fragments and unindexed rows indefinitely
- a large update into an existing indexed table could leave a big unindexed tail until some later run happened to cross the transaction threshold

The new design implemented here uses _durable LanceDB stats_ instead: fragment counts, deletion file counts, old prunable versions, and per-index unindexed row counts. Checks are driven by mutated row count rather than batch count, so the cadence tracks actual write volume.

## What Changed

- Replace `num_transactions_before_optimize` with internal stats-driven optimize thresholds.
- Trigger `table.optimize()` from:
  - small fragment count from `table.stats()`
  - deletion file count from latest version metadata
  - prunable old version count
  - `num_unindexed_rows` from every declared index
- Use mutation-row count, not batch count, to decide when to re-check stats.
- Add an internal cooldown between optimize attempts to avoid repeated optimize calls during high-write streams.
- Treat optimize evaluation/execute failures as non-fatal: mutations still succeed and failures are logged.
- Handle LanceDB version timestamps correctly when timestamps are naive local datetimes.
- Update docs to explain optimize behavior and external concurrency guidance.
- Bump LanceDB Python SDK dependency to `>=0.34.0`.

## Breaking Change

This removes the optional `num_transactions_before_optimize` kwarg.

I think removing it without a deprecation period is acceptable here because:

- It was an optional tuning kwarg that most users likely never touched
- This is still a relatively new connector with limited usage
- Keeping the option would preserve the wrong mental model: optimize should be driven by physical table/index state, not by transaction count

## Validation

Verified against the most recent LanceDB Python SDK `0.34.0`.

- `uv run --extra lancedb python -c 'import lancedb; print(lancedb.__version__)'`
  - confirmed `0.34.0`
- `uv run --extra lancedb pytest python/tests/connectors/test_lancedb_target.py`
  - `40 passed`
- `uv run ruff check python/cocoindex/connectors/lancedb/_target.py python/tests/connectors/test_lancedb_target.py`
  - passed
- `uv run maturin build --out target/wheels`
  - passed

Also ran a real smoke test through [cocoindex-v1-intro](https://github.com/thedataquarry/cocoindex-v1-intro) with LanceDB `0.34.0`:

- bulk ingested 1000 records
- then sporadically updated 10, 20, and 30 records
- optimize triggered once the cumulative unindexed index tail crossed the internal threshold
- after optimize, fragments dropped to 3 and unindexed rows dropped to 4

## Scenarios This Protects Against

I went through multiple iterations to stress-test the design with Fable 5, and I think it should be way more robust than the earlier version. Here's a summary of various scenarios users may experience in the wild that this design protects against.

| # | Scenario | What happens with stats-driven optimize | Outcome |
|---|---|---|---|
| 1. | **Initial bulk ingest** (10K+ rows, new table) | Rows land in 1-2 large batches; index is created *after* rows, so `create_index` covers everything | No tail; no optimize needed |
| 2. | **Sporadic updates** (10-20 rows, months apart) | Fresh `_RowHandler` per run, so the first batch of every run always evaluates stats, repaying any debt from prior runs | Nothing accumulates indefinitely, regardless of run frequency |
| 3. | **Bulk backfill into existing indexed table** | Stats checks are row-count-driven (every >=25 mutated rows): the bulk batch itself crosses the threshold, so the post-backfill state is evaluated and optimized in the same run | Tail integrated immediately; at most ~1K rows can remain unindexed at exit (bounded by the optimize floor) |
| 4. | **Long-lived watcher, tiny updates** | Stats round-trip only per >=25 mutated rows, and after any optimize attempt a 1,000-row floor gates further attempts | Check overhead scales with mutation volume; optimize frequency bounded even on small tables |
| 5. | **High-throughput continuous writes** | Batches coalesce large; 20K absolute / 5% relative unindexed caps bound cadence; optimize runs inline, serialized with row writes; 1,000-row floor between attempts prevents thrash | Backpressure instead of commit conflicts; every trigger is physically cleared by optimize |
| 6. | **Concurrent readers during optimize** | MVCC snapshots; default 7-day cleanup horizon retains recent versions | Reads unaffected, long-lived checkouts safe |
| 7. | **Stats/optimize failure** (e.g. remote table) | Exception logged, row mutations succeed; retried after 1,000 more mutated rows or on next run's first batch | Non-fatal with natural retry, no failure-retry loop |

Trigger design: checks are driven by **mutated row count** (not batch count), so cadence tracks actual write volume at both extremes: a 10K-row backfill is evaluated as soon as it lands, while a trickle of single-row updates only pays a stats check per 25 mutated rows. The 1,000-row floor between optimize attempts caps worst-case optimize frequency independently of table size.

## Documentation

Updated the docs page to better describe this change, and briefly explain this in case the user wants to learn more.

### DIFF
--- a/docs/src/content/docs/connectors/lancedb.mdx
+++ b/docs/src/content/docs/connectors/lancedb.mdx
@@ -71,7 +71,7 @@ async def coco_lifespan(builder: coco.EnvironmentBuilder) -> AsyncIterator[None]
 
 #### Tables (parent state)
 
-Declares a table as a target state. Returns a `TableTarget` for declaring rows.
+Declares a table as a target state. Returns a pending `TableTarget` for declaring rows.
 
 ```python
 def declare_table_target(
@@ -79,7 +79,7 @@ def declare_table_target(
     table_name: str,
     table_schema: TableSchema[RowT],
     *,
-    managed_by: Literal["system", "user"] = "system",
+    managed_by: ManagedBy = ManagedBy.SYSTEM,
 ) -> TableTarget[RowT, coco.PendingS]
 ```
 
@@ -89,8 +89,34 @@ def declare_table_target(
 - `table_name` — Name of the table.
 - `table_schema` — Schema definition including columns and primary key (see [Table Schema](#table-schema-from-python-class)).
 - `managed_by` — Whether CocoIndex manages the table lifecycle (`"system"`) or assumes it exists (`"user"`).
+  `ManagedBy` is exported from `cocoindex.connectorkits.target`.
 
-**Returns:** A pending `TableTarget`. Use the convenience wrapper `await lancedb.mount_table_target(LANCE_DB, table_name, table_schema)` to resolve.
+**Returns:** A pending `TableTarget`. Resolve it with `await target`.
+
+For the common case where you want a ready-to-use target immediately, use `mount_table_target()`:
+
+```python
+async def mount_table_target(
+    db: ContextKey[LanceAsyncConnection],
+    table_name: str,
+    table_schema: TableSchema[RowT],
+    *,
+    managed_by: ManagedBy = ManagedBy.SYSTEM,
+) -> TableTarget[RowT]
+```
+
+For lower-level composition with `coco.mount_target()`, `table_target()` returns the raw target
+state:
+
+```python
+def table_target(
+    db: ContextKey[LanceAsyncConnection],
+    table_name: str,
+    table_schema: TableSchema[RowT],
+    *,
+    managed_by: ManagedBy = ManagedBy.SYSTEM,
+) -> coco.TargetState
+```
 
 CocoIndex automatically runs LanceDB `table.optimize()` when durable table and index statistics
 cross internal maintenance thresholds, such as small fragments, deletion files, or unindexed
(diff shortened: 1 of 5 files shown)


badmonster0:
thanks for the improve ment!

Issue #2054, opened 2026-06-03
[FEATURE] Unified timeout/deadline mechanism


georgeh0:
We want to introduce a unified timeout/deadline mechanism.

## Timeout set and deadline propagation

Timeout can be set using a context manager:

```python
with coco.timeout(datatime.timedelta(seconds=30)):
  ...
```

Then the deadline (now + timeout duration) is imposed on the body of context manager, which means it's propagated through function calls, `use_mount(...)` calls, etc. But deadline is not propagated through the following paths:
- Components mounted by `mount(...)` (which is outside the critical path).
- Component execution logic outside the processor function, e.g. target states submit. Once processor logic done, we should proceed without deadline (even if there's a deadline propagated through the `use_mount(...)` call). Within the target sink logic there can be explicit deadline imposed there, which is isolated from context of the component.
- Batched function calls, especially if batching really happens. (ideally, for a batched call combined by N callers, if all N callers are cancelled / deadline exceeded, we can cancel the batched call. This is an advanced feature request and can be resolved in a separate issue after this one resolved)

Once there's already a deadline in the context, if `coco.timeout(...)` is imposed on a smaller scope, we take the minimum value.

## Deadline enforcement

Deadline is checked within certain engine logic, and exception is raised if exceeded deadline:

- `coco.mount_*()`, `coco.use_mount()`, calling a cocoindex function, ID generator
- We also expose a `coco.check_deadline()` function to check deadline explicitly and raise exception if exceeded

## Retry logic on transient failures

For retry on transient failures, previously we're capping by retry times, which is ad-hoc ([example](https://github.com/cocoindex-io/cocoindex/blob/57ea82a87bea41a515e987e387849139702cf1da/python/cocoindex/ops/entity_resolution/llm_resolver.py#L105-L120)). With deadline, the standard way of retry will be: backoff and retry on transient failures, until deadline is reached.

Gohlub:
Is this open to contributions?

badmonster0:
hi @Gohlub , we have an contributor looking at it at the moment, thanks for checking!

Gohlub:
No problem, thanks for letting me know quickly

tomz-alt:
assign me

tomz-alt:
Hi @georgeh0 , Here's the design I plan to implement, the invariants it rests on, and the test plan. One question at the end.

### How the deadline flows

[...]

PR #2251, opened 2026-07-05
feat: add cooperative timeout support

tomz-alt:
## Summary

Adds a unified, cooperative timeout/deadline mechanism for CocoIndex.

`coco.timeout(td)` stores an absolute monotonic deadline. CocoIndex checks that deadline at engine-owned checkpoints and raises `DeadlineExceededError` (a `TimeoutError`) once the deadline has passed. This does not cancel arbitrary user awaits; timeouts are observed at cooperative checkpoints.

Addresses #2054.

## Behavior

Deadlines are inherited through critical-path work:

- plain Python calls, `@coco.fn` calls, and `coco.map()` tasks
- root processors started under `with coco.timeout(...): await app.update()`
- `use_mount()` child processors

Deadlines are isolated from shared or background work:

- `mount()`, `mount_each()`, and live components
- target submit and sink bodies
- batched runner bodies
- app drop / cleanup paths

Submit remains deadline-free once processing succeeds. If the deadline expires before submit, processing fails like any other processor exception, with no sink actions and no poisoned memo entry. If the deadline expires while submit is already running, submit completes consistently and the caller observes the timeout before receiving the result.

Deadline checks are intentionally not placed inside submit or ownership-transfer retry loops. If a component is already in the submit pipeline and hits `PendingRetry` precommit backoff, deadline observation is delayed until the submit attempt finishes and the next caller/processor checkpoint runs; this preserves submit isolation and keeps timeout enforcement cooperative outside shared commit work.

## Implementation Notes

- Adds `DeadlineExceededError`, `timeout()`, `check_deadline()`, deadline snapshots, and retry helpers.
- Carries the deadline into Rust core as `DeadlineContext`; `DeadlineContext::NONE` is the single no-deadline representation in core.
- Restores captured Python deadline snapshots in Rust-spawned processor wrappers where the Python `ContextVar` would otherwise be absent.
- Checks deadlines in core before memo lookup, before processor body execution, after processor return and before submit, and when observing app/update or `use_mount()` results.
- Keeps submit, sinks, state-store, and action batching deadline-free by module boundary: `execution.rs` and `state_store/` do not import or reference the deadline concept.
- For `use_mount()`, the child finishes or fails first; then the waiting parent checks its own deadline before using the child result. This avoids a timed-out parent reporting failure while the child is still committing work.
- Uses `asyncio.TaskGroup` for `coco.map()`, drains started tasks, and reports failures deterministically by input order.
- Converts the LLM resolver retry loop to retry until the active deadline while preserving the existing numeric retry cap when no deadline is set.
- Gates testing clock controls behind `COCOINDEX_TESTING=1` and maps invalid PyO3 timeout values to `ValueError` instead of panicking across FFI.
- Adds developer-facing timeout docs.

## Invariant Coverage

- Nested timeout scopes use the earlier deadline and restore the previous context exactly.
- Expired deadlines are checked at processor entry/return, memo hits, `use_mount()` result observation, app update result observation, ID generation, `coco.map()`, and retry sleep boundaries.
- `mount()`, `mount_each()`, live processors, target sinks, and batched runner bodies are deadline-isolated.
- Deadline exceptions raised during processing do not apply sink actions and do not poison memo entries.
- Cleanup paths are deadline-free: app drop under an already-expired ambient deadline completes cleanup externally and internally.
- Rust SDK `Ctx::map()` drains started futures instead of cancelling siblings on the first error, matching the cooperative no-arbitrary-cancellation design.

## Validation

Local checks run on the pushed branch:

- `uv run maturin develop`
- `cargo test -p cocoindex_core` - 72 passed
- `cargo test -p cocoindex` - SDK tests and Rust deadline conformance subset passed
- `cargo test -p cocoindex_py` - passed
- `uv run pytest python/tests/core/test_deadline.py python/tests/ops/test_llm_pair_resolver_deadline.py -q` - 36 passed
- `uv run pytest python/` - 926 passed, 263 skipped
- `uv run mypy` - clean across 263 source files
- `uv run ruff check python/cocoindex/_internal/app.py python/cocoindex/_internal/api.py python/cocoindex/_internal/core.pyi python/cocoindex/_internal/deadline.py python/cocoindex/_internal/function.py python/tests/core/test_deadline.py python/tests/ops/test_llm_pair_resolver_deadline.py` - clean
- `cargo fmt --check` - clean
- `uv run ruff format --check python/cocoindex/_internal/app.py` - clean

Focused mutation proof already run for the post-submit deadline checkpoints:

- Temporarily removed the post-update result checkpoints.
- `test_update_blocking_checks_captured_deadline_after_submit` and `test_update_handle_checks_captured_deadline_after_submit` both failed with `DID NOT RAISE`.
- Restored the checkpoints and reran the deadline suite successfully.

## Follow-ups Out Of Scope

- Cancelling arbitrary user awaits.
- Propagating deadlines into user-created threads.
- Cancelling shared batched work once all callers expire.
- Rust-side per-environment test clock or feature-gated test clock plumbing beyond the current env-gated API.


georgeh0 (commented):
I'm expecting the core logic to maintain and check deadline/timeout lives on Rust side. It'll be easily shared across SDKs in different languages, and the in-engine timeout check will also be far more efficient. What do you think?

tomz-alt:
> I'm expecting the core logic to maintain and check deadline/timeout lives on Rust side. It'll be easily shared across SDKs in different languages, and the in-engine timeout check will also be far more efficient. What do you think?

I agree with that direction. Ideally Rust core owns the deadline state, deadline comparison, and engine checkpoints; SDKs only provide ergonomic scope APIs and pass an opaque deadline handle into core calls.

python/cocoindex/ops/entity_resolution/llm_resolver.py
```diff
@@ -118,6 +121,19 @@ async def __call__(
             )
             messages.append({"role": "assistant", "content": result.model_dump_json()})
             messages.append({"role": "user", "content": feedback})
+            return None
+
+        # With a CocoIndex deadline, wall-clock budget is the retry budget.
+        # Without one, preserve the historical fixed retry cap.
```
georgeh0: This will lead to surprising outcome: when removing deadline, we may get less retry duration.

python/cocoindex/ops/entity_resolution/llm_resolver.py
```diff
@@ -118,6 +121,19 @@ async def __call__(
...
+        if _deadline.has_deadline():
+            return await _deadline.retry_until_deadline(
+                _attempt, backoff_seconds=_RETRY_BACKOFF_SECONDS
+            )
+
+        for _ in range(1 + self._retries):
+            decision = await _attempt()
+            if decision is not None:
+                return decision
```
georgeh0: "Retry considering both timeout and retry times limit (optional)" is a common routine and we'll reuse across several places. e.g. LiteLLM embedder is another place.

rust/core/src/engine/component.rs
```diff
@@ -527,6 +531,7 @@ impl<Prof: EngineProfile> Component<Prof> {
         self,
         parent_ctx: &ComponentProcessorContext<Prof>,
         processor: Prof::ComponentProc,
+        deadline: DeadlineContext,
```
georgeh0, line 547: `parent_ctx` already has its deadline context. Do we still need to pass it explicitly?

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,113 @@
...
+
+@contextlib.contextmanager
+def timeout(duration: timedelta) -> Iterator[None]:
+    """Apply a cooperative timeout deadline to CocoIndex checkpoints."""
+    if not isinstance(duration, timedelta):
+        raise TypeError("timeout() requires a datetime.timedelta")
+
+    token = _current_deadline.set(
+        _current_deadline.get().with_timeout(duration.total_seconds())
+    )
```
georgeh0, line 43: I think this doesn't affect the deadline on Rust side (which is part of the `ComponentProcessorContext`).

It's fine that during next `use_mount(...)` call the Rust side will get the deadline we pass down explicitly, but it means we need to explicitly pass it to engine on every call that needs to do a timeout check on Rust side.  The real concern here is dual-source-of-truth (`_current_deadline` and `ComponentProcessorContext`), and there'll be confusing issues down the road. Ideally we should have a single source of truth.

docs/src/content/docs/advanced_topics/timeouts.mdx
```diff
@@ -0,0 +1,197 @@
...
+from datetime import timedelta
+
+import cocoindex as coco
+
+
+try:
+    with coco.timeout(timedelta(minutes=5)):
+        app.update_blocking()
+except coco.DeadlineExceededError:
+    print("CocoIndex update timed out")
```
georgeh0: I think the majority use case we want to support is to enforce the deadline within a processing component, e.g. timeout to process a single file, timeout to make a single LLM call. Since it's more feasible and the actual time won't grow linearly with the number of entries in the source.

Also many users don't explicitly call `await app.update()` / `app.update_blocking()` (they use the CLI).

tomz-alt: make sense, let me figure out a better example

python/cocoindex/ops/entity_resolution/llm_resolver.py
tomz-alt: I see, let me make it some generic retry helper.

python/cocoindex/_internal/deadline.py
tomz-alt: Agreed on the dual-source concern. 
How about we design it this way:
- `ComponentProcessorContext.deadline` is the single source for all
  engine-internal checkpoints (memo, entry, pre-submit). Narrowing inside
  the body intentionally doesn't affect these.
- Narrowed scopes (`with coco.timeout(...)`) stay in the per-task
  ContextVar, since the shared ctx can't hold per-task values under
  concurrent coco.map().
- The boundary between the two becomes one explicit contract: a single
  helper `deadline_for_engine()` is the only derivation path, and every
  engine entry point that checks a deadline takes it as a required
  parameter (no default) — next_id, use_mount, update. Isolated paths
  (mount, mount_each, live) take no parameter at all.

to be concrete:
```python
core.next_id(key, deadline)        # has a required `deadline` param
                                   #   → this call CHECKS the deadline

core.mount_async(processor, path)  # has NO deadline param at all
                                   #   → this call is ISOLATED, by design
```
So a forgotten hand-off becomes an immediate TypeError rather than a
silently stale check.

rust/core/src/engine/component.rs
tomz-alt: Yes, they can hold different values. parent_ctx.deadline is the parent's own deadline, frozen when the parent was mounted. The parameter is the deadline for the child, taken from the caller's current `with coco.timeout(...)` scope. 
Concrete case:

```python
@coco.fn
async def app_main() -> None:          # parent_ctx.deadline = NONE
    async def run_one(spec):
        label, secs = spec
        with coco.timeout(timedelta(seconds=secs)):
            await coco.use_mount(coco.component_subpath(label), report, label)

    # two concurrent tasks, one shared parent_ctx
    await coco.map(run_one, [("fast", 5), ("slow", 30)])
```
On deriving an intermediate copy instead (child_ctx = parent_ctx.with_timeout(5s), Go style): the ctx inner is documented"never forked" (context.rs), because it holds the component identity plus
live state such as the build permit and the logic_deps accumulator that
feeds memoization. 
The deadline field lives in that same inner, so a deadline-modified copy would fork memoization state: fingerprints recorded via the copy would be missing from the memo entry, and later edits to those functions would not invalidate it.

python/cocoindex/ops/entity_resolution/llm_resolver.py
tomz-alt: I see, it is a real slop, I will fix it in the next patch

python/cocoindex/_internal/deadline.py
georgeh0: Since many engine functions already have access to `ComponentProcessorContext`. `next_id` is one example. If they also takes an explicit deadline as argument, these function implementations will have access to both. There're still ambiguity here.

My thoughts:
- Actually we need the deadline in context even outside the scope of a component. e.g. in a batched function or a function scheduled to a runner (e.g. GPU), there's no `ComponentProcessorContext` but we need the deadline/timeout APIs working there too (e.g. users should easily specify a deadline for an outgoing RPC within a batched function). So now I'm leaning towards treating this new context itself as the source of truth, rather than a field in `ComponentProcessorContext`.

- Now how does engine check the deadline? Two options:
  - Option 1: Still add a field to `ComponentProcessorContext` but make sure it always reflects the current deadline context. i.e. **whenever a new deadline context is created within `with coco.timeout(...)` we should create a new instance of `ComponentProcessorContext` to override the current one**: we have a copy but always make sure the copy reflect the actual value.
  - Option 2: Explicitly pass down everywhere. **We don't need a field in `ComponentProcessorContext` for it at all**. Since we don't update it when there's a timeout, if we have a field there, its only legitimate use case is to check timeout on boundary (e.g. before/after each phase of the component), but since these logic is already concentrated, explicitly passing is usually more clear.

Personally I think Option 2 may be cleaner.

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,236 @@
...
+    try:
+        yield
+    finally:
+        _current_deadline.reset(token)
+
+
+def check_deadline() -> None:
+    """Raise DeadlineExceededError if the current deadline has passed."""
+    _current_deadline.get().check()
+
```
georgeh0: I'm thinking making the name of this API a little bit broader, e.g. `check_cancellation()`.

"deadline exceed" is a source of cancellation. there will also be other sources of cancellations later.

e.g. (in an ideal picture later) when 10 different sites call a batched function and result in 1 batched call. Each call site may have a different deadline, and each deadline exceed triggers a cancellation, and cancellation of all will trigger cancellation for the batched function call, and a `check_cancellation()` will terminate the current stack. (BTW, for the batcher, "all caller cancelled -> batched call cancelled" is there on Rust side, but there're missing integrations across Rust and Python)

We don't need to support the generic cancellation in this PR, but we can make the APIs more future proof.

rust/core/src/engine/component.rs
georgeh0: About "never forked" - it's for `ComponentProcessorContextInner`, and actually `ComponentProcessorContext` can be forked (e.g. we're forking it for a new `processing_stats`).

But actually we don't have to add a field to `ComponentProcessorContext` at all (see discussion in other comments).

python/cocoindex/_internal/deadline.py
tomz-alt: I prefer opion 2 as well, explicit single source of truth always preferable

rust/core/src/engine/component.rs
tomz-alt: agreed

python/cocoindex/_internal/deadline.py
tomz-alt: make sense, we can go for it.

badmonster0:
@tomz-alt let us know when it is ready for another look

tomz-alt:
> @tomz-alt let us know when it is ready for another look

yes it is ready

rust/core/src/engine/deadline.rs
```diff
@@ -0,0 +1,198 @@
...
+
+static MONOTONIC_ANCHOR: OnceLock<Instant> = OnceLock::new();
+static TEST_OFFSET_NS: AtomicU64 = AtomicU64::new(0);
+static TEST_CLOCK_ENABLED: AtomicBool = AtomicBool::new(false);
+#[cfg(test)]
+static TEST_CLOCK_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+/// Immutable deadline handle carried through engine contexts.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct DeadlineContext(u64);
```
georgeh0: Using `Instant` here is cleaner and safer.

We should almost always use types like `Instant` and `Duration` instead of raw `u64`. They carry more concrete semantics.

rust/core/src/engine/deadline.rs
```diff
@@ -0,0 +1,198 @@
...
+
+const _: () = assert!(std::mem::size_of::<DeadlineContext>() == 8);
```
georgeh0: It looks weird (already implied by the code 2 lines above) and has dependency on an implementation detail.

rust/core/src/engine/component.rs
```diff
@@ -474,6 +478,13 @@ impl<Prof: EngineProfile> ComponentMountRunHandle<Prof> {
                 Ok(())
             })?;
         }
+        // Caller-attributed post-wait check: only when there IS a waiting
+        // parent (use_mount). The root update passes parent_context=None and
+        // its post-result observation point is AppOpHandle::result(), so the
+        // update task itself must not record the deadline as its own failure.
+        if parent_context.is_some() {
+            self.deadline.check()?;
+        }
```
georgeh0: When `parent_context` is `None`, do we ever need a non-`NONE` `self.deadline`? If not, we can simply always pass a `NONE` value for `self.deadline`. The field will always be honored, which is clearer.

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,242 @@
...
+    if isinstance(retry_on, tuple):
+        return isinstance(error, retry_on)
+    return retry_on(error)
+
+
+async def retry_transient(
+    fn: Callable[[], Awaitable[_RetryResultT]],
+    *,
+    retry_on: RetryOn,
+    max_attempts: int | None = 4,
```
georgeh0: I think a non-surprising default should be `None`.

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,242 @@
...
+      the caller): raises ``DeadlineExceededError``. It fires only if the
+      user actually set one, and can only stop retries sooner.
+
+    No attempt starts past any wall, no result is accepted past the ambient
+    deadline (checked after each attempt completes), and backoff sleeps
+    never exceed the tightest remaining wall. ``bound_attempt=True``
+    additionally cancels an in-flight attempt at the remaining POLICY budget
+    via ``asyncio.wait_for`` (local policy over the caller's own coroutine).
+    The ambient deadline is deliberately never used to cancel an in-flight
+    attempt: CocoIndex timeouts are cooperative, so an attempt always runs
```
georgeh0: I feel this contract should be simpler. In my mind, a `budget` here can be a simple syntax sugar of a `with coco.timeout(..): ...`, if we want to provide that (and using the name `timeout` for the argument is clearer). Introducing a parallel concept "budget" here with slightly different behavior from the deadline system is unnecessary.

Note that "CocoIndex timeouts are cooperative" is not a hard requirement, and it's totally fine to cancel a coroutine using it if it's safe to do so. In my mind, the real requirement is: enforce of CocoIndex deadline (and cancellation - to be supported in the future) is on best-effort basis, i.e. there's no obligation to raise exception at once when deadline exceeds. But it's perfectly fine if for some circumstances we can enforce it more tightly.

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,242 @@
...
+    return _current_deadline.get().has_deadline()
+
+
+def exponential_backoff(
+    initial: float = 1.0,
+    multiplier: float = 2.0,
+    max_delay: float = 30.0,
+) -> Callable[[int], float]:
+    """Return a backoff strategy: attempt index (0-based) to delay seconds."""
+    return lambda attempt: min(initial * multiplier**attempt, max_delay)
```
georgeh0: Here we're creating a closure without leveraging the real benefit of a closure.

e.g. we can keep a `nonlocal` state for the last wait time, so we won't need an expensive `**` each time.

BTW, we can also add a little bit of random factor (jitter).

python/cocoindex/_internal/deadline.py
```diff
@@ -0,0 +1,242 @@
...
+from contextvars import ContextVar
+from datetime import timedelta
+from typing import TypeAlias, TypeVar
+
+from . import core
+
+
+DeadlineExceededError = core.DeadlineExceededError
+
+DeadlineSnapshot = core.DeadlineContext
```
georgeh0: I feel we don't need to give another name (`DeadlineSnapshot`) to `DeadlineContext`. Just calling it the deadline context everywhere is clearer.

tomz-alt: removed the aliase

python/cocoindex/_internal/deadline.py
tomz-alt: Done. The closure now keeps the last delay as nonlocal state and advances it with one multiply per call, so there is no ** per attempt. 
 Also added a jitter parameter that scales each delay by random.uniform(1 - jitter, 1 + jitter); it defaults to 0 so existing schedules stay exact, and call sites can opt in.

python/cocoindex/_internal/deadline.py
tomz-alt: Done. budget is removed; 
 `retry_transient(timeout=...)` is now literally sugar for running the loop inside with coco.timeout(...), so it min-nests with any ambient deadline and there is one time concept. 
 
 Exhaustion raises `DeadlineExceededError` (this changes the LiteLLM 10-minute limit from re-raising the last transient error to raising `DeadlineExceededError` — flagging as a behavior change). 

With cooperative not being a hard requirement, `bound_attempt=True` now cancels an in-flight attempt at the remaining effective deadline via `wait_for` and translates that to `DeadlineExceededError`; 
 
 a `TimeoutError` raised by the operation itself before the deadline is not rewritten. Docs updated to say enforcement is best-effort with tighter enforcement where safe.

python/cocoindex/_internal/deadline.py
tomz-alt: fixed

rust/core/src/engine/deadline.rs
tomz-alt: agreed and changed

rust/core/src/engine/component.rs
tomz-alt: Right, it never needs one: the root's post-wait check belongs to AppOpHandle::result, which already holds the update deadline. 
 The root now stores NONE, the field is renamed caller_deadline, and result() checks it unconditionally with no branch.

rust/core/src/engine/deadline.rs
tomz-alt: Agreed, this is redundant and too tied to the current implementation. , already removed

docs/src/content/docs/advanced_topics/timeouts.mdx
```diff
@@ -0,0 +1,218 @@
...
+If the parent times out while waiting for `handle.ready()`, the timeout belongs to the parent wait. It does not retroactively cancel the child processor.
+
+## What happens on timeout
+
+| When the deadline is reached | Observable behavior |
+| --- | --- |
+| Before a processor starts | The processor body is not run. |
+| During processor code | The processor raises `DeadlineExceededError` at the next check. |
+| After target states were declared in memory | Those declarations are discarded; no connector sink actions are applied for that failed processor run. |
+| Before returning a memoized result | The cached value is not returned after the deadline. |
```
georgeh0: "Before returning a memoized result" - this looks like an implementation detail.

I think many items here are more detailed than what users need to care.

IMO the contract is:
- Any function *may* raise `DeadlineExceededError` if deadline exceeds before it returns.
- Certain CocoIndex APIs guarantees `DeadlineExceededError` will be raised if deadline already exceeds when the function is called.

Users shouldn't rely on the exact time point when `DeadlineExceededError` is actually raised.

georgeh0 (commented):
Thanks! Most parts look good! Some comments on documentation.

docs/src/content/docs/advanced_topics/timeouts.mdx
tomz-alt: Rewrote the section to state exactly that contract

badmonster0:
let us know when it is ready for another look!

tomz-alt:
> let us know when it is ready for another look!

requested review already

georgeh0 (approved):
Thanks a lot!

badmonster0:
@tomz-alt ty for the PR!

PR #2256, opened 2026-07-05
feat(memo): add stable memo type identifiers

### LINKED ISSUE #2229: [FEATURE] customized type identifier to replace full-qualified module path + type name
Currently, when computing memo key for an object based on a class type, we incorporate the full-qualified module path + type name as the memo key. This introduces difficulty for migration, e.g. if the type is moved to another module after refactor, existing memoizations will be invalidated.

We want to allow users to provide a stable distinct type identifier, so they can remain stable upon migration.

---

MaxRong:
Closes: #2229

## Summary

This PR adds opt-in **stable type IDs** for memo fingerprints, allowing a semantic Python type to keep the same memo namespace across a module move or class rename.

By default, CocoIndex keeps the existing behavior: type-aware fingerprints include the Python type's module and qualified name.

## Public API

Types controlled by the application can define a stable ID directly:

```python

@dataclass
class SourceEntry:
    __coco_memo_type_id__ = "com.example.SourceEntry/v1"

    name: str
    version: int
```

Types that cannot be modified use the existing registration API:

```python
from pathlib import Path

import cocoindex as coco

coco.register_memo_key_function(
    Path,
    lambda path: str(path.resolve()),
    stable_type_id="stdlib.pathlib.Path/v1",
)
```

A stable type ID can also be registered without a key function when CocoIndex already knows how to fingerprint the exact type:

```python
coco.register_memo_key_function(
    SourceEntry,
    stable_type_id="com.example.SourceEntry/v1",
)
```

`register_memo_key_function(...)` is the single public registration API. It supports:

- key function only;
- key plus state function;
- stable type ID only;
- key/state functions plus stable type ID.

Each call replaces the exact type's complete registered key/state/stable-ID configuration. The previously proposed `register_memo_type_identifier(...)` API is not exposed.

## Class objects

A class object means the class value itself—for example, `ProductRow`—rather than an instance such as `ProductRow(...)`.

Class values honor explicit key/state registrations on their custom metaclass or `type`. Lookup stops before `object`. If there is no matching registration, CocoIndex uses the class's stable type ID or its module and qualified name.

## Migration

Stable type IDs do not rewrite existing memo entries.

To preserve reuse across a move or rename:

1. Add the stable type ID before the refactor.
2. Execute affected memoized calls to seed the new namespace.
3. Move or rename the class while retaining the same stable ID.

A cached parent does not execute nested memoized calls, so affected parents may need to be invalidated or reprocessed during seeding.

## Breaking changes

Raw class values no longer call `__coco_memo_key__()` or `__coco_memo_state__()` attributes found on the class or its metaclass, including callable `@classmethod` or `@staticmethod` forms.

Applications intentionally relying on that behavior should migrate to:

- per-function `memo_key=` for local behavior;
- registration on a dedicated custom metaclass for one class family;
- registration on `type` for intentional process-wide class behavior.

Existing instance memo-key and memo-state methods are unaffected.

## Scope / non-goals

This PR intentionally does not:

- migrate memo entries created before a stable type ID was added;
- make arbitrary pickle-fallback objects refactor-stable without `__coco_memo_key__` or a registered memo-key function;
- fingerprint class schema or mutable class-level state automatically;
- allow a registration on `object` to customize raw class values;
- change function logic identity, which still uses the function's module and qualified name.

## Testing
- [x] uv run pytest python/`
 skipped
- [x] `uv run mypy`
 - [x] `cd docs && npm ci && npm run build`

### DIFF
--- a/docs/src/content/docs/advanced_topics/memoization_keys.mdx
+++ b/docs/src/content/docs/advanced_topics/memoization_keys.mdx
@@ -2,22 +2,36 @@
 title: "*Memoization keys* & states"
 description: >
   Customize how CocoIndex fingerprints memoized inputs via __coco_memo_key__,
-  registered functions, or stable semantic type identifiers, and layer state
-  validation (e.g. mtime, then content hash) on top.
+  registered memo key functions, or stable type IDs, and layer state validation
+  (e.g. mtime, then content hash) on top.
 ---
 As described in [Function — Change detection](../programming_guide/function#change-detection), CocoIndex detects [logic, input, and context changes](../programming_guide/function#change-detection) to decide whether a memo can be reused. Function arguments, [`deps`](../programming_guide/function#deps) values, and [context values](../programming_guide/context#change-detection) with `detect_change=True` are all fingerprinted through the same **data fingerprinting** pipeline. By default, most types are fingerprinted automatically. This page covers how to customize that pipeline — how objects are fingerprinted and validated:
 
 - **Memoization keys** — how to control what CocoIndex uses as the fingerprint for your objects.
 - **Memo states** — how to add post-fingerprint validation to check freshness beyond simple equality.
-- **Stable type identifiers** — how to keep a type's memo namespace stable across refactors.
+- **Stable type IDs** — how to keep a type's memo namespace stable across refactors.
+
+:::tip[Choose the smallest mechanism]
+Most values need no customization. CocoIndex already handles primitives, containers, dataclasses, Pydantic models, class objects, and other picklable objects.
+
+- Only one function should transform or ignore an argument → use [`memo_key=`](#override-at-the-call-site-with-memo_key).
+- You control the type and every use should share the same key → define [`__coco_memo_key__`](#define-__coco_memo_key__-when-you-control-the-type).
+- You don't control the type and every use should share the same key → [register a key function](#register-a-key-function-when-you-dont-control-the-type).
+- A matching key still needs a freshness check, such as mtime or ETag → use a [memo state](#memo-state-validation).
+- A class may move or be renamed while keeping the same memo namespace → use a [stable type ID](#use-stable-type-ids-across-refactors).
+- A value must never participate in memoization → inherit from [`NotMemoKeyable`](#preventing-memoization).
+:::
 
 ## How data fingerprinting works
 
-For each data value (function argument, `deps` value, or context value), CocoIndex derives a canonical form with this precedence:
+Timing depends on where the value comes from. Function arguments are fingerprinted when the function is called. Change-detected context values are fingerprinted when they are provided (for example, by `builder.provide()`), before `use_context()` reads them. `deps` values are fingerprinted once when the `@coco.fn(..., deps=...)` or `@coco.fn.as_async(..., deps=...)` decorator is applied.
 
-1. If the object implements **`__coco_memo_key__()`**, CocoIndex uses its return value.
-2. Otherwise, if you registered a **memo key function** for the object's type, CocoIndex uses that.
-3. Otherwise, CocoIndex falls back to structural canonicalization for a limited set of primitives/containers.
+For each data value, CocoIndex derives a canonical form with this precedence:
+
+1. **Class objects** (`type`) — the class values themselves, such as `ProductRow`, not instances such as `ProductRow(...)`. CocoIndex first checks for an explicit memo key registration on the class object's metaclass or its bases, including `type`. If none exists, it uses the built-in class-object identity path. It never calls memo-key or memo-state attributes on the class object itself. See the class-object notes below for what is and isn't included.
+2. For non-class objects, if the object implements **`__coco_memo_key__()`**, CocoIndex uses its return value.
+3. Otherwise, if you registered a **memo key function** for the object's type, CocoIndex uses that.
+4. Otherwise, CocoIndex uses built-in canonicalization for supported automatic types, including primitives, containers, dataclasses, Pydantic models, and pickle fallback.
 
 The following types are handled automatically (no custom key needed):
 
@@ -25,7 +39,7 @@ The following types are handled automatically (no custom key needed):
 - **Containers**: `list`, `tuple`, `dict`, `set`, `frozenset` (recursively canonicalized)
 - **Dataclass instances**: all fields included in definition order
 - **Pydantic v2 models**: all fields included
-- **Class objects** (`type`): identified by semantic type ID when configured, otherwise by module and qualified name
+- **Class objects** (`type`): identified by a stable type ID when configured, otherwise by module and qualified name
 - **Other picklable objects**: used as a fallback via `pickle`
 
 The canonical forms are combined into a deterministic fingerprint. If the fingerprint matches a cached entry, the cached result is reused — unless **memo states** indicate it's stale (see [Memo state validation](#memo-state-validation) below).
@@ -60,9 +74,22 @@ class UserRow:
         return ("users", self.user_id, self.updated_at)
 ```
 
+The method is used whenever a `UserRow` is fingerprinted as a function argument, `deps` value, or change-detected context value. For example:
+
+```python
+import cocoindex as coco
+
+@coco.fn(memo=True)
+def format_user(row: UserRow) -> str:
+    print("format_user executed")
+    return f"{row.user_id}: updated at {row.updated_at}"
+```
+
+Inside a processing component, the first update executes `format_user`. When logic and context also remain unchanged, a later update with another `UserRow` containing the same `user_id` and `updated_at` reuses the cached result. Changing either value changes the memo key, so the function executes again.
+
 ### Register a key function (when you don't control the type)
 
-If you can't add `__coco_memo_key__` (stdlib / third-party types), register a handler:
+If you can't add `__coco_memo_key__` (stdlib / third-party types), register a key function:
 
 ```python
 from pathlib import Path
@@ -79,12 +106,13 @@ register_memo_key_function(Path, path_key)
 - Registration is **MRO-aware**: if you register both a base class and a subclass, the **most specific** match wins.
 - Your key function must return the same kinds of stable objects as `__coco_memo_key__` (small primitives/tuples).
 
+### Use stable type IDs across refactors
 
-### Stabilize type identity across refactors
+This is an advanced migration feature. If you are not preserving existing memo entries across a class move or rename, skip this section and continue with [`memo_key=`](#override-at-the-call-site-with-memo_key) or [memo state validation](#memo-state-validation).
 
-By default, CocoIndex includes a type's module and qualified name around type-aware values such as dataclasses, Pydantic models, values with `__coco_memo_key__`, registered memo-key functions, and class objects. If a class moves or is renamed, that default identity changes and old memo entries are not reused.
+By default, CocoIndex includes a type namespace (module plus qualified name) for dataclass instances, Pydantic model instances, objects handled by `__coco_memo_key__` or a registered memo key function, and class objects passed as values. If a class moves or is renamed, that default type namespace changes and old memo entries are not reused.
 
-Add a stable semantic identifier when the type should keep the same memo namespace across refactors:
+A stable type ID replaces that type-name portion when the type should keep the same memo namespace across refactors:
 
 ```python
 from dataclasses import dataclass
@@ -98,37 +126,184 @@ class SourceEntry:
     content: str
 ```
 
-For a third-party type that is already handled by `register_memo_key_function(...)`, register the type ID on that same owner type:
+To preserve reuse **after** a refactor, add the stable type ID before moving or renaming the class and ensure every affected memoized call executes under that ID. This seeding execution is cold for affected values: the stable type ID replaces the old module-plus-qualified-name namespace, so CocoIndex creates new memo entries instead of reusing the old ones.
+
+An application update seeds only memoized calls that actually execute. A cached parent does not run its body or nested `@coco.fn` calls, so invalidate or reprocess affected parents—or otherwise invoke the affected nested calls—before the refactor.
+
+After seeding, move or rename the class while keeping the same stable type ID so later updates can reuse the entries created under that ID. Stable type IDs preserve an input type's identity across refactors; normal function logic tracking still invalidates memo entries when the decorated function's source changes. Adding the stable type ID only after the move does not reconstruct entries created under the old module plus qualified name.
+
+If you control the class definition, set `__coco_memo_type_id__` on the class. If you cannot or do not want to edit the class definition, register the same stable type ID with `register_memo_key_function`.
+
+Use `stable_type_id` without a key function only when CocoIndex already fingerprints the exact runtime type: dataclass instances, Pydantic model instances, instances with `__coco_memo_key__()`, or the class object itself. Subclasses do not inherit this form; give the subclass its own stable type ID unless a registered key function handles it through MRO lookup.
+
+Rule of thumb: if CocoIndex would otherwise use pickle for your object, `stable_type_id` by itself is not enough; register a key function and stable type ID together.
+
+Omit the key function argument entirely when registering only a stable type ID; do not pass `None` as the second argument.
 
 ```python
-from pathlib import Path
+from dataclasses import dataclass
 import cocoindex as coco
 
-coco.register_memo_key_function(Path, lambda path: str(path))
-coco.register_memo_type_identifier(Path, "stdlib.pathlib.Path/v1")
+@dataclass
+class ProductRow:
+    sku: str
+    updated_at: int
+
+PRODUCT_ROW_TYPE_ID = "com.example.ProductRow/v1"
+
+coco.register_memo_key_function(
+    ProductRow,
+    stable_type_id=PRODUCT_ROW_TYPE_ID,
+)
 ```
 
-Direct stable type IDs are exact-type: `__coco_memo_type_id__` and direct `register_memo_type_identifier(SomeType, ...)` lookup apply only to that exact type. For values handled by `register_memo_key_function(...)`, the selected registered owner type supplies the type identity for subclasses, matching the MRO-aware key-function lookup described above.
+This registration is process-global. Register before the value is fingerprinted: for change-detected context values, before `builder.provide()`; for function arguments, before memoized calls that should use it; and for `deps`, before the `@coco.fn(..., deps=...)` or `@coco.fn.as_async(..., deps=...)` decorator executes. Later calls replace the previous registration for that exact type.
 
-Stable type IDs do not retroactively migrate memo entries created before the ID existed, and they do not make plain pickle-fallback instances refactor-stable; use `__coco_memo_key__` or `register_memo_key_function(...)` when the object's value also needs a stable custom key.
+This registration has two effects:
 
-Class objects have one extra boundary: if no explicit class-object hook is selected, the registered `type` memo-key path can use the class object's stable type ID. An explicit zero-argument `__coco_memo_key__` hook on a class object is a custom key, so CocoIndex does not automatically wrap that hook payload in the class object's stable ID; include any stable namespace in the hook return value when needed.
+- `ProductRow(...)` instances keep their normal dataclass field fingerprint, but the surrounding type namespace uses `PRODUCT_ROW_TYPE_ID`.
+- The class object `ProductRow` also uses `PRODUCT_ROW_TYPE_ID` when you pass the class itself (`ProductRow`), not an instance (`ProductRow(...)`).
+
+If the memoized function takes the class itself and reads class-level details, a stable type ID alone is not enough. Examples include field names, field types/defaults, validators, Pydantic config, class attributes, or version constants. Because `memo_key=` replaces the normal fingerprint for `row_type` in the example below, include both the stable type ID and the class-level details the function reads:
+
+```python
+# Continuing the ProductRow example above.
+from dataclasses import fields
+import cocoindex as coco
+
+@coco.fn(
+    memo=True,
+    memo_key={
+        "row_type": lambda row_type: (
+            PRODUCT_ROW_TYPE_ID,
+            tuple(field.name for field in fields(row_type)),
+        )
+    },
+)
+def product_row_fields(row_type: type[ProductRow]) -> tuple[str, ...]:
+    return tuple(field.name for field in fields(row_type))
+
+product_row_fields(ProductRow)
+```
+
+:::::caution[Class object schema is not fingerprinted automatically]
+A stable type ID says the class should keep the same memo namespace after a move or rename. It does not say the class schema is unchanged. If cached results depend on class-level details, include those details with `memo_key=` for that function. If a class change should invalidate every memo that uses this stable type ID, change the stable type ID, for example from `/v1` to `/v2`.
+:::::
+
+Use stable type IDs directly for class objects only when the memoized computation does not depend on class details such as module, qualified name, display name, schema, or state. Those details are intentionally excluded from the class object's fingerprint when a stable type ID is used.
+
+For a third-party type that would otherwise fall back to pickle, register both a custom key function and a stable type ID on the same owner type:
+
+```python
+from some_library import ExternalRecord
+import cocoindex as coco
+
+coco.register_memo_key_function(
+    ExternalRecord,
+    lambda record: (record.id, record.version),
+    stable_type_id="com.example.ExternalRecord/v1",
+)
+```
+
+Registration is replacement-based, not additive. Each call replaces the full registration for that exact type:
+
+| Call shape | Registered key/state functions after the call | Registered stable type ID after the call |
+| --- | --- | --- |
+| `register_memo_key_function(T, key_fn, state_fn=..., stable_type_id="...")` — see [`state_fn` requirements](#register-a-state-function-when-you-dont-control-the-type) | `key_fn` and `state_fn` are installed. | The provided stable type ID is installed. |
+| `register_memo_key_function(T, key_fn)` | `key_fn` is installed and any previous `state_fn` is cleared unless a new `state_fn` is supplied. | Any previous registered stable type ID is cleared. |
+| `register_memo_key_function(T, stable_type_id="...")` | Any previous key/state functions are cleared. | The provided stable type ID is installed. |
+
+A `__coco_memo_type_id__` defined on the class is separate from the process-global registration table. If both are set for the same exact type, the registered stable type ID takes precedence. Provide `stable_type_id=` in a registration only when you intentionally want that override; otherwise, register only the key/state function and use the class-defined ID.
+
+For a registry-owned stable type ID, if a type needs a key function, state function, and stable type ID, register all of them in the same call. Avoid splitting registration across helper modules where import order can determine which registration wins.
+
+If a type already has a key or state function and you want to add a stable type ID, update the same call:
+
+```python
+coco.register_memo_key_function(
+    T,
+    key_fn,
+    state_fn=state_fn,
+    stable_type_id="com.example.T/v1",
+)
+```
+
+Do not add the stable type ID with a separate later call; that later call replaces the registration and clears the previous key/state functions.
+
+#### Stable type ID without a key function: exact type only
+
+`__coco_memo_type_id__` and `register_memo_key_function(SomeType, stable_type_id=...)` without a key function apply only to exactly `SomeType`, not subclasses. For example:
+
+```python
+class Parent:
+    ...
+
+class Child(Parent):
+    ...
+
+coco.register_memo_key_function(Parent, stable_type_id="com.example.Parent/v1")
+# Child(...) does not use Parent's stable type ID, because this registration
+# has no key function and applies only to exactly Parent.
+```
+
+#### Stable type ID with a key function: owner selected by MRO
+
+When `register_memo_key_function(Parent, key_fn, stable_type_id=...)` handles a `Child(Parent)` value through MRO lookup, the selected registered owner (`Parent`) supplies the type namespace. That owner namespace can come from the registration's `stable_type_id=...` argument or from `Parent.__coco_memo_type_id__`. For example:
+
+```python
+class Parent:
+    ...
+
+class Child(Parent):
+    ...
+
+coco.register_memo_key_function(
+    Parent,
+    lambda value: value.id,
+    stable_type_id="com.example.Parent/v1",
+)
+# Child(...) can use Parent's stable type ID when Parent's key function wins
+# MRO lookup and handles the Child value.
+```
+
+Stable type IDs do not retroactively migrate memo entries created before the stable type ID existed.
+
+#### Class objects as inputs
+
+Class objects are the class values themselves, such as `ProductRow`, not instances such as `ProductRow(...)`.
+
+When CocoIndex fingerprints a class object:
+
+- It never calls `__coco_memo_key__` or `__coco_memo_state__` attributes on the class object itself.
+- It checks memo key functions explicitly registered on the class object's metaclass or its bases, including `type`. The most specific registered owner wins; a `state_fn` from the same registration participates in state validation, and that owner supplies the stable type ID or module-plus-qualified-name namespace.
+- If no metaclass or `type` key function is registered, it uses the class object's own stable type ID when present, otherwise its module plus qualified name.
+- A key/state registration on `ProductRow` customizes `ProductRow(...)` instances only. To customize the raw `ProductRow` class globally, register a key function on its metaclass. A stable-ID-only registration on a metaclass or on `type` does not customize all class objects.
+- It excludes class schema and class state. If a memoized result depends on fields, field types/defaults, validators, Pydantic config, class attributes, or any other class-level details, pass that data separately with `memo_key=` or change the stable type ID.
+
+Do not add a `@classmethod` or `@staticmethod` named `__coco_memo_key__` to customize class-object memoization. Class objects ignore it, but instances can still see it and accidentally give every instance the same memo key.
 
 ### Override at the call site with `memo_key=`
 
-The two approaches above are **type-level**: every memoized function sees the same fingerprint for a given object. To customize fingerprinting **only for a specific function**, pass `memo_key=` to `@coco.fn` (or `@coco.fn.as_async`). It maps parameter names to either a callable (transform the value before fingerprinting) or `None` (exclude the parameter entirely).
+The approaches above are **type-level**: every memoized function sees the same fingerprint for a given object. To customize fingerprinting **only for a specific function**, pass `memo_key=` to `@coco.fn` (or `@coco.fn.as_async`). It maps parameter names to either a callable (transform the value before fingerprinting) or `None` (exclude the parameter entirely).
 
 ```python
-@coco.fn(memo=True, memo_key={"entry": lambda e: (e.name, e.version), "extra": None})
-def transform(entry: SourceDataEntry, extra: str) -> str:
-    ...
+@coco.fn(
+    memo=True,
+    memo_key={
+        "entry": lambda entry: (entry.name, entry.version),
+        "request_id": None,
+    },
+)
+def transform(entry: SourceDataEntry, request_id: str) -> str:
+    # request_id is request metadata and does not affect the result.
+    return f"{entry.name}@{entry.version}"
 ```
 
 Each entry in `memo_key`:
 
 - **Callable** — applied to the argument; its return value is fingerprinted in place of the original. Semantically the same as `__coco_memo_key__()` on the type, but scoped to this one function. Useful when the type's default fingerprint is correct everywhere *else*, and only this function should treat the argument differently.
 - **`None`** — the parameter is excluded from the memo key. Changing its value never invalidates the cache. Useful for arguments that don't affect the result (logger handles, debug flags, request-scoped context that isn't part of the computation).
-- **Not listed** — the parameter is fingerprinted normally (type-level `__coco_memo_key__`, registered handler, or default canonicalization).
+- **Not listed** — the parameter is fingerprinted normally (type-level `__coco_memo_key__`, registered memo key function, or default canonicalization).
 
 It works for every parameter kind:
 
@@ -139,11 +314,6 @@ It works for every parameter kind:
 
 `memo_key` is validated at decoration time: unknown parameter names raise `ValueError`, and values that are neither callable nor `None` raise `TypeError`.
 
-:::tip[Picking the right tool]
-- You control the type and the fingerprint is the same wherever it's used → **`__coco_memo_key__`**.
-- You don't control the type but want a global handler → **`register_memo_key_function`**.
-- Only *this* function should treat an argument specially (transform or skip it) → **`memo_key=`**.
-:::
 
 ## Memo state validation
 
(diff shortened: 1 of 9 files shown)


docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,6 +79,41 @@ register_memo_key_function(Path, path_key)
...
+
+coco.register_memo_key_function(Path, lambda path: str(path))
+coco.register_memo_type_identifier(Path, "stdlib.pathlib.Path/v1")
+```
+
+Direct stable type IDs are exact-type: `__coco_memo_type_id__` and direct `register_memo_type_identifier(SomeType, ...)` lookup apply only to that exact type. For values handled by `register_memo_key_function(...)`, the selected registered owner type supplies the type identity for subclasses, matching the MRO-aware key-function lookup described above.
+
+Stable type IDs do not retroactively migrate memo entries created before the ID existed, and they do not make plain pickle-fallback instances refactor-stable; use `__coco_memo_key__` or `register_memo_key_function(...)` when the object's value also needs a stable custom key.
+
+Class objects have one extra boundary: if no explicit class-object hook is selected, the registered `type` memo-key path can use the class object's stable type ID. An explicit zero-argument `__coco_memo_key__` hook on a class object is a custom key, so CocoIndex does not automatically wrap that hook payload in the class object's stable ID; include any stable namespace in the hook return value when needed.
```
georgeh0: I failed to understand this paragraph.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,6 +79,41 @@ register_memo_key_function(Path, path_key)
...
+```
+
+For a third-party type that is already handled by `register_memo_key_function(...)`, register the type ID on that same owner type:
+
+```python
+from pathlib import Path
+import cocoindex as coco
+
+coco.register_memo_key_function(Path, lambda path: str(path))
+coco.register_memo_type_identifier(Path, "stdlib.pathlib.Path/v1")
```
georgeh0: An alternative way is to allow to register both in a single function (e.g. users can provide both identifier and key function, and can also only provide one of them).

What do you think about pros and cons of both approaches?

georgeh0 (commented):
Thanks for making this PR!

docs/src/content/docs/advanced_topics/memoization_keys.mdx
MaxRong: I kept type identity and key function separate because it felt more explicit, especially since you can reasonably want one without the other. That said, I agree a unified function seems very reasonable.

If you think one entry point would be more ergonomic, I’m happy to add it. I’d probably keep the explicit APIs available too, and implement a register_memo_type function as a convenience wrapper over them.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
MaxRong: Sorry, I should make it more clear. 

- Stable type IDs only help for memo keys created after the ID exists. They don’t rewrite old memo cache entries. Additionally, If Cocoindex falls back to pickle for a plain object with no memo key hook the stable ID is not used. So if a refactor-stable identity is needed, the user should define one with `__coco_memo_key__` or `register_memo_key_function(...)` unless key function includes namespace.
- And, for class objects, if explicit __coco_memo_key__ is defined that hook overrides default behavior, so it's important for users to include the stable type id in the return.

georgeh0: I was asking for the exact last paragraph. I think I can understand the two paragraphs above.

I still don't really understand the last paragraph, until I read the implementation: I just know what does *class-object* really mean here (I thought it simply meant an object constructed by a class type).

Observations and suggestions:
- When someone says they fails to understand a part of the document, consider if the document needs to be improved.
- Usually using examples will make a document easier to understand.
- This paragraph uses the term "hook" multiple times, but it's never defined or used anywhere else in the doc. So readers easily get lost.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +122,192 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+        raw_fn = getattr(typ, _MEMO_STATE_ATTR)
+        return _make_state_fn_entry(state_hook, raw_fn)
+    return None
+
+
+def _type_identity_parts(typ: type) -> tuple[Fingerprintable, Fingerprintable]:
+    """Return stable-ID or module+qualname type identity parts."""
+    identifier = _lookup_memo_type_identifier(typ)
+    if identifier is not None:
+        return (("__coco_memo_type_id__", identifier), None)
```
georgeh0: curious why not directly return `("__coco_memo_type_id__", identifier)` here

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +122,192 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    is no ``self`` to bind.  Zero-argument functions assigned as class
+    attributes, descriptor-based hooks, callable objects, and metaclass hooks
+    remain valid class-object hooks.
+    """
+    for base in cls.__mro__:
+        raw = base.__dict__.get(_MEMO_KEY_ATTR)
+        if raw is None:
+            continue
+        hook = getattr(cls, _MEMO_KEY_ATTR, None)
+        if isinstance(raw, (classmethod, staticmethod)):
```
georgeh0: I understand the motivation, but feel it's tricky if we expect users to provide a `@staticmethod`/`@classmethod` for `__coco_memo_key__` for class-level memo key extraction: when come to the memo key extraction for an instance of the class, it'll end up with the same value for all instances of the class. It's a footgun.

For the class itself, maybe we should never call `__coco_memo_key__`.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
georgeh0: > it felt more explicit, especially since you can reasonably want one without the other

I think providing one without the other is always possible even under the same API


I prefer a single unified API since they're for the same purpose. Separated APIs + convenient wrapper will broaden the API surface unnecessarily, resulting in choice paralysis and diverged coding styles (e.g. consider if we want to search for registrations for given types, we need to remember searching for both APIs).

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -1,14 +1,15 @@
 ---
 title: "*Memoization keys* & states"
 description: >
-  Customize how CocoIndex fingerprints memoized inputs via __coco_memo_key__ or
-  registered functions, and layer state validation (e.g. mtime, then content hash)
-  on top — plus NotMemoKeyable to opt out for stateful types.
+  Customize how CocoIndex fingerprints memoized inputs via __coco_memo_key__,
+  registered functions, or stable semantic type identifiers, and layer state
```
badmonster0: what does this mean? is it defined below?

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -1,14 +1,15 @@
...
+  validation (e.g. mtime, then content hash) on top.
 ---
 As described in [Function — Change detection](../programming_guide/function#change-detection), CocoIndex detects [logic, input, and context changes](../programming_guide/function#change-detection) to decide whether a memo can be reused. Function arguments, [`deps`](../programming_guide/function#deps) values, and [context values](../programming_guide/context#change-detection) with `detect_change=True` are all fingerprinted through the same **data fingerprinting** pipeline. By default, most types are fingerprinted automatically. This page covers how to customize that pipeline — how objects are fingerprinted and validated:
 
 - **Memoization keys** — how to control what CocoIndex uses as the fingerprint for your objects.
 - **Memo states** — how to add post-fingerprint validation to check freshness beyond simple equality.
+- **Stable type identifiers** — how to keep a type's memo namespace stable across refactors.
```
badmonster0: i see multiple definitions for similar stuff
stable semantic type identifiers,
Stable type identifiers
semantic type ID 

are they the same thing?

python/cocoindex/_internal/memo_fingerprint.py
MaxRong: I wanted to tag the stable ID case explicitly to distinguish from the normal (module, __qualname__). There's also some minor collision risk but a non-issue for average users likely, can simplify if preferred for maintainability

docs/src/content/docs/advanced_topics/memoization_keys.mdx
MaxRong: noted, I'll make this a lot more clear, thank you for the detailed tips!

docs/src/content/docs/advanced_topics/memoization_keys.mdx
MaxRong: Yes, I will standardize to 'stable type IDs' and make documentation a lot more clear.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
MaxRong: understood, I completely agree. Will work in the direction of extending the register_memo_key function in that case.

python/cocoindex/_internal/memo_fingerprint.py
MaxRong: great point, I overlooked this and tried to get too clever. Will definitely simplify and take your suggestion.

MaxRong:
Thanks again for the detailed review. I pushed the follow-up commit:

 ```text
 a2d0a2ae feat(memo): implement reviewer feedback, fix bugs
 ```

and also edited the PR body.

 ### What changed

- **Unified registration:** `register_memo_key_function(...)` now handles
key, state, stable-ID-only, and combined registration. Each call replaces
the exact type's complete registered configuration. The separate
`register_memo_type_identifier(...)` API was removed.
- **Class-object behavior:** class values such as `ProductRow` no longer
invoke memo-key/state attributes found on the class itself. Explicit
custom-metaclass and `type` registrations still work, while lookup stops
before `object`.
- **Canonicalization:** custom memo-key fragments now share their parent's
traversal, preserving parent cycles and aliases. Traversal-local strong
references prevent temporary object-ID reuse from creating false references.
- **Identity handling:** registration is exact by Python type-object
identity, and the registries do not keep registered types alive. Equal or
unhashable metaclasses are covered.
- **Timing and state:** the implementation and docs now distinguish
argument call time, context `builder.provide()` time, and `deps` decoration
time. Updated memo state is persisted when a cached result remains reusable.
- **Documentation:** the guide now defines class objects explicitly,
removes undefined “hook” terminology, explains stable-ID seeding and
replacement, and adds concrete key/state examples.
- **Regression coverage:** added focused tests for class dispatch,
`object` exclusion, registration replacement, cycles, aliases, temporary
 fragments, hostile equality/hash behavior, state timing, and rename reuse.

### Compatibility note

Raw class values no longer call class-level `__coco_memo_key__()` /
`__coco_memo_state__()` methods, including `@classmethod` or `@staticmethod`
forms. Instance methods are unaffected.

Raw-class customization should use per-function `memo_key=`, a dedicated
metaclass registration, or intentional process-wide registration on `type`.

### Verification

- 6 focused class-dispatch/registration regressions passed;
- 2 focused `object`-boundary/temporary-fragment regressions passed;
- documentation build passed with 95 pages;
- documentation checks passed.

I would especially appreciate another look at:

- registration replacement(behaving as expected)?
- revised documentation(migration guidance, clarity).

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -1,30 +1,45 @@
...
+
+:::tip[Choose the smallest mechanism]
+Most values need no customization. CocoIndex already handles primitives, containers, dataclasses, Pydantic models, class objects, and other picklable objects.
+
+- Only one function should transform or ignore an argument → use [`memo_key=`](#override-at-the-call-site-with-memo_key).
+- You control the type and every use should share the same key → define [`__coco_memo_key__`](#define-__coco_memo_key__-when-you-control-the-type).
+- You don't control the type and every use should share the same key → [register a key function](#register-a-key-function-when-you-dont-control-the-type).
+- A matching key still needs a freshness check, such as mtime or ETag → use a [memo state](#memo-state-validation).
+- A class may move or be renamed while keeping the same memo namespace → use a [stable type ID](#use-stable-type-ids-across-refactors).
+- A value must never participate in memoization → inherit from [`NotMemoKeyable`](#preventing-memoization).
```
georgeh0, line 30: Nice summarization!

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -1,30 +1,45 @@
...
 
-1. If the object implements **`__coco_memo_key__()`**, CocoIndex uses its return value.
-2. Otherwise, if you registered a **memo key function** for the object's type, CocoIndex uses that.
-3. Otherwise, CocoIndex falls back to structural canonicalization for a limited set of primitives/containers.
+For each data value, CocoIndex derives a canonical form with this precedence:
+
+1. **Class objects** (`type`) — the class values themselves, such as `ProductRow`, not instances such as `ProductRow(...)`. CocoIndex first checks for an explicit memo key registration on the class object's metaclass or its bases, including `type`. If none exists, it uses the built-in class-object identity path. It never calls memo-key or memo-state attributes on the class object itself. See the class-object notes below for what is and isn't included.
```
georgeh0: I feel this is not a major case, and placing it as a first item (with many words) will make it significantly harder to understand.

I think we can remove it. We only need to add one sentence to "2" (which will become "1") to clarify, e.g. something like:

```markdown
1. If the object implements **`__coco_memo_key__()`**, CocoIndex uses its return value. This applies to class instances, but not class objects (`type`).
```

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,21 +106,204 @@ register_memo_key_function(Path, path_key)
...
+class SourceEntry:
+    __coco_memo_type_id__: ClassVar[str] = "com.example.SourceEntry/v1"
+
+    id: str
+    content: str
+```
+
+To preserve reuse **after** a refactor, add the stable type ID before moving or renaming the class and ensure every affected memoized call executes under that ID. This seeding execution is cold for affected values: the stable type ID replaces the old module-plus-qualified-name namespace, so CocoIndex creates new memo entries instead of reusing the old ones.
+
+An application update seeds only memoized calls that actually execute. A cached parent does not run its body or nested `@coco.fn` calls, so invalidate or reprocess affected parents—or otherwise invoke the affected nested calls—before the refactor.
```
georgeh0: > `parents—or`

It's really hard to distinguish em dash and a regular dash (especially under monospace fonts). `parents—or` looks like one hyphenated word. Please either avoid em dash, or add spaces before/after the em dash.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,21 +106,204 @@ register_memo_key_function(Path, path_key)
...
+
+@dataclass
+class SourceEntry:
+    __coco_memo_type_id__: ClassVar[str] = "com.example.SourceEntry/v1"
+
+    id: str
+    content: str
+```
+
+To preserve reuse **after** a refactor, add the stable type ID before moving or renaming the class and ensure every affected memoized call executes under that ID. This seeding execution is cold for affected values: the stable type ID replaces the old module-plus-qualified-name namespace, so CocoIndex creates new memo entries instead of reusing the old ones.
```
georgeh0: I feel we can make this section simpler.

The "seeding execution" described in this section seems not really necessary for most cases. Once users start to set or change `__coco_memo_type_id__`, previous memoizations are invalidated any way. So IMO "seeding execution" is not really meaningful.

We have two options here:
* We can suggest users that, if memo stability is really important and you cannot rule out potential refactor in the future, set a `__coco_memo_type_id__`. That's the only way to avoid a memo invalidation in future migration.
* We can also allow users directly carry over the auto-generated type ID we used before when `__coco_memo_type_id__` was not set, e.g. if `foo.Class1` is moved to a another module, users can right something like:

  ```python
  __coco_memo_type_id__ = coco.prev_type_id('foo', 'Class1')
  ```
  
  This is more friendly for migration.


What do you think?

MaxRong: Thanks for the review!

I agree with what you've said, your suggestions would make this process a lot clearer. Let me clear up the comments and implement the prev_type_id function.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -193,46 +412,124 @@ def __coco_memo_key__(self) -> typing.NoReturn:
...
+    stable_type_id: str,
+) -> None: ...
 
-    Resolution is MRO-aware: the most specific registered base type wins.
 
-    If *state_fn* is provided it is stored separately and used for memo state
-    validation (see ``_canonicalize``).
+def register_memo_key_function(
+    typ: type,
+    key_fn: _KeyFn | object = _KEY_FN_UNSET,
```
georgeh0: Why not simply use `None` as unset?

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +146,199 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+
+
+def _lookup_stable_type_id(typ: type) -> str | None:
+    """Resolve a registered or exact ``__coco_memo_type_id__`` stable type ID."""
+    stable_type_id = _registered_stable_type_id(typ)
+    if stable_type_id is not None:
+        return stable_type_id
+    if "__coco_memo_type_id__" in typ.__dict__:
+        return _validate_stable_type_id(
+            typ.__dict__["__coco_memo_type_id__"],
```
georgeh0: `"__coco_memo_type_id__" in typ.__dict__` and `typ.__dict__["__coco_memo_type_id__"]` are doing redundant field loop.

I think we can simply do `getattr(typ, '__coco_memo_type_id__', None)` once and check the result.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -31,12 +32,21 @@
...
+
+
 class _MemoFns(typing.NamedTuple):
     key_fn: _KeyFn
     state_fn: _StateFn | None = None
 
 
-_memo_fns: dict[type, _MemoFns] = {}
+_memo_fns: dict[int, tuple[weakref.ReferenceType[type], _MemoFns]] = {}
+_stable_type_ids: dict[int, tuple[weakref.ReferenceType[type], str]] = {}
```
georgeh0: I think a cleaner way is to simply add a field to `_MemoFns` for the new type ID.  We only need a single dict. These 3 fields are usually needed set and used together. The logic will be more clear since all registered information for 3rd party types are kept together.

We only need to give `_MemoFns` a broader name, e.g. `_MemoTypeRegistry`.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
georgeh0: Looks good. A `prev_type_id` function is also the preferred approach in my mind. Please implement it. Thanks!

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -1,30 +1,45 @@
...
+- You don't control the type and every use should share the same key → [register a key function](#register-a-key-function-when-you-dont-control-the-type).
+- A matching key still needs a freshness check, such as mtime or ETag → use a [memo state](#memo-state-validation).
+- A class may move or be renamed while keeping the same memo namespace → use a [stable type ID](#use-stable-type-ids-across-refactors).
+- A value must never participate in memoization → inherit from [`NotMemoKeyable`](#preventing-memoization).
+:::
 
 ## How data fingerprinting works
 
-For each data value (function argument, `deps` value, or context value), CocoIndex derives a canonical form with this precedence:
+Timing depends on where the value comes from. Function arguments are fingerprinted when the function is called. Change-detected context values are fingerprinted when they are provided (for example, by `builder.provide()`), before `use_context()` reads them. `deps` values are fingerprinted once when the `@coco.fn(..., deps=...)` or `@coco.fn.as_async(..., deps=...)` decorator is applied.
```
georgeh0: This is a nice elaboration for the existing short paragraph.

This clearly answers a question of "when is fingerprinting needed" (rather than "how"), and [the first paragraph](https://github.com/cocoindex-io/cocoindex/blob/5071f199b107135f4cf2bbd0cb0202cfae48da6e/docs/src/content/docs/advanced_topics/memoization_keys.mdx?plain=1#L8) of this doc aims at covering it. Maybe can elaborate in the 1st paragraph there (e.g. split into separate bullets to be clearer), and keep this one concise (since it's for "how").

MaxRong:
Thanks for the round of feedback! I addressed the concerns:
- Replaced the custom omitted-key_fn sentinel with None; explicit None now supports stable-ID-only registration.
- Consolidated __coco_memo_type_id__ resolution into one typ.__dict__.get(..., _MISSING) lookup. I used the exact class dictionary rather than getattr to avoid unintentionally inheriting a parent class’s ID and to distinguish a missing attribute from an invalid explicit None.
- Unified key, state, and stable-ID registrations into one _MemoTypeRegistry record and one identity-keyed weak-reference table.
- Added prev_type_id(module, qualname) for types that have already been memoized under their automatic (module, qualname) identity and are later moved or renamed. Assigning its result to __coco_memo_type_id__, or passing it as stable_type_id, keeps emitting the previous automatic identity so compatible existing memo entries remain addressable without  copying or migrating records. The documentation covers canonical module/qualified-name values, nested classes, __main__, limitations, and collision risks. 
- Moved fingerprint timing details into the guide introduction

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +141,211 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    ref, registry = entry
+    if ref() is typ:
+        return registry
+    _memo_type_registry.pop(type_id, None)
+    return None
+
+
+def _lookup_stable_type_id(typ: type) -> str | None:
+    """Resolve a registered or exact ``__coco_memo_type_id__`` stable type ID."""
+    registry = _registered_memo_type_registry(typ)
```
georgeh0: In the case when the type is in the registry, we need the other fields in the registry too. Ideally we only need to lookup the registry once for the type, in `_canonicalize()`.

BTW, when a type both `__coco__*` members also registered in the registry, the `__coco__*` members should take precedence.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +141,211 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    if registry is not None and registry.stable_type_id is not None:
+        return registry.stable_type_id
+
+    stable_type_id = typ.__dict__.get("__coco_memo_type_id__", _STABLE_TYPE_ID_MISSING)
```
georgeh0: I think we can directly use `None` instead of `_STABLE_TYPE_ID_MISSING`. `None` isn't a valid value for `__coco_memo_type_id__` any way.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +141,211 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+
+    def __getnewargs__(self) -> tuple[str]:
+        return (str(self),)
+
+    def _identity_parts(self) -> tuple[str, str]:
+        module_length_str, separator, payload = self.partition(":")
+        if separator == "":
+            raise ValueError("invalid previous type identity payload")
+        module_length = int(module_length_str)
+        return payload[:module_length], payload[module_length:]
```
georgeh0: This is on the hot path (`_canonicalize()` is called very frequently). We have better ways than computing the parts out of the `str` on the fly. Can we simply store the `module` and `qualname` as new field slots?

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +141,211 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+
+    @property
+    def module(self) -> str:
+        return self._identity_parts()[0]
+
+    @property
+    def qualname(self) -> str:
+        return self._identity_parts()[1]
```
georgeh0: Seems they're not used anywhere.

MaxRong:
Just rebased everything, newest changes on the last commit. Thanks for the new round of feedback! Here is the changes:

- The exact runtime type's registry record is now resolved once in
 `_canonicalize()` and passed through the identity/canonicalization
 path instead of being looked up repeatedly.
   - `__coco_memo_key__` / `__coco_memo_state__` members
 take precedence over registered key/state functions.
   - A non-`None` class-declared `__coco_memo_type_id__` now takes
 precedence over the registry's stable type ID, including when an
 MRO-selected registered key function handles the value.
   - Removed the separate missing-value sentinel
   - `_PreviousTypeId` now stores `(module, qualname)` directly, so
 the hot path no longer reparses its string payload. The unused
 public-looking accessors were removed, while copy/deepcopy/pickle
 round trips remain covered.
   - I kept a single `typ.__dict__.get("__coco_memo_type_id__")`
 lookup rather than `getattr(...)` because stable IDs are
 intentionally exact-type-only and should not be inherited from a
 parent class.
   - Updated the guide to reflect the corrected class-declared-ID
 precedence and added focused regressions for precedence and registry
 lookup counts.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +140,205 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+        )
+    return stable_type_id
+
+
+def _validate_previous_type_id_part(value: object, *, source: str) -> str:
+    """Validate and normalize one previous automatic identity part."""
+    if isinstance(value, str):
+        return _validate_stable_type_id(str.__str__(value), source=source)
+    return _validate_stable_type_id(value, source=source)
+
```
georgeh0: I feel we don't have to have these two separate util functions just to validate arguments `module` and `qualname` for `prev_type_id()`. `module` and `qualname` are already annotated with type hint `str` in `prev_type_id()` and we usually just trust that (otherwise too many things to validate everywhere).

Also IMO we don't have to check against `""` too: there can be many different types of wrong values that don't match the actual previous type, and `""` isn't special from others. The contract of our API is "if the passed-in names match previous one, we keep memoization key stability", but we have no obligation to validate they actually match (since it's impossible).

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +140,205 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    _identity_parts: tuple[str, str]
+
+    def __new__(cls, module: str, qualname: str) -> _PreviousTypeId:
+        marker = super().__new__(cls, f"{len(module)}:{module}{qualname}")
+        object.__setattr__(marker, "_identity_parts", (module, qualname))
+        return marker
+
+    def __setattr__(self, name: str, value: object) -> typing.NoReturn:
+        raise AttributeError(f"{type(self).__name__} is immutable")
+
```
georgeh0, line 152: Since `__setattr__` is blocked, we may also want to consider blocking `__delattr__` to make  it really immutable.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +140,205 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    state_methods: list[StateFnEntry],
+) -> Fingerprintable:
+    """Canonicalize a memo-key fragment within the current root traversal.
+
+    Sharing traversal state preserves cycles through the parent object and keeps
+    temporary fragment objects alive so their IDs cannot be reused during this
+    traversal.
+    """
+
+    return _canonicalize(obj, state, state_methods)
```
georgeh0: `_canonicalize_key_fragment()` is a trivial wrapper around `_canonicalize()`. Seems this layer provides no value. We may eliminate it to simplify.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +140,205 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    module/qualname identity shape used by type-aware canonical forms. The
+    tagged first slot keeps stable type IDs disjoint from ordinary module
+    names; ``None`` fills the qualname slot.
+    """
+    stable_type_id = typ.__dict__.get("__coco_memo_type_id__")
+    if stable_type_id is not None:
+        stable_type_id = _validate_stable_type_id(
+            stable_type_id,
+            source=f"{_memo_type_label(typ)}.__coco_memo_type_id__",
+        )
```
georgeh0: IMO this is another unnecessary validation and it's on hot path. We can't really meaningfully validate that, so we don't have to.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,21 +111,228 @@ register_memo_key_function(Path, path_key)
 - Registration is **MRO-aware**: if you register both a base class and a subclass, the **most specific** match wins.
 - Your key function must return the same kinds of stable objects as `__coco_memo_key__` (small primitives/tuples).
 
-### Override at the call site with `memo_key=`
+### Use stable type IDs across refactors
+
```
georgeh0, line 115: This entire section has 150 lines now, which is too verbose. I think the underlying semantics is deadly simple: if you move/rename your type, assign a stable type ID with the previous name. We don't need a section with 150 lines.

MaxRong:
Hello, I update the PR based on feedback:

- Simplified `prev_type_id()` and `_PreviousTypeId` (added `__delattr__`, put
`_identity_parts` in slots, dropped the extra validation helpers and wrapper).
- Pruned `memoization_keys.mdx` down to ~45 lines focused on the main migration patterns, and
clarified metaclass/`type` precedence.
- Swapped the weakref registry table for a normal `dict[type, _MemoTypeRegistry]` and removed
the old weak-identity tests.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +136,133 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    module/qualname identity shape used by type-aware canonical forms. The
+    tagged first slot keeps stable type IDs disjoint from ordinary module
+    names; ``None`` fills the qualname slot.
+    """
+    stable_type_id = typ.__dict__.get("__coco_memo_type_id__")
+    if stable_type_id is None and registry is not None:
+        stable_type_id = registry.stable_type_id
+
+    if isinstance(stable_type_id, _PreviousTypeId):
+        return stable_type_id._identity_parts
```
georgeh0, line 222: Actually here doing another simple `isinstance(stable_type_id, str)` check is warranted, since we need type check any way, and there's to make the Python type checker validate the type for `__coco_memo_type_id__`.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -193,46 +338,113 @@ def __coco_memo_key__(self) -> typing.NoReturn:
...
+        raise TypeError(
+            "register_memo_key_function() key_fn must be callable, "
+            f"got {type(key_fn).__name__}"
+        )
+    if state_fn is not None and not callable(state_fn):
+        raise TypeError(
+            "register_memo_key_function() state_fn must be callable, "
+            f"got {type(state_fn).__name__}"
+        )
+
```
georgeh0, line 443: Since we verified other arguments here, let's also verify type of `stable_type_id` for consistency.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -118,6 +136,133 @@ def canonical_module_name(obj: typing.Any) -> str:
...
+    registry: _MemoTypeRegistry | None,
+) -> tuple[Fingerprintable, Fingerprintable]:
+    """Return stable type ID or module+qualname type identity parts.
+
+    The stable type ID case still returns two parts to preserve the existing
+    module/qualname identity shape used by type-aware canonical forms. The
+    tagged first slot keeps stable type IDs disjoint from ordinary module
+    names; ``None`` fills the qualname slot.
+    """
+    stable_type_id = typ.__dict__.get("__coco_memo_type_id__")
```
georgeh0: Following up on our earlier exchange (I originally suggested `getattr` just for brevity; you kept `__dict__` to avoid unintentional inheritance): after more thought, let's switch to `getattr`, i.e. stable type IDs **are inherited** by subclasses.

Reasons:

- There are valid use cases for inheritance: a family of classes sharing one base, where the base fully decides how memo keys are computed and subclasses are implementation details (e.g. the same data represented in different storages). Since `__coco_memo_key__` is inherited but the type ID is not, such a family today silently gets fragmented namespaces unless every subclass redeclares the ID.
- It gives one uniform dispatch rule for all `__coco_*` members: normal attribute lookup on the object first (own or inherited — we can't distinguish anyway), registered configuration as fallback, then module + qualname. Today methods inherit but the field doesn't; that asymmetry is hard to teach.
- It matches ordinary Python semantics for a class attribute; the `__dict__`-only lookup is the surprising behavior.
- Docs get simpler: the "exact-type-only, subclasses don't inherit" caveat goes away.

On the aliasing concern that motivated `__dict__` (parent/child instances sharing one namespace): declaring an ID on a base class is a deliberate statement that the family shares a namespace. A subclass that diverges semantically should override the ID: same discipline as overriding `__eq__` when equality semantics change. For dataclass/Pydantic families the field tuples keep payloads distinct anyway. So this is acceptable opt-in behavior; just document it.

Two things to handle together with the switch:

1. `prev_type_id` on a non-final base class: each subclass's automatic identity was its own `(module, qualname)`, so a subclass inheriting the parent's marker would emit the parent's old identity and silently mismatch the subclass's own old entries. Add one sentence to the docs: declare `prev_type_id` on each moved class; like any class attribute, a base-class declaration propagates to subclasses.
2. Precedence: pin down that a declared `__coco_memo_type_id__` (own or inherited) takes precedence over a registered `stable_type_id`, including the subclass's own exact-type registration: keep the uniform "declared beats registered" rule. Please state it in the docs and cover it with a small test.

Let's make stable-ID-only registrations MRO-aware as well, so declared and registered stable type IDs both apply to subclasses (most specific registered type wins; a subclass overrides by declaring or registering its own). Today a registered ID covers subclasses only when the registration also carries a key function, which is a conditional scope rule that's hard to teach, and registration is the only mechanism available for third-party hierarchies, where the shared-family case matters most. Precedence stays mechanism-major, mirroring key extraction: declared ID (normal attribute lookup) → registered ID (MRO) → module + qualname.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,21 +111,85 @@ register_memo_key_function(Path, path_key)
...
+Types using pickle fallback fingerprint raw bytes without type namespacing. Setting or registering a stable type ID alone has no effect on them — register a memo key function together with the stable type ID.
+:::
+
+### Class objects as inputs
+
+A class object is the class value itself (such as `ProductRow`), not an instance (such as `ProductRow(...)`).
+
+When CocoIndex fingerprints a class object:
+
+* It uses the class's `__coco_memo_type_id__` when defined, or falls back to its module and qualified name (`module.QualName`).
```
georgeh0: This is not precise. `register_memo_key_function` also affects this.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -193,46 +338,113 @@ def __coco_memo_key__(self) -> typing.NoReturn:
...
+    key_fn: _KeyFn | None = None,
+    *,
+    state_fn: _StateFn | None = None,
+    stable_type_id: str | None = None,
+) -> None:
+    """Register a memo key function and/or stable type ID for a type.
+
+    Key-function resolution is MRO-aware: the most specific registered base
+    type wins. Stable type IDs registered without a key function apply to the
+    exact type only; stable type IDs registered with a key function identify
```
georgeh0: Since `register_memo_key_function` is now the way to register something that isn't a key function (and clears one), consider renaming to `register_memo_type`. We can keep `register_memo_key_function` taking the original signature as a shortcut / for backward compatibility.

python/cocoindex/_internal/memo_fingerprint.py
```diff
@@ -193,46 +338,113 @@ def __coco_memo_key__(self) -> typing.NoReturn:
...
             "This type maintains internal state that is incompatible with memoization."
         )
 
-    _memo_fns[typ] = _MemoFns(_raise_not_memo_keyable)
+    _register_memo_type_registry(typ, _MemoTypeRegistry(key_fn=_raise_not_memo_keyable))
 
 
 def unregister_memo_key_function(typ: type) -> None:
-    """Remove a previously registered memo key function (best-effort)."""
+    """Remove registered memo key function and stable type ID (best-effort)."""
```
georgeh0: nit: we can remove the `(best-effort)` here.

docs/src/content/docs/advanced_topics/memoization_keys.mdx
```diff
@@ -78,21 +111,85 @@ register_memo_key_function(Path, path_key)
...
+    ExternalRecord,
+    lambda record: (record.id, record.version),
+    stable_type_id="com.example.ExternalRecord/v1",
+)
+```
+
+:::note[Pickle fallback]
+Types using pickle fallback fingerprint raw bytes without type namespacing. Setting or registering a stable type ID alone has no effect on them — register a memo key function together with the stable type ID.
+:::
+
```
georgeh0, line 169: Somewhere we may want to briefly mention the "full replace" semantics, e.g.

Each call replaces the type's entire registration: register the key function, state function, and stable type ID together in one call.

python/cocoindex/_internal/memo_fingerprint.py
MaxRong: Sounds good, I'll implement this immediately.

MaxRong:
## PR Update

Just pushed the latest patch of changes addressing the comments:

### What Changed
- Inherited declared stable type IDs via `getattr` with MRO registration fallback (Declared -> Registered MRO -> module.QualName).
- Renamed the primary registration function to `register_memo_type` with full-replacement semantics for key function, state function, and stable type ID.
- Retained `register_memo_key_function` as a backward-compatible shortcut taking its original signature.
- Enforced string type checks for stable type IDs, prevented raw class MRO traversal from reaching `object`, and updated docs and tests.

### Key Decisions
- When a base class registers a key function, subclasses inherit and execute that key function, but any declared or registered stable type ID on the subclass overrides the identity namespace. If the subclass defines no stable ID, it falls back to the base owner's identity.
- Considered consolidating registered key-owner and stable-ID resolution into a single MRO pass, but decided to keep `_type_identity_parts()` decoupled. Key extraction and type-identity namespacing are distinct responsibilities (consumed independently by intrinsic hooks, dataclasses, and Pydantic models), and thought combining them would unnecessarily couple the dispatch loop especially since declared IDs already resolve in O(1) via `getattr` and bypass MRO traversal entirely. Additionally, python MROs are rarely deep.
- For the legacy shortcut `register_memo_key_function`, replaces `key_fn`/`state_fn` but intentionally preserves any existing `stable_type_id`. The legacy signature cannot accept `stable_type_id`, so clearing it on key-function registration would be unexpected and detrimental for those using the legacy function.
- Maintained symmetrical behavior in the internal test fixture functions `unregister_memo_type` and `unregister_memo_key_function`. Kept intentionally private to reduce public API surface (and discourage improper patterns).

### CI Note
The standard Windows Python 3.11 job timed out at the 30.0s mark on an untouched test (`test_bound_method_memo_with_use_mount`) due to runner I/O latency. Could we rerun the Github Actions suite?

PR #2262, opened 2026-07-06
Slides to Speech: local CPU TTS (Pocket TTS) + DSPy vision

prrao87:
Modernizes the slides-to-speech example end-to-end, with a cheaper CPU-based TTS model and a more idiomatic DSPy pipeline to make the program more composable/optimizable. Also updates to the latest versions of CocoIndex, LanceDB SDKs and uses `gemini-3.5-flash` as the reasoning model.

## What changed

- **TTS is now fully local on CPU** — swapped Piper (which leaned on a GPU runner) for Kyutai's Pocket TTS (~100M params, CPU-only, no API). The CPU synthesis runs one-at-a-time on a worker thread, so it never blocks the event loop, and the slides' vision-LLM calls still run concurrently
- **DSPy for the vision LLM** — replaced instructor + LiteLLM with a typed DSPy signature + `dspy.Predict`. DSPy can be better than instructor + LiteLLM for programs that have even a tiny chance of being reused, composed, evaluated, or optimized later — the call stays a declarative signature rather than hand-written prompt + JSON-parsing glue.
- **Model & voice** — vision model gemini-3.5-flash; narration uses the "anna" voice.
- **Dependencies** — tracks cocoindex 1.0.16 and its LanceDB 0.34.0 connector; adds an `audioop` backport so MP3 export works on Python 3.13+.
- **Query fix** — search now uses cosine distance, so printed relevance scores read a sensible 0–1 instead of going negative.
- **Docs** — refreshed the walkthrough and example catalog to match.

## Verified

Ran end-to-end: a 3-slide deck yields 3 LanceDB rows (notes + MP3 + embedding), and semantic search ranks the expected slide first with correct scores; audio is valid 24 kHz MP3.

### DIFF
--- a/docs/src/content/example-posts/slides-to-speech.md
+++ b/docs/src/content/example-posts/slides-to-speech.md
@@ -1,71 +1,68 @@
 ---
 title: Slides to Narrated Search
-description: 'Turn slide decks into a narrated, searchable index with CocoIndex V1 — a vision LLM writes speaker notes for each slide, Piper synthesizes them to audio locally, and the notes are embedded into LanceDB for semantic search.'
+description: 'Turn slide decks into a narrated, searchable index with CocoIndex V1 — a vision LLM writes speaker notes for each slide, Pocket TTS synthesizes them to audio locally on the CPU, and the notes are embedded into LanceDB for semantic search.'
 slug: slides-to-speech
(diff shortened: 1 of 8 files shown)


badmonster0:
Amazing, thanks @prrao87 !

PR #2266, opened 2026-07-10
fix(tests): prevent Windows test environment database path leakage

Sujit-1509:
## Summary

Fixes a Windows-specific bug in the test environment where LMDB databases could be created inside the repository instead of the temporary test directory.

## Root Cause

`create_test_env()` built database names using string manipulation:

- `_PATH_PREFIX` contained forward slashes
- Windows test paths contained backslashes
- `removeprefix()` therefore became a no-op

The resulting path still contained an absolute Windows path (e.g. `C:\Users\...`).

When joined with the temporary directory using `pathlib.Path`, the absolute path discarded the temporary directory entirely, causing LMDB databases to be written into the repository instead of isolated temporary locations.

This leaked state across test runs and caused stale tracking information to persist between tests.

## Fix

- Replace fragile string-based path handling with `pathlib.Path.relative_to()`
- Generate database names from the relative test path
- Normalize path separators into `__`
- Remove the now-unused `_PATH_PREFIX` module-level constant (dead code)
- Ensure every test environment always uses an isolated temporary database

## Result

- Prevents persistent LMDB state leakage on Windows
- Restores proper test isolation
- Eliminates failures caused by stale database state
- 512 tests pass, 0 failures (verified on Windows)

### DIFF
--- a/python/tests/common/environment.py
+++ b/python/tests/common/environment.py
@@ -14,18 +14,20 @@ def get_env_db_path(name: str) -> pathlib.Path:
     return _tmp_db_path_base / name
 
 
-_PATH_PREFIX = str(pathlib.Path(__file__).parent.parent) + "/"
-
-
 def create_test_env(
     test_file_path: str,
     suffix: str | None = None,
     *,
     exception_handler: cocoindex.ExceptionHandler | None = None,
 ) -> cocoindex.Environment:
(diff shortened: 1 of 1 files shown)


georgeh0 (approved):
Thanks for the fix!

PR #2267, opened 2026-07-10
Make slides-to-speech incremental per slide

prrao87:
## Summary

This updates the slides-to-speech example so that a slide, as well as for the entire PDF, are the incremental processing boundaries. Either unchanged files _or_ slides should be able to be reused without repeating the expensive vision calls, text-to-speech, or embedding work.

Also, in the previous version, changing the Pocket TTS voice ("Anna" → "Vera") wouldn't result in a recomputation. This update passes in the voice name as a context key so CocoIndex can now track it, and this would detect voice changes as well as model name changes and trigger the recompute accordingly.

## Why

The previous example memoized only the PDF-level processing function, but not per slide. Any change to a PDF invalidated that function and remapped every slide, which made a one-slide edit look like a whole-deck update. That is both expensive and misleading for an example intended to demonstrate incremental, multimodal indexing.

## What changed

- The PDF stage renders pages and mounts one memoized processing component per page.
- Each slide component declares its own LanceDB row, allowing CocoIndex to retain unchanged rows and recompute only changed slide inputs.
- The selected Pocket TTS voice is now a change-detected context value, so voice changes invalidate narration deliberately.
- The example documentation now describes the per-slide boundary and the resulting behavior when a PDF changes.
- The search helper relies on LanceDB's default ranked vector search instead of calculating a custom similarity score.

## Validation

- `uv run ruff format --check examples/slides_to_speech/main.py`
- `uv run ruff check examples/slides_to_speech/main.py`
- A public-release integration check covering unchanged, edited, added, and removed slides. An edited page recomputed only that page's vision, TTS, and embedding work; unchanged pages retained their rows.

@badmonster0 the blog post proposal for this is in https://github.com/cocoindex-io/blog-proposal/pull/2.

### DIFF
--- a/examples/slides_to_speech/main.py
+++ b/examples/slides_to_speech/main.py
@@ -173,7 +173,7 @@ async def process_slide(
     )
 
 
-@coco.fn
+@coco.fn(memo=True)
 async def process_file(file: FileLike, table: lancedb.TableTarget[SlideRecord]) -> None:
     slides = await pdf_to_slides(await file.read())
     await coco.mount_each(
--- a/examples/slides_to_speech/main.py
+++ b/examples/slides_to_speech/main.py
@@ -40,11 +40,11 @@ from cocoindex.resources.file import FileLike, PatternFilePathMatcher
 LANCEDB_TABLE = "slides_to_speech"
 LANCEDB_URI = os.environ.get("LANCEDB_URI", "./lancedb_data")
 EMBED_MODEL = "sentence-transformers/all-MiniLM-L6-v2"
-POCKET_TTS_VOICE = os.environ.get("POCKET_TTS_VOICE", "alba")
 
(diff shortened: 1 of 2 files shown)


examples/slides_to_speech/main.py
```diff
@@ -173,10 +173,15 @@ async def process_slide(
     )
 
 
-@coco.fn(memo=True)
+@coco.fn
```
georgeh0: I think we can keep this memoized too, so unchanged files can be skipped very fast.

georgeh0 (commented):
Thanks a lot for the improvements!

examples/slides_to_speech/main.py
prrao87: Great catch, thank you! Fixed.

georgeh0 (approved):
Thanks a lot for the nice improvement!

PR #2271, opened 2026-07-13
Add CocoIndex + zvec video hybrid search example

Haleshot:
This is a _first_ pass at a local-video CocoIndex + zvec example.

The example watches a local videos/ folder, cuts each video into short scenes, embeds one keyframe with CLIP, transcribes the audio with faster-whisper, and writes one row per scene into `zvec`. Each row has a dense keyframe vector, a full-text transcript field, and scalar fields for the video path and time window. The query script stays on the zvec side and combines dense search, transcript FTS, and scalar filters (based on what the [connector](https://cocoindex.io/docs/connectors/zvec/) supports now).

The CocoIndex part is the indexing and freshness layer: add or delete a clip and the scene rows follow; change the CLIP model and the embedding step reruns while the transcript step stays cached. If the vector size changes, zvec rebuilds the collection, which the README calls out.

The example is tested on two Blender open movie clips, downloaded at runtime rather than checked in, with an attribution file for the sources. There's also a generated offline fallback for people who just want to test the pipeline mechanics. 

_Not added to the docs [examples gallery](https://cocoindex.io/docs/examples/) yet. I kept this PR focused on getting an example up and running; happy to add the docs/gallery wiring in a _follow-up_ if this works!_

_Related: https://github.com/cocoindex-io/blog-proposal/issues/1_

### DIFF
--- a/examples/video_hybrid_search/main.py
+++ b/examples/video_hybrid_search/main.py
@@ -195,6 +195,27 @@ def _extract_keyframe(video_path: str, ts: float) -> bytes:
     return out.stdout
 
 
+def _has_audio_stream(video_path: str) -> bool:
+    out = subprocess.run(
+        [
+            "ffprobe",
+            "-v",
+            "error",
+            "-select_streams",
+            "a",
+            "-show_entries",
+            "stream=index",
+            "-of",
+            "csv=p=0",
+            video_path,
+        ],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return bool(out.stdout.strip())
+
+
 def _extract_audio(video_path: str, start: float, dur: float) -> bytes:
     out = subprocess.run(
         [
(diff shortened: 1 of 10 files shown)


examples/video_hybrid_search/main.py
```diff
@@ -0,0 +1,368 @@
...
+    with tempfile.NamedTemporaryFile(suffix=pathlib.Path(video_path).suffix) as tmp:
+        tmp.write(content)
+        tmp.flush()
+        duration = await asyncio.to_thread(_ffprobe_duration, tmp.name)
+
+        for start, end in _scene_bounds(duration):
+            scene_id = _scene_id(video_path, start)
+            mid = (start + end) / 2.0
+            frame_bytes = await asyncio.to_thread(_extract_keyframe, tmp.name, mid)
+            audio_bytes = await asyncio.to_thread(
```
feihongxu0824: Silent videos currently fail indexing because _extract_audio() is called unconditionally. Could we skip transcription and use an empty transcript when no audio stream exists, so visual search still works?

Haleshot: Ah, good catch, thanks. I built this around the sample clips, which all have audio, so I assumed every video has an audio stream and didn't guard _extract_audio.

_Edit_: Fixed it such that when there isn't an audio stream, it skips transcription and stores an empty transcript, so the keyframe and visual side still indexes.

PR #2274, opened 2026-07-13
Add shared sparse vector support

tomz-alt:
## Summary

- add a canonical shared `SparseVector` value type, `SparseVectorSchema`, and connector-kit schema resolver
- add sparse-vector support for PostgreSQL/pgvector and shared-value interop for zvec and Qdrant
- fail fast in unsupported connectors and reject sparse mappings before they can be written as dense vector keys
- preserve nullable and nested `Annotated` metadata precedence while centralizing validation
- document the shared API and add connector, resolver, normalization, and regression coverage

## Why

Sparse vectors were represented differently by each connector. That left nullable annotations, mapping-form values, schema conflicts, and backend overrides with inconsistent behavior. In the worst case, a sparse mapping could be accepted as a dense vector and only its keys would be persisted.

This change establishes one validated interchange format and one schema-resolution contract, then makes each connector either support it explicitly or reject it before writing.

## Impact

PostgreSQL can now create and write `sparsevec(n)` columns, including HNSW indexes. zvec and Qdrant accept the same canonical sparse values, while dense-only connectors provide actionable errors. Existing zvec list-based dense vectors and legacy sparse definitions remain supported.

## Validation

- `uv run pytest -q python/`: 1,155 passed, 180 skipped
- `uv run mypy`: no issues in 281 files
- focused Ruff checks and `git diff --check`
- docs build: 98 pages plus agent-output validation
- real on-disk zvec insert/update/query coverage
- real pgvector PostgreSQL sparse round-trip and HNSW index test
- live Qdrant 1.15.5 dense, sparse, and hybrid query test

### DIFF
--- a/dev/agent-skills/target-connector/SKILL.md
+++ b/dev/agent-skills/target-connector/SKILL.md
@@ -217,6 +217,37 @@ _shared_sink = coco.TargetActionSink.from_fn(_apply_actions)
 
 When building queries from user-provided names (table, column, index) or values (record IDs, keys), you must guard against injection and ensure correctness. See [input_safety.md](input_safety.md) for patterns on identifier validation, parameterized queries, and value escaping.
 
+### Vector Schema Resolution
+
+Table-like connectors that infer dense or sparse vector columns from annotations must use the shared connector-kit resolver:
+
+```python
+from cocoindex.connectorkits import resolve_vector_schemas
+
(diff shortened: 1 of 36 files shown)


ZhiHanZ:
I think it is ready for review @badmonster0 could you help me open it from the draft

badmonster0:
marked and tagged @georgeh0 ty!

PR #2277, opened 2026-07-14
Rewrite the GPUPool in Rust (#2243)

### LINKED ISSUE #2243: [FEATURE] Support multiple GPU and fractional GPU allocations  - RUST impl
follow up comments from https://github.com/cocoindex-io/cocoindex/pull/2224 from george

I'm thinking if it's good to put the most intensive logic (e.g. the GPU resource allocation) to Rust side? I think the logic will also grow further. Putting it on Rust side will enable maximal reuse (e.g. by the Rust SDK in the future).

---

YaxinCheng:
Closes #2243

# Summary
* Rewrite GPUPool in Rust and expose it as PyObj
* Support >1 integer values to acquire multiple fully available GPUs
* Introduce `temp-env` dev dependency for 

# Motivation
Previous pr (#2224) introduced the `GPUPool` to the system, but it was written in Python. Rewriting it in Rust would enhance the reusability for future Rust SDK and would improve performance for this most intensive logic. 

# Changes
## Refactored: `GPUPool` (`gpu_pool.rs`)
* Completely rewrote `GPUPool` in Python (`runner.py`) to `GPUPool` (`gpu_pool.rs`) allows acquiring and releasing a fraction of GPU resources from and to a GPU. 
  * acquiring and releasing logics are updated now to be more optimized
* Rewrote related helper functions such as `detect_num_gpus` in Rust, which reads different environment variables to determine the number of GPUs.
* Add a new function that accepts an integer value `n` where `n >= 1` and acquires `n` number of **fully available** / **non-occupied** / **capacity = 1.0** GPUs. When there are not enough GPUs, no GPU would be acquired until **all** are available to be acquired

## New `PyGPUPool`
To use the `GPUPool` in Python, a new bridge object `PyGPUPool` is created. It exposes all the public functions from `GPUPool` to the Python end.

## New [temp_env](https://crates.io/crates/temp-env) dev dependency for tests
Added `temp_env` dev dependency to properly test functions that depends on environment variables.

Calling [std::env::set_var](https://doc.rust-lang.org/std/env/fn.set_var.html) is unsafe and it messes up the actual environment. For tests, using `temp_env` is more manageable and safer.

## Tests (`gpu_pool.rs`)
22 tests:
* Rewrote all python tests for `GPUPool` and `detect_num_gpus` in Rust.
* New tests testing `acquire_full_gpus`:
  * Acquire when there are enough GPUs
  * Acquire when GPUs are not enough, and can only succeed after others release
  * No partial acquiring test
  * Attempting to acquire more GPUs than available

Since the `detect_num_gpus` can be calling `nvidia-smi` external command to acquire number of GPUs, in the test this behaviour is stubbed / mocked using `echo` to directly return a given value passed in. This stub can be controlled through a few **test-only** environment variables using `temp-env`. For example, `MOCK_NVIDIA_SMI_STDOUT` controls what is returned from the `echo` command, and `MOCK_NVIDIA_SMI_EXIT_CODE` controls what exit code is returned from running the command.

# Backward Compatibility 
The `GPUPool` is bridged into Python to replace the original Python `GPUPool`. The callers have been updated to use the new function names if there are any.


badmonster0:
thanks a lot @YaxinCheng , @georgeh0 can help take a look!

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    /// Acquires a given integer number of fully available GPUs (capacity == 1.0) from the GPU pool.
+    ///
+    /// # Warning
+    /// * When the given gpu_count is larger than the total gpus, it crashes.
+    /// * All GPUs will be acquired at simultaneously.
+    ///   For instance, if user attempts to acquire 5 GPUs,
+    ///   the function will not partially acquire 4 and waiting for the last GPU.
+    pub async fn acquire_full(&self, gpu_count: NonZeroUsize) -> Vec<usize> {
```
georgeh0: This function is only used by tests now.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,504 @@
...
+            .take(count.get())
+            .collect()
+    }
+
+    pub async fn release(&self, gpu_id: usize, fraction: f32) {
+        {
+            let mut cap = self.capacity.lock().await;
+            cap[gpu_id] += fraction;
+        }
+        self.release.notify_waiters();
```
georgeh0: I think there may be a thundering herd problem here.  e.g. if N calls to `acquire()`, after 1 GPU becomes available, all N will be notified, and N-1 of them will be wasted. The overall time complexity of notifying will be O(N^2), which may become a bottleneck.

To be fair, the original Python code has exactly the same problem. But since Rust core is a more permanent location for this logic, it's a better time to make this more robust now.

rust/utils/src/gpu_pool.rs
YaxinCheng: Just to clarify, we are not going to make this function accessible publicly? It will only be used for test? I have changed it to test only for now

rust/utils/src/gpu_pool.rs
YaxinCheng: Updated with `notify_one()` instead. I originally thought about using `notify_one`, but wanted to keep in sync with the Python logic

rust/utils/src/gpu_pool.rs
YaxinCheng: Ugh, I misunderstood you. I have now exposed it to python interface now. PHAL

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// * All GPUs will be acquired at simultaneously.
+    ///   For instance, if user attempts to acquire 5 GPUs,
+    ///   the function will not partially acquire 4 and waiting for the last GPU.
+    pub async fn acquire_full(&self, gpu_count: NonZeroUsize) -> Vec<usize> {
+        assert!(
+            gpu_count.get() <= self.num_gpus(),
+            "Attempted to acquire {} GPUs but only has {}.",
+            gpu_count.get(),
+            self.num_gpus
+        );
```
georgeh0: For this case, I think we should return an error. `assert!` will panic and crashes the entire process, which should be avoided.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,504 @@
...
+            {
+                let mut cap = self.capacity.lock().await;
+                if let Some(index) = Self::find_available(&cap, fraction) {
+                    cap[index] -= fraction;
+                    return index;
+                }
+            }
+            notified.await;
+        }
+    }
```
georgeh0: - If there're 2 tasks, each need 0.5 GPU in queue, now 1 GPU is released, the 1st task will be notified and scheduled, but the 2nd task will still wait until the first one is done (which isn't necessary).

- If there's 1 task that needs 1 GPU, and there're many tasks that need 0.5 GPUs and they keep being added. There's never 1 whole GPU get available, so the task that needs 1 GPU may never get a chance to be scheduled (a.k.a. starvation).

rust/utils/src/gpu_pool.rs
YaxinCheng: Fixed this by changing it to return Result<Vec<usize>> so it does not crash. Similarly, it will throw exception in Python if the wrong number is given

rust/utils/src/gpu_pool.rs
YaxinCheng: I have switched to a more verbose but better solution here. Instead of the broadcasting, I created a queue of waiters, and each waiter has two parameters: `(desired_capacity, channel_sender)`. It works as follows:

1. When we try to acquire GPUs, we lock the `capacity` in the pool, find if we can get a GPU (or GPUs) that satisfy:
  * if yes, good, everything works as before
  * if no, we register this as a waiter by creating a oneshot channel, and put the desired capacity (either a fraction f32 number, or a usize count for multiple GPUs) and the sender of the channel to the queue. Then wait for the GPU label (or labels for multiple GPUs) while waiting at the reader.
2. When we try to release a fraction to a GPU, we lock the `capacity` and add the fraction as before. Then we start to dequeue our waiting queue, and there will also be a few conditions:
  * if the waiter is not waiting anymore (sender closed), we can just skip it
  * if we can find a GPU (or multiple GPUs) that satisfies the required capacity of the waiter, **we deduct the capacity in the `release` function**, then send the GPU label (or labels) back through the sender
  * if we cannot find a GPU (or multiple GPUs) that satisfies, we simply add it back to the queue to wait longer

This strategy avoids the hunger issue, and also avoids the thunder herd.

I am thinking, we can potentially go even one step further with the queue. Currently it's a FIFO queue, so the waiters are handled FIFO. We could instead use a PriorityQueue, so waiters need smaller capacities can go earlier, and the waiters waited longer can also get some boost. But this would need further discussions. Let me know what you think, thanks

YaxinCheng: @georgeh0 wonder if this is what you have in mind? Let me know what you think, thanks

georgeh0: Thanks @YaxinCheng!

The current solution solves the 1st problem, but looks like it still doesn't solve the 2nd problem (starvation): a task that needs 1 GPU may never get scheduled if tasks that need 0.5 GPUs keep being added to the queue.

I think there's one more potential issue: if there're N pending tasks in the queue, each release for a single one is O(N) and overall it's O(N^2) after N tasks.

Using a PriorityQueue will make smaller tasks easier to be fulfilled faster, but it'll make the issue of large tasks starvation worse.

Here's my thoughts:

* An acceptable solution is: strict FIFO, i.e. schedule tasks at queue front until as long as there're available GPUs, until a task that we don't have sufficient available GPU (even if the remaining GPU is sufficient for a later task).

* A further improvement: The above avoids the starvation problem, but with the tradeoff that a later task always needs to wait for an earlier to be scheduled *even if scheduling for the later one doesn't deprive the earlier one*.

     e.g. in queue: Task1(gpu=1), task2(gpu=0.5). When we have available GPU = [0.5, 0.5], actually we only need to *reserve* one specific GPU for task1 to avoid its starvation, not have to be both. 

    "reserve" means for this GPU, the current and any additional available quota will also be allocated to a certain task until it's scheduled. For a task at queue front, if it already get it's reserved GPU(s) and there're still other available GPUs, we can try to reserve or allocate other GPUs to the next task in the queue.

The first one is acceptable in my mind, while the 2nd one is more ideal if it can be implemented.

Let me know if you have any other thoughts and ideas. Thanks!

YaxinCheng: Thanks for the reply. Let me start with the first solution, and then we have a separate PR for the second if we can. Would this work for you? I will have this PR updated soon, hopefully by end of tomorrow

georgeh0: Sounds good. Thanks!

YaxinCheng: Hi @georgeh0 , I made some changes to jump start to the second solution. Please have a look. Now when we try to acquire one or more GPU resources, and there is not enough, it will enter into FIFO queues waiting for resources for specific GPUs. For other GPUs not reserved, they are free to be used. Each GPU has their own queue of task waiting for resources. I am not feeling well today, so the code may be buggy. If this is in the right direction, I can fix the issues later.

Also one extra question, do we need to return Error if the `acquire` function retries to acquire a fraction > 1.0, similarly if a `release` function tries to release to the sum of > 1.0, should we return errors?

YaxinCheng: @georgeh0  I have a commit (not pushed yet) which makes `acquire` and `release` return errors when unexpected values are given (e.g. release 1.5 fraction, or acquire 1.5 fraction).  It also changes the new function parameter from `usize` to `NonZeroUsize`. Let me know if this is necessary. Thanks

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,685 @@
...
+/// ``CUDA_VISIBLE_DEVICES``, or ``nvidia-smi`` (falling back to 1).
+/// Call ``configure_gpu_pool(N)`` to override programmatically.
+pub struct GPUPool {
+    num_gpus: usize,
+    state: Mutex<GPUPoolState>,
+}
+
+struct GPUPoolState {
+    capacity: Vec<f32>,
+    reserved: Vec<VecDeque<(f32, oneshot::Sender<()>)>>,
```
georgeh0: Actually there's a pitfall by using floating-point here: operations like `x - fraction + fraction` are not guaranteed to return x, and over time, the accumulator will drift.

We may actually consider using an integer here, e.g. use a scale factor (such as `1000000`) and store `f` multiplied by the factor .

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,685 @@
...
+        capacity: &[f32],
+        fraction: f32,
+        exclude: &[VecDeque<T>],
+    ) -> Option<usize> {
+        let max_position = capacity
+            .iter()
+            .enumerate()
+            .filter(|(gpu_id, _)| exclude[*gpu_id].is_empty())
+            .max_by(|(_, cap_a), (_, cap_b)| cap_a.partial_cmp(cap_b).unwrap())
+            .map(|(gpu_id, _)| gpu_id)?;
```
georgeh0: maybe we can consider picking the one with just sufficient available capacity, i.e. among those "available >= fraction" pick the least one.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,685 @@
...
+        Self::reserve_gpu_with_exclusion(fraction, state, &HashSet::new())
+    }
+
+    fn reserve_gpu_with_exclusion(
+        fraction: f32,
+        state: &mut GPUPoolState,
+        exclude_gpus: &HashSet<usize>,
+    ) -> (usize, oneshot::Receiver<()>) {
+        let (sender, recv) = oneshot::channel();
+        let reserved_gpu = Self::find_shortest_queue(&state.reserved, exclude_gpus);
```
georgeh0: I think using shortest queue may still end up with sub-optimal scheduling, e.g. if there's task running super long, tasks will be pending in the queue while others are idle (without either running or reserved tasks).

In my mind a better approach will be:

- One shared queue for all waiting tasks.

- Each GPU is reserved for at most 1 task at any time, as there's no point to reserve a CPU for a second task before the existing one is scheduled.

    - Fractional (<1 GPU): When capacity opens up on a reserved GPU, the reserved task either absorbs the freed capacity or acquires enough total capacity to be scheduled.
    - Full GPU(s): Claims intact GPUs. While waiting for remaining GPUs, its already-reserved GPUs are held and not given to others.

- Queue processing: When a GPU without a reserved task has capacity, inspect the head of the shared queue. The head task reserves/allocates available GPUs; once all its required GPUs are reserved, it pops off the queue.

Let me know if it makes sense. Thanks!

rust/utils/src/gpu_pool.rs
georgeh0: Hi @YaxinCheng,

Hope you're feeling better!

Yes, returning `Error` for unexpected arguments is the right call.

rust/utils/src/gpu_pool.rs
YaxinCheng: So if I understand correctly, when a GPU frees up, it goes to the shared queue to get the next task. What will happen if the capacity of this GPU is not enough to handle the requirement of the next task? Say, the GPU has a capacity of 0.6 after releasing, but the top task in the queue required 0.7? Does it just wait until the top task is handled by another GPU, then check the top of the queue again?

rust/utils/src/gpu_pool.rs
YaxinCheng: Updated. Now it finds the min among all the available >= fraction

rust/utils/src/gpu_pool.rs
YaxinCheng: I have introduced a new type `GPUFraction` which scales up the f32 to a u32 internally. It also implements Add and Sub, so we can easily use it internally to represent the fractions

rust/utils/src/gpu_pool.rs
YaxinCheng: IIUC, acquire would then have two behaviours:
1. if we can find a GPU with enough capacity, we assign to the GPU and deduct the capacity
2. if not, we enqueue the task to the shared queue and wait

In this case, for release, let's represent it in pseudocode:

```
fn release(capacity, gpu) {
  capacities[gpu] += capacity

  loop { 
    if !has_reservation(gpu) && !task_queue.is_empty() {
      add_reservation(gpu, task_queue.pop_left())
    }
    if let Some(reserved_task) = get_reservation(gpu) {
      if reversed_task.demand_capacity <= capacities[gpu] {
        capacities[gpu] -= reversed_task.demand_capacity
        remove_reservation(gpu)
        reserved_task.notify_processing()
      }
    }
    if has_reservation(gpu) || task_queue.is_empty() {
      break
    }
  }
}
```

is this what you meant?

georgeh0: The pseudocode almost aligns with the direction in my mind (and it already answers your question in the previous comment), with one revision:  if the task at the front of the queue **requires multiple GPUs**, it's only popped out of the queue once all required GPUs are reserved for it.

e.g. if there're 3 GPUs (`G1`, `G2`, `G3`) all without available capacity, 2 tasks in the queue, `T1` requires 2 GPU and `T2` requires 0.5 GPU

```
GPUs: G1 (capacity=0), G2(capacity=0), G3(capacity=0)
Queue: T1(req=2) T2(req=0.5)
```

Now G1 is released with 0.5 capacity, we can reserve it for T1, but T1 remains at the front of the queue:

```
GPUs: G1 (capacity=0.5, reserved_for=T1), G2(capacity=0), G3(capacity=0)
Queue: T1(req=2, reserved=[G1]) T2(req=0.5)
```

Later, G2 is released with 0.5 capacity, we can reserve it for T1, now T2 is reserved with 2 GPUs, so it can be popped from the queue front

```
GPUs: G1 (capacity=0.5, reserved_for=T1), G2(capacity=0.5, reserved_for=T1), G3(capacity=0)
Queue: T2(req=0.5)
Fully reserved tasks: T1(req=2, reserved=[G1, G2])
```

From now on, if any available capacity is released for G3, it can be reserved for T2.

WDYT?

YaxinCheng: Hi @georgeh0 , thanks for the explanation.

I have a draft implementation on this, but I feel like there are a few more scenarios that need clarifications.

1. Should we randomly reserve or keep the task in the queue?
```
GPUs: G1 (cap=0.5), G2 (cap=0.6)
Queue: T1 (req=0.7)
```
My concern is, if we reserve, say G1, then maybe G2 will be freed up next and we cannot change it

2. Following case 1, if we do not reserve, and a new task comes in:
```
GPUs: G1 (cap=0.5), G2 (cap=0.6)
Queue: T1 (req=0.7), T2 (req = 0.1)
```
this will get stuck

Which path should we pick?

georgeh0: Hi @YaxinCheng ,

Good question.

I think we may just reserve either G1 or G2 for T1. Probably G2 is better as the available cap is higher. Yes, if G1 is freed up sooner, G1 will be idle which is a waste. But since we only reserve a GPU for at most 1 task at a time, the waste is bounded. So I think it's acceptable. If we go with 2 (i.e. not reserve), there'll be risk of starvation, which is worse.

Besides, there's an alternative way in my mind now: instead of reserving a fixed GPU determined at reserving time, we may reserve N GPUs with most available capacity,  dynamically. 

- In your example, when `T1` comes, we reserve "the GPU with most available cap" for it, which may change over time. Later, if `G1` is freed up, now `G1` has most capacity, so it's reserved one, and `T1` can be assigned to it.
- If multiple tasks have reserved GPUs, e.g.  `T1` and `T2` each has 1 reserved GPU, if `T1` comes earlier, then the GPU with most cap is reserved for `T1`, the GPU with the 2nd most cap is reserved for `T2`, etc.

This will make sure each task always get more available GPUs than tasks coming later, but later tasks still get a chance to be scheduled earlier if there're extra GPUs that fits it.

This alternative way might be slightly better, but may be more complex to implement.

WDYT?

YaxinCheng: > If multiple tasks have reserved GPUs, e.g.  T1 and T2 each has 1 reserved GPU, if T1 comes earlier, then the GPU with most cap is reserved for T1, the GPU with the 2nd most cap is reserved for T2, etc.

This will have another issue when T1 requires more resources than T2, and the most available one has enough for T2 but not T1. Due to the ordering, T2 will have to wait.

Or in a different case, 

```
GPUs: G1 (cap=0.4, reserved=T1), G2 (cap = 0.3, reserved=T2)
Tasks: T1 (req=0.6), T2 (req=0.4)
```
If we release 0.2 to G2, it will become 0.5, which is enough for T2 but not T1. However, since T1 reserves the most available one, now it will have G2, and T2 will have G1

```
GPUs: G1 (cap=0.4, reserved=T2), G2 (cap = 0.5, reserved=T1)
Tasks: T1 (req=0.6), T2 (req=0.4)
```

In this case, the two tasks will still be stuck. Is this actually okay?

YaxinCheng: Hi @georgeh0 , the current version does the basic acquiring and releasing:

1. when it acquires (including acquire_full), it acquires, or reserves, or get put into the queue
2. when it releases, it will add back the capacity, then keep popping the head of the queue if it can be fulfilled or reserved

I am thinking maybe we can have a separate PR for the ordering solution. But before that, please have a look at this version and let me know. Thank you

YaxinCheng: Hi @georgeh0 , I was able to update the reserving logic to the flex strategy we discussed. The first reserved task will use the GPU with the most capacity, even when the capacities change, the most available will be reserved for this task. Please have a look, I have reduced the logic, so it should be slightly easier to review. Thank you

YaxinCheng: I have taken a look at the 1200 lines of code again, most of them are test related code or comments. Sorry for bloating up the CL, but it may be the most concise it can be at this stage. Thanks for taking the time to review it

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1102 @@
...
+/// The default pool size is auto-detected from ``COCOINDEX_NUM_GPUS``,
+/// ``CUDA_VISIBLE_DEVICES``, or ``nvidia-smi`` (falling back to 1).
+/// Call ``configure_gpu_pool(N)`` to override programmatically.
+pub struct GPUPool {
+    num_gpus: usize,
+    capacities: Mutex<SortedVec<GPUCapacity>>,
+    task_queue: Mutex<VecDeque<PendingTask>>,
+}
+
+type PendingTask = (GPUCapacity, oneshot::Sender<usize>);
```
georgeh0: consider define a `struct` for it. It'll have better readability (similar to Google's C++ style guide)

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1102 @@
...
+/// GPU with enough remaining capacity is available, then returns its id.
+/// ``release(gpu_id, fraction)`` restores capacity and wakes waiters.
+///
+/// The default pool size is auto-detected from ``COCOINDEX_NUM_GPUS``,
+/// ``CUDA_VISIBLE_DEVICES``, or ``nvidia-smi`` (falling back to 1).
+/// Call ``configure_gpu_pool(N)`` to override programmatically.
+pub struct GPUPool {
+    num_gpus: usize,
+    capacities: Mutex<SortedVec<GPUCapacity>>,
+    task_queue: Mutex<VecDeque<PendingTask>>,
```
georgeh0: I feel sharing the same `Mutex` (use a `struct` to wrap these 2) is clearer and simpler. When we access `capacities` we're already always holding the mutex for `task_queue`.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1102 @@
...
+    /// e.g. "the GPU with the most available capacity" is reserved for the head task in the queue.
+    ///
+    /// The function would try to host the task using the remaining GPUs,
+    /// or it will send it to the queue which will reserve a GPU now or later.
+    ///
+    pub async fn acquire(&self, fraction: GPUCapacity) -> Result<usize> {
+        if fraction == GPUCapacity::ZERO {
+            return Err(anyhow::anyhow!(
+                "Acquired fraction must be between 0.0 and 1.0, got 0"
+            ));
```
georgeh0: Let's use error types and helper functions/macros defined in https://github.com/cocoindex-io/cocoindex/blob/main/rust/utils/src/error.rs

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1102 @@
...
+        }
+        let receiver = {
+            let mut task_queue = self.task_queue.lock().await;
+            if task_queue.len() < self.num_gpus {
```
georgeh0: This condition seems not taking tasks requiring multiple GPUs into consideration.

rust/utils/src/gpu_pool.rs
georgeh0: I think synchronous `Mutex` (i.e. `std::sync::Mutex`) may just work here. When we hold the mutex, we only need calculations without any `await` (note: I think the sender can be called after we pop tasks out of the queue and release the mutex), so `Mutex` is just sufficient and more lightweight.

georgeh0 (commented):
Thanks! Haven't gone through all code yet. Sent out what I have now.

rust/utils/src/gpu_pool.rs
YaxinCheng: For tasks taking multiple GPUs, we will have multiple tasks consecutively. So this should be fine

rust/utils/src/gpu_pool.rs
YaxinCheng: Done

rust/utils/src/gpu_pool.rs
YaxinCheng: I was going to argue that there are cases we don't need to lock `capacities`, then realized there is no case that we only lock `capacities`. So, agreed, and done

rust/utils/src/gpu_pool.rs
YaxinCheng: I did not realize the error exists. I have switched. Let me know if this is not the right way to use the macros

rust/utils/src/gpu_pool.rs
YaxinCheng: I was under the impression that sync::Mutex would block tasks from executing in tokio, but turns out it was wrong. As long as the MutexGuard does not pass .await, it will be fine. I have updated to use `std::sync::Mutex ` now. Thanks

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1117 @@
...
+    }
+
+    /// processes pending task queue following the rules:
+    ///
+    /// 1. strict FIFO - when task A comes in earlier than task B, task A always gets processed earlier.
+    /// 2. only pop if available - the task remains in the queue until a GPU has enough availability to process it.
+    ///
+    fn process_task_queue(pool: &mut PoolState) {
+        while let Some(pending_task) = pool.task_queue.front()
+            && let Some(gpu_id) = pool.capacities.find(&pending_task.demand)
```
georgeh0: Here seems not strictly FIFO.

If the 1st task in the queue requires 1 but the GPU with maximum capacity only has 0.5, and the 2nd one only requires 0.5, this will assign the GPU to the second one. This may end up with starvation for the 1st task.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1117 @@
...
+            );
+        }
+        let (mut acquired_gpus, receivers) = {
+            let mut pool = self.state.lock().expect("lock poisoned");
+            let mut acquired_gpus = Vec::with_capacity(gpu_count);
+            if pool.task_queue.len() < self.num_gpus {
+                while acquired_gpus.len() < gpu_count
+                    && let Some(gpu_id) = pool
+                        .capacities
+                        .find_excluding_top_n(&GPUCapacity::MAX, pool.task_queue.len())
```
georgeh0: This function need to skip the first `pool.task_queue.len()` each time (by the `nth()` call), and it's in a loop. I think we can only need to scan the `BTreeSet` once, e.g. pass in a number to `find_excluding_top_n()` and make it return a `Vec` (can also expose a different function).

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1117 @@
...
+/// ``CUDA_VISIBLE_DEVICES``, or ``nvidia-smi`` (falling back to 1).
+/// Call ``configure_gpu_pool(N)`` to override programmatically.
+pub struct GPUPool {
+    num_gpus: usize,
+    state: Mutex<PoolState>,
+}
+
+struct PoolState {
+    capacities: SortedVec<GPUCapacity>,
+    task_queue: VecDeque<PendingTask>,
```
georgeh0: I think the name is a little bit confusing, since an entry is not a task. Maybe we can use something like "acquisition" instead of "task".

rust/utils/src/gpu_pool.rs
YaxinCheng: If the first task is in queue, say it requires 0.6, then it reserves the one with the max capacity, say 0.5. Every other GPU should have a capacity less than or equal to 0.5 in this case. The second task comes in requires 0.5, and it will only be assigned if there is another GPU with 0.5 capacity (the max one is reserved by the first task in queue), or it will be put into the queue and wait for the first task to be processed. This should be fine, right?

rust/utils/src/gpu_pool.rs
YaxinCheng: Updated to `Acquisition`

rust/utils/src/gpu_pool.rs
YaxinCheng: Updated to separate the excluding and finding. Now it should only call the `nth` once then find a vec of GPU ids. Nice catch, thanks

rust/utils/src/gpu_pool.rs
```diff
@@ -414,6 +413,33 @@ mod container {
...
+
+        pub fn find_n(&self, target: &T, count: usize) -> Vec<usize> {
+            self.find_n_iter(target, count).collect()
+        }
+
+        fn find_n_iter(&self, target: &T, count: usize) -> impl Iterator<Item = usize> {
+            (self.upper_bound.is_some() && target <= &self.upper_bound.unwrap().0)
+                .then(|| {
+                    self.source
+                        .range(&(target.clone(), 0_usize)..=self.upper_bound.unwrap())
```
georgeh0: I feel this `SplitSortedVec` is complicating things.

We only need to skip the top `pool.acquisition_queue.len()` then check the next N. It's a straightforward iterator operation. Why do we need an abstraction `SplitSortedVec` to note the upper bound and then `range()`?

rust/utils/src/gpu_pool.rs
georgeh0: Sorry, looking at the code again, I misunderstood that. There won't be starvation. 

But it has another problem, e.g.

- Available GPUs: 0.5 0.4
- Tasks demad: 0.6, 0.3

In this case, we reserve the 1st GPU for the 1st task, and the 2nd GPU can be assigned to the 2nd task.


Note that "blocking others until the first is scheduled" is a simple and feasible solution. And this is the original solution we discussed earlier this [this thread](https://github.com/cocoindex-io/cocoindex/pull/2277#discussion_r3611615812). But its implementation can be much simpler: just a single queue without a reservation system. Since we're already using a reservation system, we should achieve its benefit: it reserves a dedicated GPU to the earlier task, at the same time, allocations for the remaining GPUs are not blocked by it.

rust/utils/src/gpu_pool.rs
YaxinCheng: Was thinking maybe it could be more readable if I split the excluding and finding. I have merged them

rust/utils/src/gpu_pool.rs
YaxinCheng: I am sorry, I don't get this. In the example you provided, the second task (0.3) won't even be in the queue. Because when the task comes in, the situation is:

```
GPUs: G1 (0.5, reserved), G2 (0.4)
Queue: T1 (0.6)
```

The task 2 shows up, T2(0.3), and it will directly be assigned, and won't even reach the queue. Everything in the queue is strictly Fifo. Even if we managed to work around and put the T2 into the queue, then we have

```
GPUs: G1 (0.5, reserved), G2 (0.4)
Queue: T1 (0.6), T2 (0.3)
```

In this case, the queue is frozen. Even if T2 can technically be handled by G2, it won't. Because in the code, the `process_acquisition_queue` strictly follows FIFO, and it won't work on T2 until T1 is done.

As mentioned [here](https://github.com/cocoindex-io/cocoindex/pull/2277#discussion_r3663345535):

>  instead of reserving a fixed GPU determined at reserving time, we may reserve N GPUs with most available capacity, dynamically
> ... when T1 comes, we reserve "the GPU with most available cap" for it, which may change over time. Later, if G1 is freed up, now G1 has most capacity, so it's reserved one, and T1 can be assigned to it...

What is the issue that we need to fix here? Are you saying, if T1 reserves G1, and we cannot assign any task until T1 is finished, even though there are enough GPUs available to process T2?

georgeh0: > In the example you provided, the second task (0.3) won't even be in the queue. Because when the task comes in, the situation is:
>
> ```
> GPUs: G1 (0.5, reserved), G2 (0.4)
> Queue: T1 (0.6)
> ```
> 
> The task 2 shows up, T2(0.3), and it will directly be assigned, and won't even reach the queue.

e.g. when T2 shows up, G2 may have less capacity (e.g. `0.2`) than its demand, so T2 will be inserted into the queue:

```
GPUs: G1 (0.5, reserved), G2 (0.2, reserved)
Queue: T1 (0.6), T2 (0.3)
```

Later, G2 has another 0.2 capacity becoming available, now we have,

```
GPUs: G1 (0.5, reserved), G2 (0.4, reserved)
Queue: T1 (0.6), T2 (0.3)
```

Please let me know if it's not how the current code is working.

> What is the issue that we need to fix here? 

The problem we want to solve is: T1 already has its reserved GPU, so it shouldn't prevent other GPUs to be assigned to other tasks. This is holding more than the necessary GPU(s) for a task, and will cause inefficiency.


> In this case, the queue is frozen. Even if T2 can technically be handled by G2, it won't. Because in the code, the `process_acquisition_queue` strictly follows FIFO, and it won't work on T2 until T1 is done.
> 
> As mentioned [here](https://github.com/cocoindex-io/cocoindex/pull/2277#discussion_r3663345535):
> 
>  >  instead of reserving a fixed GPU determined at reserving time, we may reserve N GPUs with most available capacity, dynamically
>  > ... when T1 comes, we reserve "the GPU with most available cap" for it, which may change over time. Later, if G1 is freed up, now G1 has most capacity, so it's reserved one, and T1 can be assigned to it...

To clarify, the problem I'm trying to point out in my latest comment is not about FIFO unfollowed (this was what my first comment is saying, which was misleading and based on my misunderstanding for potential starvation; I corrected in my next comment).

FIFO should be applied to tasks before they're either reserved or assigned. Once they already get their reserved GPU, later tasks get a chance to be reserved or assigned, no more FIFO here. If we follow "strict FIFO" even for reserved tasks, we shouldn't even directly assign non-reserved GPU to new task when it already has enough capacity, right? We don't have to be subject to "strict FIFO across all tasks including reserved ones".

Also the paragraph quoted above ("... when T1 comes ...") is just using a simple example to explain how dynamic adjusted reservation works for one task. It says nothing about freezing later GPUs/tasks in the queue.

YaxinCheng: > e.g. when T2 shows up, G2 may have less capacity (e.g. 0.2) than its demand, so T2 will be inserted into the queue:
>
> ```
> GPUs: G1 (0.5, reserved), G2 (0.2, reserved)
> Queue: T1 (0.6), T2 (0.3)
> ```
>
> Later, G2 has another 0.2 capacity becoming available, now we have,
> 
> ```
> GPUs: G1 (0.5, reserved), G2 (0.4, reserved)
> Queue: T1 (0.6), T2 (0.5)
> ```

Yes, this is correct. It will be like this. But I think we previously agreed that the queue should be strict FIFO. In this case, G2 should not be handled, because it's blocked by G1. Do we want the G2 to be handled in this case instead?

georgeh0: Strict FIFO is one acceptable option I discussed [in this early thread](https://github.com/cocoindex-io/cocoindex/pull/2277#discussion_r3611615812), but the implementation should be much simpler and don't need the "reserve" mechanism.

Since we already introduce the "reserve" mechanism, it's no longer strict FIFO. As discussed in that thread, the main motivation of introducing the "reserve" mechanism is to avoid the head task block others when there're more available GPUs. If we want to go with strict FIFO, "reserve" is not needed at all.

YaxinCheng: I see. I think I have misunderstood it. I have just pushed a commit, which should allow T2 in this case gets assigned to G2. The queue is no longer strictly FIFO. PHAL

YaxinCheng: @georgeh0 can you please have a look and let me know if there are any other issues? Thank you

python/cocoindex/_internal/runner.py
```diff
@@ -213,110 +213,6 @@ def in_subprocess() -> bool:
...
-                count = int(output[0].strip())
-                if count > 0:
-                    return count
-    except Exception:
-        pass
-
-    return 1
-
-
-class GPUPool:
```
georgeh0, line 259: `GPUPool` is removed from `runner`, but it's still imported in [`api.py`](https://github.com/cocoindex-io/cocoindex/blob/aee7b27da33a6cdf6beac5794f443cb6d100316a/python/cocoindex/_internal/api.py#L95)

python/cocoindex/_internal/core.pyi
```diff
@@ -704,3 +704,39 @@ class RateLimiter:
...
+    
+    async def acquire_full(self, gpu_count: int) -> list[int]:
+        """
+        Acquires a given integer number of fully available GPUs (capacity == 1.0) from the GPU pool.
+
+        The gpu_count should be greater or equal to 1 and less than or equal to the number of GPUs in the pool.
+        """
+        ...
+
+    async def release(self, gpu_id: int, fraction: float) -> None:
```
georgeh0: I think this method isn't `async` according to the Rust side's implementation.

python/cocoindex/_internal/core.pyi
```diff
@@ -704,3 +704,39 @@ class RateLimiter:
         cls, max_rows_per_second: float, burst_window_secs: float = 1.0
     ) -> "RateLimiter": ...
     def acquire(self, n: int = 1) -> Coroutine[Any, Any, None]: ...
+
+########################################################
+# GPU pool
+########################################################
+
+class GPUPool:
+    def __init__(self, num_gpus: int = 1) -> None: ...
```
georgeh0: This `= 1` isn't consistent with the definition on the Rust side.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1266 @@
...
+            }
+            let gpus_to_be_acquired = gpu_count - acquired_gpus.len();
+            let receivers = std::iter::repeat_with(|| {
+                Self::send_acquisition_to_queue(&mut pool.acquisition_queue, GPUCapacity::MAX)
+            })
+            .take(gpus_to_be_acquired)
+            .collect::<Vec<_>>();
+            (acquired_gpus, receivers)
+        };
+        let gpu_ids = futures::future::try_join_all(receivers).await?;
```
georgeh0: We need to consider cancellation safety. e.g. if the coroutine is cancelled during this `await` in the middle, we may leak some GPU capacities: they may never be refilled. We may consider add a drop guard.

YaxinCheng: Updated the code to add an error handling branch. Now it should release the already acquired GPUs when the waiting is cancelled. PHAL

python/cocoindex/_internal/core.pyi
YaxinCheng: Done

rust/utils/src/gpu_pool.rs
georgeh0: Thanks Yaxin!

But I don't think the updated code covers the cancellation case.

`try_join_all(receivers)` only returns `Err` when a **sender** is dropped, i.e. an `Acquisition` is dropped from the pool side without a `send`. In production that never happens: entries only leave the queue via a successful `send` in `process_acquisition_queue`.

Actual cancellation works in the opposite direction: when the Python task is cancelled (`asyncio.wait_for` timeout, task cancellation at shutdown), pyo3 drops the Rust **future** mid-`await`. Then neither match arm ever runs, and the partially acquired GPUs are still leaked. This reproduces it against the current head:

```rust
#[tokio::test]
async fn acquire_full_cancelled_by_caller() -> Result<()> {
    let pool = Arc::new(GPUPool::new(NonZeroUsize::new(3).unwrap()));
    let busy = pool.acquire(GPUCapacity::unchecked(0.6)).await?; // GPU0 partially busy
    let cloned = pool.clone();
    let task =
        tokio::spawn(async move { cloned.acquire_full(NonZeroUsize::new(3).unwrap()).await });
    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
    task.abort(); // == caller-side cancellation: drops the future mid-await
    let _ = task.await;
    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
    let state = pool.state.lock().unwrap();
    assert_eq!(state.capacities[1], GPUCapacity::MAX); // FAILS: stays ZERO forever
    assert_eq!(state.capacities[2], GPUCapacity::MAX); // FAILS: stays ZERO forever
    Ok(())
}
```

When I run this, capacities stay `[0.4, 0, 0]` and the dead `MAX` entry stays in the queue, where it keeps reserving the top-capacity GPU and gating the fast path.

What I had in mind with the drop guard is roughly:

```rust
struct AcquireGuard<'a> {
    pool: &'a GPUPool,
    acquired: Vec<usize>,                     // fast-path GPUs already zeroed
    receivers: Vec<oneshot::Receiver<usize>>, // queued demands
}

impl Drop for AcquireGuard<'_> {
    fn drop(&mut self) {
        for gpu in self.acquired.drain(..) {
            let _ = self.pool.release(gpu, GPUCapacity::MAX);
        }
        for mut rx in self.receivers.drain(..) {
            // Closes the race where release() already sent a gpu_id (capacity already deducted).
            if let Ok(gpu) = rx.try_recv() {
                let _ = self.pool.release(gpu, GPUCapacity::MAX);
            }
            // Dropping rx makes any later send() fail → no deduction.
        }
    }
}
```

`acquire_full` awaits through the guard and disarms it on success; `acquire` wraps its single receiver the same way (releasing `fraction` instead of `MAX`). `Drop` runs outside the pool lock in all these paths, so calling `release()` from it is fine.

Two related points:

- A cancelled fractional `acquire` has the same problem: the dead entry keeps counting toward `queue.len()` and keeps a GPU reserved until some release happens to satisfy its demand. e.g. 1 GPU with 0.9 held, a cancelled `acquire(0.5)` in the queue — a new `acquire(0.05)` that fits the free 0.1 blocks behind the phantom. Skipping entries whose `notifier.is_closed()` in `process_acquisition_queue` (without counting them as pending) fixes that.
- Minor: the `client_bail!` in the new branch reports a pool-side condition, not a caller mistake — `internal_error!` fits better if the branch stays.

I'd also suggest making the regression test use `task.abort()` (as above) rather than clearing the queue, since that's the shape production cancellation actually takes.

Thanks!

YaxinCheng: I see what you meant. Very good catch. I have updated the code accordingly, and hopefully it catches the potential bugs once for all. Please let me know if there are missed cases. Thank you!

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+        pool.acquisition_queue
+            .retain(|acq| !acq.notifier.is_closed());
+        let length = pool.capacities.len();
+        let mut pending_acquisitions = Vec::with_capacity(length);
+        while pending_acquisitions.len() < length
+            && let Some(acquisition) = pool.acquisition_queue.pop_front()
+        {
+            if acquisition.notifier.is_closed() {
+                continue;
+            }
```
georgeh0, line 273: This check is unreachable: the `retain` at the top of this function already removed all closed entries, and nothing can close a notifier while we hold the state lock (both guards' `Drop` impls take the lock first). Fine to delete, or keep it with a comment marking it as defensive.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+                pool: self,
+                acquired_gpus,
+                receivers,
+                defused: false,
+            }
+        };
+
+        while let Some(res) = guard.receivers.next().await {
+            let gpu_id = res.map_err(|err| {
+                client_error!("GPUPool reservation cancelled while waiting: {err}")
```
georgeh0, line 212: This is the same underlying condition as line 145 in `acquire` (sender dropped without a send), but there it's `internal_error!` and here `client_error!`. Let's make both `internal_error!` for consistency: it's a pool invariant violation, not a caller mistake.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+    receiver: oneshot::Receiver<usize>,
+    defused: bool,
+}
+
+impl<'a> Drop for AcquireGuard<'a> {
+    fn drop(&mut self) {
+        if self.defused {
+            return;
+        }
+        let mut state = self.pool.state.lock().expect("lock poisoned");
```
georgeh0, line 47: Inside a `Drop` impl, prefer `lock().unwrap_or_else(|e| e.into_inner())` over `.expect(...)` (here and in `AcquireFullGuard::drop` at line 71): if the lock is ever poisoned, panicking inside `Drop` during unwinding aborts the whole process. The `expect`s in normal functions are fine.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+    /// 1. The `n` is the size of the queue.
+    /// 2. The top n are defined by capacities.
+    /// 3. The top n are reserved for acquisitions already in the queue.
+    ///
+    /// Reservation is for an abstract concept,
+    /// e.g. "the GPU with the most available capacity" is reserved for the head acquisition in the queue.
+    ///
+    /// The function would try to host the acquisition using the remaining GPUs,
+    /// or it will send it to the queue which will reserve a GPU now or later.
+    ///
```
georgeh0, line 116: Could you document the cancellation guarantee here (and on `acquire_full`)?

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+    /// After releasing 0.5 capacity to G1:
+    /// ```text
+    /// GPUs: G1(capacity=0.5), G2(capacity=0), G3(capacity=0)
+    /// Queue: T1(req=0.7, reserved=[G1]), T2(req=0.5, reserved=[G2])
+    /// ```
+    /// After releasing 0.6 capacity to G2:
+    /// ```text
+    /// GPUs: G1(capacity=0.5), G2(capacity=0.6), G3(capacity=0)
+    /// Queue: T1(req=0.7, reserved=[G2]), T2(req=0.5, reserved=[G1])
+    /// ```
```
georgeh0, line 238: Under this example, T2 should be able to be scheduled on G1.

python/cocoindex/_internal/runner.py
georgeh0, line 17: It's unused now.

georgeh0, line 329: `release` is now a sync method on the Rust side.

python/cocoindex/_internal/core.pyi
```diff
@@ -704,3 +704,39 @@ class RateLimiter:
...
+        """
+        Releases a fraction of capacity back to the specified GPU ID.
+        """
+        ...
+
+    @staticmethod
+    def default() -> "GPUPool": ...
+
+    @staticmethod
+    def from_config(config_str: str) -> "GPUPool": ...
```
georgeh0, line 742: It doesn't exist on the Rust side now.

rust/py/src/gpu_pool.rs
```diff
@@ -0,0 +1,82 @@
+//! Python bindings for batching infrastructure.
+//!
+//! Exposes BatchQueue and Batcher to Python for implementing batched function execution.
+//!
+//! Design: Multiple batchers can share the same queue (e.g., for GPU serialization),
+//! and each batcher has its own runner function. When a batcher creates a batch,
+//! that batch carries the batcher's runner function.
```
georgeh0, line 7: This module doc (copy-pasted from `batching.rs`) should be updated.

rust/py/src/gpu_pool.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    pub fn acquire<'py>(&self, py: Python<'py>, fraction: f32) -> PyResult<Bound<'py, PyAny>> {
+        let fraction =
+            GPUCapacity::try_from(fraction).map_err(|e| PyValueError::new_err(e.to_string()))?;
+        let gpu_pool = self.inner.clone();
+        future_into_py(py, async move {
+            gpu_pool
+                .acquire(fraction)
+                .await
+                .map_err(|e| PyValueError::new_err(e.to_string()))
```
georgeh0, line 52: Let's use `.into_py_result()` (from the prelude, `rust/py_utils/src/error.rs`) instead of hand-rolling `PyValueError` here and in the other four conversion sites.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+    ///
+    /// 1. ``COCOINDEX_NUM_GPUS`` environment variable (explicit override).
+    /// 2. ``CUDA_VISIBLE_DEVICES`` environment variable (count of entries).
+    /// 3. ``nvidia-smi`` command output (if available).
+    /// 4. Default to ``1``.
+    ///
+    fn detect_num_gpus() -> Result<usize> {
+        if let Some(env_num) = std::env::var("COCOINDEX_NUM_GPUS")
+            .ok()
+            .and_then(|s| s.parse::<usize>().ok())
```
georgeh0, line 312: I think we want to fail loudly on parse failure here. We may trim, parse non-empty strings and fail on parse failure.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1566 @@
...
+        if let Ok(cuda_visible) = std::env::var("CUDA_VISIBLE_DEVICES") {
+            let count = cuda_visible
+                .split(',')
+                .map(str::trim)
+                .filter(|s| !s.is_empty())
+                .count();
+            return Ok(std::cmp::max(1, count));
+        }
+        #[cfg(not(test))]
+        let output = std::process::Command::new("nvidia-smi")
```
georgeh0, line 325: The old Python detection used `subprocess.run(..., timeout=5)`; `Command::output()` here has no timeout. `nvidia-smi` hanging on a wedged driver is a known failure mode, and this runs inside the sync `PyGPUPool::default()` — with the GIL held — under the default-pool lock on the first GPU acquire, so a hang freezes the whole process. Two asks:

1. Restore a bounded wait (e.g. ~5s) with a fallback to 1, like the Python version had.
2. Release the GIL around the probe: have `PyGPUPool::default(py: Python<'_>)` wrap the `GPUPool::default()` call in `py.detach(|| ...)`. That doesn't unblock the calling thread (the event loop still waits up to the timeout on first acquire), but it keeps a slow/hung probe from stalling every other Python thread in the process.

georgeh0 (commented):
Thanks for the fix! Most parts look good now. The new comments are mostly small mechanical ones.

rust/utils/src/gpu_pool.rs
YaxinCheng: The original python did not fail though. I think it used the default value 1. But I have changed the code to panic if the environment variable exists but it unable to be parsed

rust/py/src/gpu_pool.rs
YaxinCheng: Updated. PHAL. Thank you

rust/utils/src/gpu_pool.rs
YaxinCheng: Updated. I used a loop and polling to do the timeout, let me know if there is a better way of doing it.

Also the GIL is updated, not sure if that's what you meant, so please have a look

rust/utils/src/gpu_pool.rs
YaxinCheng: I have updated the doc to correct this. PHAL

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1599 @@
...
+    /// # Detection order:
+    ///
+    /// 1. ``COCOINDEX_NUM_GPUS`` environment variable (explicit override).
+    /// 2. ``CUDA_VISIBLE_DEVICES`` environment variable (count of entries).
+    /// 3. ``nvidia-smi`` command output (if available).
+    /// 4. Default to ``1``.
+    ///
+    fn detect_num_gpus() -> Result<usize> {
+        if let Some(env_num) = std::env::var("COCOINDEX_NUM_GPUS").ok() {
+            match env_num.parse::<usize>() {
```
georgeh0: let's `trim()` before parsing, and also treat empty string (after trim) as not set.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1599 @@
...
+                Ok(num) => return Ok(std::cmp::max(1, num)),
+                Err(err) => panic!("Failed to parse COCOINDEX_NUM_GPUS={env_num}: {err}"),
```
georgeh0: Let's avoid `panic!` here. We may propagate error by `Result` all the way up into Python side, so it'll be a Python exception rather than panicking.

rust/utils/src/gpu_pool.rs
```diff
@@ -0,0 +1,1599 @@
...
+                    .take()
+                    .and_then(|mut err| err.read_to_end(&mut stderr).ok());
+                return Ok(std::process::Output {
+                    status,
+                    stdout,
+                    stderr,
+                });
+            }
+            if start.elapsed() >= timeout {
+                let _ = child.kill();
```
georgeh0: `kill()` sends the signal but doesn't reap — add `let _ = child.wait();` after it, or the killed child stays a zombie for the process lifetime.

Issue #2300, opened 2026-07-18
[FEATURE] Persist readable names for provider-only target-state path segments


junzh0u:
#2296 is making `cocoindex show` render target-state paths readably by resolving fingerprinted segments from persisted records (the owner index + the owning component's tracking items) and, for live apps, the provider registry.

One class of segment remains unresolvable: **provider-only path elements** — segments that are providers but never declared target states, so nothing persisted maps their fingerprint back to a key. Concretely:

1. **Root providers when inspecting via `--db`/`--app-name`** — registered by name at module import; with no module loaded, the registry is empty and the first segment of every path renders as `#hex`.
2. **Attachments of nested providers, in all flows** — e.g. a table's `vector_index`: the attachment provider is registered only in the build-local registry (discarded when the run finishes) and has no owner-index entry, so `/@pg/"my_table"/#hex/"idx"` never fully resolves, even with the app loaded.

Proposal: persist segment names at provider registration/declaration time — e.g. a small `Fingerprint → StableKey` keyspace written idempotently (write-once per unique segment; names are tiny). The inspect resolver would consult it as an additional source, eliminating the `#hex` fallback for both cases above.

Considerations: cleanup policy (segments are shared across paths; leaving names forever is probably fine given their size), and whether to key by lone segment fingerprint or full prefix path.

Issue #2296, opened 2026-07-17
[FEATURE] improve cocoindex CLI's support for inspecting tracked metadata for target states


georgeh0:
There're two gaps:

- Currently, `cocoindex show` supports listing tracked metadata for components. It shows tracked target states too, but each part of the target state path is a fingerprint, which is not readable, e.g.

  ```
  /@process_file/"1706.03762v7.md"
      type:component version:1 processor:process_file
      has_memoization:true target_state_count:31
      Target states:
        - path:/#96330b5a3630afc01b1795d8252d5cb5/#4866c946624f9cd1978aa71f36c12173/#0ccbdd4e7da2776b816d55ae631db7be[provider_id=0] key:(13,)
          states:1:Existing schema_version:3 generation:None
        - path:/#96330b5a3630afc01b1795d8252d5cb5/#4866c946624f9cd1978aa71f36c12173/#1619563f097cbdbb7a5065a1198b0033[provider_id=0] key:(19,)
          states:1:Existing schema_version:3 generation:None
        - path:/#96330b5a3630afc01b1795d8252d5cb5/#4866c946624f9cd1978aa71f36c12173/#1a5e80a0cde6d04b955760e351ab37b8[provider_id=0] key:(18,)
          states:1:Existing schema_version:3 generation:None
        - path:/#96330b5a3630afc01b1795d8252d5cb5/#4866c946624f9cd1978aa71f36c12173/#1ec659bd62c4867c2d50cb722e01ab66[provider_id=0] key:(14,)
          states:1:Existing schema_version:3 generation:None
        - path:/#96330b5a3630afc01b1795d8252d5cb5/#4866c946624f9cd1978aa71f36c12173/#1fc77917a99f30ad0e6905272eb767d0[provider_id=0] key:(12,)
          states:1:Existing schema_version:3 generation:None
  ```

  We should display them in a readable way.

- Currently there's no way to list tracked target state paths, i.e. entries under [`TargetState`](https://github.com/cocoindex-io/cocoindex/blob/efff4a996db17356838addde77bfeaeafe83ce7b/rust/core/src/state/db_schema.rs#L162)

Potential way to get readable target state key from the key in forms of fingerprint:
[...]

PR #2301, opened 2026-07-18
feat: render tracked target-state paths readably in cocoindex show

junzh0u:
First half of #2296: `cocoindex show` renders tracked target-state paths readably instead of as opaque fingerprints.

## Before / after

`examples/files_transform`, `cocoindex update main` then `cocoindex show main -l`, same database on both sides:

**Before:**
```
  /@process_file/"bizarre_animals.md"
    type:component version:1 processor:process_file
    has_memoization:true target_state_count:1
    Target states:
      - path:/#9142763800ab4da61971eddf69b897de/#8c5e6914737a99848bcdefd959e71a3c key:(None, 'output_html/data__bizarre_animals.md.html')
        states:1:Existing schema_version:0 generation:None
```

**After:**
```
  /@process_file/"bizarre_animals.md"
    type:component version:1 processor:process_file
    has_memoization:true target_state_count:1
    Target states:
      - path:/@cocoindex/localfs/[null,"output_html/data__bizarre_animals.md.html"]
        states:1:Existing schema_version:0 generation:None
```

**After, with `--fingerprints`** — toggles back to the stored form (readable is the default):
```
  /@process_file/"bizarre_animals.md"
    type:component version:1 processor:process_file
    has_memoization:true target_state_count:1
    Target states:
      - path:/#9142763800ab4da61971eddf69b897de/#8c5e6914737a99848bcdefd959e71a3c
        states:1:Existing schema_version:0 generation:None
```

## How it works

Fingerprints are one-way hashes, but the original `StableKey`s are recoverable from records the engine already persists: for each path prefix, resolve the owning component via the inverted owner index, then storekey-decode the pre-fingerprint `key` from that component's tracking item (cached per read txn). Root providers and their attachments are never declared as target states, so they resolve from the live provider registry instead. Anything unresolvable falls back to `#hex` — never an error.

## Decisions

1. **Reconstruct at display time, not persist readable names.** No schema change, works retroactively on existing databases. Persisting is more robust (no joins, survives crash windows) but needs a migration — deferred to #2300; the approaches compose.
2. **Readable replaces the existing `target_state_path` string; raw stays available** via the new `fingerprint_path` field and the `--fingerprints` flag, since fingerprints remain the join key against the physical DB.
3. **Roots from the live registry.** Covers `show <app>` fully; `--db` runs without importing the app's modules, so roots keep the `#hex` fallback there — also closed by #2300.
4. **Dropped the `key:` display field** — the leaf of the readable path *is* the key now (the old field used Python repr: `(13,)` vs `[13]`). The API attribute remains.

## Follow-up

#2300: persist segment names for provider-only segments (`--db` roots; attachments of nested providers, which live in discarded build-local registries).

## Tests

Rust: fallback, nested-chain reconstruction (`root → "table" → 13`, incl. `[provider_id=N]`), registry-seed roots. E2E: `show -l` fully readable from a loaded app; `--fingerprints` raw form; `show --db -l` leaf-readable with root fallback.

### DIFF
--- a/rust/core/src/inspect/db_inspect.rs
+++ b/rust/core/src/inspect/db_inspect.rs
@@ -106,9 +106,15 @@ struct TargetKeyResolver {
 }
 
 impl TargetKeyResolver {
-    fn new() -> Self {
+    /// `provider_keys` seeds the cache with the original keys of live
+    /// registered providers (see [`provider_key_seed`]); root provider
+    /// segments are resolvable only through this seed.
+    fn new(provider_keys: std::collections::HashMap<TargetStatePath, StableKey>) -> Self {
         Self {
-            cache: std::collections::HashMap::new(),
+            cache: provider_keys
+                .into_iter()
+                .map(|(path, key)| (path, Some(key)))
+                .collect(),
         }
     }
 
@@ -197,6 +203,26 @@ impl TargetKeyResolver {
         }
         Ok(rendered)
     }
+}
+
+/// Collect the original keys of all target-state providers registered in the
+/// live environment: root providers and their attachments, registered by name
+/// at module import time and never persisted, so this registry is the only
+/// source that can render their path segments readably. Two classes of
+/// provider-only segments remain unresolvable and keep the fingerprint
+/// fallback: providers whose modules aren't imported in this process (e.g.
+/// inspecting a database without loading the app), and attachments of nested
+/// providers, which live in build-local registries that are discarded when
+/// the run finishes.
+fn provider_key_seed<Prof: EngineProfile>(
+    env: &Environment<Prof>,
+) -> std::collections::HashMap<TargetStatePath, StableKey> {
+    let registry = env.target_states_providers().lock().unwrap();
+    registry
+        .providers
+        .iter()
+        .map(|(path, provider)| (path.clone(), provider.stable_key().clone()))
+        .collect()
 }
 
 fn summarize_target_state_items(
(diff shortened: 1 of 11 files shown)


badmonster0:
thanks @junzh0u ! @georgeh0 can help take a look!

rust/core/src/inspect/db_inspect.rs
```diff
@@ -96,12 +100,143 @@ fn decode_target_state_key(key_bytes: &[u8]) -> StableKey {
     }
 }
 
+/// Resolves fingerprinted target state path segments back to their original
+/// `StableKey`s via the inverted owner index + the owner's tracking info.
+/// Caches per path prefix, including misses, so items sharing ancestors are
+/// resolved once per read txn.
+struct TargetKeyResolver {
+    cache: std::collections::HashMap<TargetStatePath, Option<StableKey>>,
+}
```
georgeh0, line 109: Ideally this should be shared across multiple `StablePath`'s resolve. Everything provider paths will have a high potential of reuse, which will reduce tons of DB read operations.

If you prefer, can also be addressed in a follow-up PR.

georgeh0 (approved):
Looks pretty good! Thanks!

One comment, which can be addressed separately if you prefer.

rust/core/src/inspect/db_inspect.rs
junzh0u: Make sense! Let me gauge how trivial a change it is and come back with either a code change or a follow up proposal.

junzh0u: Here is my proposal after evaluation: the cache is already shared within a txn (`--tree`/targeted queries use one resolver for the whole subtree) — the gap is `show -l`, which makes one call per path, each with a fresh txn and resolver. Fixing that means holding one txn/resolver across the whole listing (e.g. a streaming detail API), which I'd rather do as a follow-up than grow this PR. Note #2300 would also make each resolve a single point lookup instead of the current two-get join, so my plan is to do that first and re-measure whether the cross-call caching is still worth the plumbing.

georgeh0: sounds good! thanks for the proposal!

PR #2302, opened 2026-07-18
docs(agents): document prek as the pre-submission check suite

junzh0u:
AGENTS.md's build/test commands are a subset of what CI actually enforces — the prek hook suite additionally runs `cargo fmt`, CLI docs regeneration (`docs/cli.mdx` from `cli.py`), and a few other checks, so changes verified only with the listed commands can still fail CI (as happened on #2301's first push).

Adds a Pre-submission Checks section documenting `prek run --all-files`, plus `cargo fmt` alongside the ruff commands and a row in the workflow summary table.


badmonster0:
thanks @junzh0u ! cc @georgeh0

PR #2303, opened 2026-07-19
feat: add --target-states listing to cocoindex show

junzh0u:
Second half of #2296: `cocoindex show --target-states` lists every entry in the target-state keyspace with the component that owns it — the inverted view of `show -l`, which only shows target states nested under each component and can't surface entries whose owner is gone or incomplete.

## The new listing

`examples/files_transform`, after `cocoindex update main`:

```
$ cocoindex show main --target-states
Target states:
  /@cocoindex/localfs/[null,"output_html/data__bizarre_animals.md.html"]
    owner:/@process_file/"bizarre_animals.md"
  /@cocoindex/localfs/[null,"output_html/data__chuck_norris.md.html"]
    owner:/@process_file/"chuck_norris.md"
```

**With `--fingerprints`** — same toggle as the detail view, renders paths as stored:

```
$ cocoindex show main --target-states --fingerprints
Target states:
  /#9142763800ab4da61971eddf69b897de/#8c5e6914737a99848bcdefd959e71a3c
    owner:/@process_file/"bizarre_animals.md"
  ...
```

**Via `--db`** — works without importing the app's module; the root provider segment keeps the `#hex` fallback there (same known gap as the first half, closed by #2300):

```
$ cocoindex show --db ./cocoindex.db --app-name FilesTransform --target-states
Target states:
  /#9142763800ab4da61971eddf69b897de/[null,"output_html/data__bizarre_animals.md.html"]
    owner:/@process_file/"bizarre_animals.md"
  ...
```

**With `--tree`** — nests entries under their parents; works because stored order yields a parent before its descendants (no path parsing needed). Root provider segments get a node line even though they have no entry of their own:

```
$ cocoindex show main --target-states --tree
Target states:
- @cocoindex/localfs
  - [null,"output_html/data__bizarre_animals.md.html"] owner:/@process_file/"bizarre_animals.md"
  - [null,"output_html/data__chuck_norris.md.html"] owner:/@process_file/"chuck_norris.md"
```

Entries whose owner has no matching tracking item — what a crashed or interrupted run leaves behind — are listed with a `[dangling]` marker instead of being skipped.

## API

New `iter_target_states(_by_name)` async iterators in `cocoindex.inspect`, reusing the path resolver from #2296's first half and the same streaming shape as the component listing. The resolver's caching got coarser along the way (per owner instead of per item), which also speeds up `show -l`. The whole listing runs in one read txn with one resolver instance, so shared provider prefixes resolve once — the cache-sharing raised in #2301's review applies per-call there, but is structural here.

A closer look during review also revealed the old Vec-returning stable-path listing chain is dead (nothing calls it — the Python wrapper collects from the iterator), so a small commit removes it.

## Decisions

1. **A flag on `show`, not a new subcommand** — reuses the APP_TARGET / `--db` plumbing, and `--tree` composes with it. The `STABLE_PATH` argument and `-l`/`-r`/`-p` don't apply to this listing yet (scoping by subtree needs parsing readable paths back into fingerprints — follow-up), so combining errors rather than silently ignoring them.
2. **`--fingerprints` is the one raw-path toggle everywhere** — same replace-not-append semantics as the detail view, rather than a second knob.
3. **Streamed, in stored (fingerprint) order** — changed from a sorted `Vec` after review: target states can outnumber components, and stored order already keeps children of the same parent adjacent (a parent's entry precedes its descendants). No global human ordering; a `--sort` option can come later if wanted.
4. **Dangling entries flagged, not hidden or errored** — consistent with the first half's "never an error" stance.

## Follow-up

#2300: persist provider-only segment names — closes the `#hex` root segment in the `--db` output above.


python/cocoindex/_internal/inspect_api.py
```diff
@@ -108,6 +108,20 @@ async def query_stable_path_details_by_name(
     )
 
 
+async def list_target_states(app: App[Any, Any]) -> list[core.TargetStateEntry]:
+    """List all tracked target states with their owner components."""
+    core_app = await app._get_core()
+    return core.list_target_states(core_app)
```
georgeh0: Ideally we want to return the entries as a stream, similar to the components. The overall amount of target states are similar or even more than the number of components (e.g. for file->embedding use cases, usually a component is a file, which fanouts to many target states).

I understand this will miss the opportunity of sorting. Note that it still preserves affinity for children of the same parent (e.g. `a/a`, `a/b` and `a/c` are placed together after `a`, and they're not interleaved with `b/*`): this property is what we preserve under the current target states key encoding, because it matters more than ordering of children under the same parent. Besides, showing in the same ordering as how we store them by default is also reasonable.

In the future, we may offer a `--sort` option if sorting them by pre-fingerprinted keys is important.

WDYT?

python/cocoindex/cli.py
```diff
@@ -658,13 +668,24 @@ def show(
         raise click.ClickException(
             "-r/--recursive and -p/--parents require a stable_path argument."
         )
+    if target_states and (stable_path or tree or long_format or recursive or parents):
+        raise click.ClickException(
+            "--target-states cannot be combined with stable_path, --tree, -l, -r or -p."
+        )
```
georgeh0: ideally in my mind most of these options (probably other than `-l` for now, unless we start to store more information under these entries in the future) also work together with target states. Target states and components are quite similar that they have hierarchical structure. It's very reasonable if users only inspect a certain sub tree similar to components.

With this said, it's also OK to scope some of the functionalities to later PRs.

python/cocoindex/_internal/inspect_api.py
junzh0u: Done — dropped the sort and switched to streaming, mirroring how the component listing streams; the CLI prints entries as they arrive.

The `Vec` design came from following the existing list-flavored stable-path API — which a closer look revealed is dead (the Python wrapper already collects from the iterator; nothing calls the Vec chain). Streaming was the only live precedent all along; included a small commit cleaning up the dead chain.

Agreed a `--sort` by pre-fingerprinted keys can come later if needed.

python/cocoindex/cli.py
junzh0u: Implemented `--tree` in this PR — it needed no path parsing, since stored order yields parents before descendants. `STABLE_PATH` scoping and `-r`/`-p` need the reverse direction: parsing a readable path back into fingerprints, which means a parser for the full key grammar that doesn't exist yet (current CLI path parsing only handles plain strings). Proposing those as a follow-up.

rust/core/src/state_store/storage.rs
```diff
@@ -432,73 +432,39 @@ impl Storage {
...
+    /// Run `f` with `app_store`'s `(db, txn, sender)` on a
+    /// `tokio::task::spawn_blocking` thread, streaming items over the
+    /// returned channel. LMDB cursors (`RoPrefix`) wrap a raw
+    /// `*mut MDB_cursor` and are `!Send`, so iteration can't be held across
+    /// an `.await`; the sync loop on the blocking-pool thread should use
+    /// `blocking_send` for backpressure and stop when it fails (receiver
+    /// dropped). The rtxn open uses the same `MDB_READERS_FULL` retry policy
+    /// as [`AppStore::read_txn`], but sync (since we're off the runtime).
+    /// An `Err` from `f` is sent as the final item.
+    pub(crate) async fn spawn_read_iter<T, F>(
```
georgeh0: Now the function itself doesn't have anything to do with iterator. It creates a read transaction and returns a receiver, so consider naming it something like `spawn_read_txn_receiver`.

georgeh0 (approved):
Thanks!

rust/core/src/state_store/storage.rs
junzh0u: Renamed to `spawn_read_txn_receiver`.

georgeh0:
Thanks a lot!

Issue #2304, opened 2026-07-19
Add benchmarks for state-store read paths


junzh0u:
Two perf work items are queued with no way to measure them:

* #2301's review flagged that `show -l` makes one call per stable path, each opening a fresh read txn and `TargetKeyResolver`, so nothing is shared across paths. I committed to re-measure after #2300 — but today "re-measure" means eyeballing wall time on a toy example. No baseline, no repeatability.
* #2300 changes what the resolver can answer from the DB (persisted provider segment names), which shifts per-segment resolution cost.

## What exists today

- `rust/sdk/cocoindex/benches/sdk_microbench.rs` — criterion microbench: file walk, msgpack, memo key fingerprinting.
- `benchmarks/file_summarization/` — end-to-end pipeline benchmark (same pipeline in Rust SDK vs Python SDK), measuring cold / warm / incremental runs. Covers update-pipeline and memoization wall time, but as a comparative harness — not a regression baseline for individual paths.
- `benchmarks/entity_resolution/` — synthetic benchmark for `resolve_entities` (call counts, latency breakdown).

Nothing covers the state-store (LMDB) read/inspect paths this issue is about, and no benchmark runs in CI.

## Proposal

Criterion benches in `rust/core` over a seeded synthetic store. One shared fixture that generates a store with a configurable shape — number of components, target states per component, provider nesting depth, shared-prefix fanout — so all benches measure against the same data at, say, 1k / 10k / 100k target states.

Benchmarks:

1. **Per-path detail query, fresh txn+resolver per call** — the current `show -l` shape. This is the baseline the #2301 comment is about; any fix (streaming detail iterator, txn-scoped batching) gets measured against it.
2. **Full target-state listing** (`list_target_states`) — the single-txn/single-resolver counterpart, which also serves as the "how much is on the table" reference for (1).
3. **Resolver cost vs. prefix sharing** — vary how many target states share provider prefixes; this is the axis #2300 moves.

### Running

Local `cargo bench` on demand first; no CI job initially. Once numbers stabilize we can revisit continuous tracking (e.g. criterion history comparison or a hosted tracker), but that's out of scope here.

georgeh0:
Thanks for filing the issue! Very thoughtful!

We definitely need more comprehensive E2E benchmarks. On the other hand, note that the state-store (LMDB) read/inspect path is mainly for inspection / debugging purpose and logic is relatively straightforward (at least for now), so benchmarking is usually less critical.

[...]

PR #2305, opened 2026-07-20
feat: add criterion benchmarks for state-store read paths

junzh0u:
Part of #2304.

Two queued perf work items had no way to be measured: #2301's review flagged that `show -l` opens a fresh read txn + resolver per stable path, and #2300 shifts per-segment resolution cost. Both needed a repeatable baseline instead of eyeballing wall time on toy examples.

This PR adds criterion benches over a store seeded by a **real end-to-end noop app** (per the review discussion below):

- **Seeding fixture** (`benches/state_store_read.rs`, `mod fixture`): a noop CocoIndex app written against the public Rust SDK APIs — `register_root_target_states_provider`, `mount_target`, attachments, `mount_each` components declaring target states. The store is committed by the engine's own precommit/commit path, so the record shapes (tracking blobs, owner-index entries) stay consistent with the engine by construction, with knobs for component count, states per component, provider fanout, and attachment-like provider-only chain depth. The seeded LMDB dir is then reopened store-scoped (no live app), matching the `show --db` flows.
- **Location**: the bench lives in the SDK crate (`rust/sdk/cocoindex/benches/`) — the SDK depends on `cocoindex_core`, so seeding via the SDK from core's benches would be a dev-dependency cycle. The two store-scoped query functions in `db_inspect` become `pub` so the bench measures exactly what the CLI wrappers drive.
- **Benchmark groups** (all against the fixture at ~1k / 10k / 100k target states):
  1. `detail_per_path` — per-path detail query, fresh txn + resolver per call: the current `show -l` shape and the baseline for any txn-scoped batching / streaming-detail fix.
  2. `list_target_states` — full listing in one txn with one shared resolver: the "how much is on the table" reference for (1).
  3. `resolver_prefix_sharing` — same total states, varying how many target states share provider prefixes.
- Runs locally on demand — no CI benchmark job (per the issue). The `cargo test` hook compile-checks (not runs) bench targets so they can't bit-rot.

Full run on this branch (M-series laptop):

```
$ cargo bench -p cocoindex --bench state_store_read
detail_per_path/1k      time:   [6.1276 ms 6.1865 ms 6.2418 ms]
                        thrpt:  [16.021 Kelem/s 16.164 Kelem/s 16.320 Kelem/s]
detail_per_path/10k     time:   [6.5737 ms 6.6009 ms 6.6293 ms]
                        thrpt:  [15.085 Kelem/s 15.150 Kelem/s 15.212 Kelem/s]
detail_per_path/100k    time:   [35.859 ms 36.208 ms 36.517 ms]
                        thrpt:  [2.7384 Kelem/s 2.7618 Kelem/s 2.7887 Kelem/s]
list_target_states/1k   time:   [4.3878 ms 4.5350 ms 4.6732 ms]
                        thrpt:  [217.41 Kelem/s 224.04 Kelem/s 231.55 Kelem/s]
list_target_states/10k  time:   [41.983 ms 42.425 ms 42.878 ms]
                        thrpt:  [233.59 Kelem/s 236.09 Kelem/s 238.58 Kelem/s]
list_target_states/100k time:   [424.76 ms 428.17 ms 431.91 ms]
                        thrpt:  [231.57 Kelem/s 233.59 Kelem/s 235.46 Kelem/s]
resolver_prefix_sharing/list/fanout1
                        time:   [51.547 ms 53.463 ms 55.469 ms]
                        thrpt:  [180.30 Kelem/s 187.06 Kelem/s 194.02 Kelem/s]
resolver_prefix_sharing/detail/fanout1
                        time:   [4.3248 ms 4.3334 ms 4.3430 ms]
                        thrpt:  [23.026 Kelem/s 23.076 Kelem/s 23.123 Kelem/s]
resolver_prefix_sharing/list/fanout100
                        time:   [51.787 ms 52.285 ms 52.817 ms]
                        thrpt:  [191.23 Kelem/s 193.17 Kelem/s 195.03 Kelem/s]
resolver_prefix_sharing/detail/fanout100
                        time:   [7.5088 ms 7.5276 ms 7.5495 ms]
                        thrpt:  [13.246 Kelem/s 13.284 Kelem/s 13.318 Kelem/s]
```

What the numbers say: full listing runs a flat ~220–235K entries/s at every scale, while the per-path shape (`detail_per_path` queries a fixed 100-path sample) pays ~62 µs/path — rising to ~360 µs/path when components hold 100 items — quantifying the #2301 review comment. The `resolver_prefix_sharing` detail rows show fresh-resolver cost growing ~1.7× as fanout goes 1→100, since each call re-resolves more unique provider prefixes. Follow-up PRs use these benches to measure #2300's resolver change and the shared-resolver streaming fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2620,6 +2620,7 @@ dependencies = [
  "ciborium",
  "clap",
  "criterion-plot",
+ "futures",
  "is-terminal",
  "itertools 0.10.5",
  "num-traits",
@@ -2632,6 +2633,7 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "tinytemplate",
+ "tokio",
  "walkdir",
 ]
 
--- a/dev/run_cargo_test.sh
+++ b/dev/run_cargo_test.sh
@@ -70,7 +70,11 @@ if [[ -n "${PYTHONPATH_DETECTED}" ]]; then
 fi
 
 if [[ "${COCOINDEX_SKIP_UV:-}" == "1" ]]; then
-  exec cargo test "$@"
+  cargo test "$@"
 else
-  exec uv run cargo test "$@"
+  uv run cargo test "$@"
 fi
+
+# Plain `cargo test` never compiles bench targets; compile-check (not run)
+# them here to keep them from bit-rotting.
+cargo check -p cocoindex --benches
--- a/rust/core/src/inspect/db_inspect.rs
+++ b/rust/core/src/inspect/db_inspect.rs
@@ -386,7 +386,10 @@ fn read_detail_in_txn(
     })
 }
 
-async fn get_stable_path_detail_from_store(
+/// Store-scoped detail query: one fresh read txn + resolver per call —
+/// the per-path shape `show -l` drives via the App/Environment wrappers
+/// below. Public so benchmarks can measure it directly against a store.
+pub async fn get_stable_path_detail_from_store(
     store: &AppStore,
     provider_keys: std::collections::HashMap<TargetStatePath, StableKey>,
     path: &StablePath,
@@ -606,7 +609,11 @@ fn for_each_target_state_in_txn(
     Ok(())
 }
 
-async fn spawn_target_state_iter(
+/// Store-scoped target-state listing: one read txn + one shared resolver
+/// for the whole iteration. Public so benchmarks can measure it directly
+/// against a store (the App/Environment wrappers below add only the
+/// provider-key seed).
+pub async fn spawn_target_state_iter(
     store: AppStore,
     provider_keys: std::collections::HashMap<TargetStatePath, StableKey>,
 ) -> impl Stream<Item = Result<TargetStateEntry>> + Send + 'static {
--- a/rust/sdk/cocoindex/Cargo.toml
+++ b/rust/sdk/cocoindex/Cargo.toml
@@ -131,7 +131,7 @@ redis = { version = "0.32.7", default-features = false, optional = true, feature
 tempfile = "3"
 serde_json = "1"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
-criterion = "0.5"
+criterion = { version = "0.5", features = ["async_tokio"] }
 glob = "0.3"
 # Used by the live Kafka integration test to consume produced records back.
 rskafka = { version = "0.6", default-features = false }
@@ -142,4 +142,8 @@ rand = "0.8"
 
 [[bench]]
 name = "sdk_microbench"
+harness = false
+
+[[bench]]
+name = "state_store_read"
 harness = false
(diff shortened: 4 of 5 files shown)


rust/core/src/state_store/bench_support.rs
georgeh0, line 1: Is it possible to make the benchmark seeding logic based on CocoIndex's E2E APIs? The seeding logic here is simulating storage behavior of real cocoindex applications, but our engine changes over time, and we have to maintain the seeding logic here to be consistent.

FYI, we had a micro-benchmark in another repo (cocoindex-plus, our enterprise edition) and I just backported it to open source after pruning enterprise-specific stuffs like Postgres support (https://github.com/cocoindex-io/cocoindex/pull/2308/). This uses an E2E cocoindex application that does noop in components' logic and target states.

Will it be possible for us to leverage a similar approach here? I understand the difference - we're using Rust's `criterion` framework for benchmarking, so an E2E application in Python doesn't work well here. Maybe use the Rust SDK or call a lower-level API exposed by the `core` crate. If it's too much, we can also keep this specific DB seeding logic for now and update later after we have a established Rust SDK.

WDYT?

junzh0u: Agreed — ideally the benchmark should be built on public APIs, and the Rust SDK looks like the right route. It already covers everything the fixture needs (root providers, child target states, attachments, mounting N components), so seeding could become a real noop app — same idea as #2308, but in-process so it works with criterion. Options, in order of preference:

1. **Rewrite around the Rust SDK now**, if you consider it stable enough. Records would come from the real commit path, so nothing needs manual syncing with engine changes. Note the bench would move to the SDK crate (SDK depends on `core`, so the reverse would be a dev-dependency cycle).
2. **Ship as is + follow-up issue** to rework the seeding once the SDK is more established — if there's a big overhaul planned that this would collide with.
3. **Treat this as throwaway.** The perf numbers are already captured in the follow-up PRs, so we could abandon this PR and redo it properly later.

I'd go with 1 unless SDK churn makes it premature, then 2. Fine with 3 too if you prefer. WDYT?

georgeh0: 1 sounds good :)

Note that Rust SDK isn't very stable now (it's not released), but since the benchmark seed logic will be simple, it wouldn't be hard to keep it in sync in case the SDK will have any changes.

Thanks!

junzh0u: Implemented option 1 — the fixture is now a noop E2E app on the Rust SDK, seeded through the engine's own commit path; the bench moved to the SDK crate (SDK depends on `core`, so the reverse would be a dev-dependency cycle). All numbers re-measured on the new fixture and the three PR bodies updated.

One residue: #2306's names-off shape needs a `bench-support`-gated core hook that deletes `TargetSegmentName` entries after seeding — a targeted deletion through the typed keyspace, not record simulation.

georgeh0 (approved):
Thanks!

PR #2306, opened 2026-07-20
feat: persist readable names for provider-only target-state path segments

### LINKED ISSUE #2300: [FEATURE] Persist readable names for provider-only target-state path segments

junzh0u:
Closes #2300.

`cocoindex show` resolves fingerprinted target-state path segments from persisted records (owner index + tracking items) and, for live apps, the provider registry. Two classes of provider-only segments had no persisted mapping and always rendered as `#hex`: root providers when inspecting via `--db`/`--app-name` (registry empty without the module), and attachments of nested providers (build-local registry, discarded after the run).

This PR persists segment names in a new `TargetSegmentName` keyspace:

- **Keyed by lone segment fingerprint**, not full prefix path — a segment's fingerprint is a pure function of its key, so one entry serves every path sharing that segment and write-once semantics are safe.
- **Written at precommit time**: `pre_commit` collects each declared target state's provider-only segments into the precommit write plan, walking the provider chain only up to the nearest ancestor backed by a declared target state (per review — backed segments resolve via their declaring component's tracking records, and that component's own pre-commit covers the chain above them, so N sibling declarations under one provider contribute no entries and no apply-time existence checks). Preview mode and `PendingRetry` stay write-free; the apply step persists missing entries (get-before-put) inside the existing batched txn.
- **Read as a fallback**: the inspect resolver consults the names only after owner-index/tracking resolution misses, kept out of `resolve_segment` itself so the `dangling` diagnostic still reflects owner/tracking consistency (test pins this).
- **No cleanup**: entries are tiny and shared across paths, per the issue's considerations. The live-registry seed remains for databases written before this feature; upgraded DBs fill in as components rebuild.

Demo — app declaring `table → vector_index → index` (root provider `demo/table_store`), inspected from a fresh process without the module. Same commands on both sides, fresh DB each time:

**Before** (parent commit of this PR):

```
$ cocoindex update ./attachment_demo_app.py
$ cocoindex show --db ./cocoindex.db --app-name AttachmentDemo --target-states
Target states:
  /#6581a8809bdcec2ded1d60b6d5319465/"my_table"
    owner:/@declare_table
  /#6581a8809bdcec2ded1d60b6d5319465/"my_table"/"row1"
    owner:/
  /#6581a8809bdcec2ded1d60b6d5319465/"my_table"/#52fbd0f1fa2fe96eba65ffd4bff4199f/"idx"
    owner:/
```

**After** (this PR):

```
$ cocoindex update ./attachment_demo_app.py
$ cocoindex show --db ./cocoindex.db --app-name AttachmentDemo --target-states
Target states:
  /@demo/table_store/"my_table"
    owner:/@declare_table
  /@demo/table_store/"my_table"/"row1"
    owner:/
  /@demo/table_store/"my_table"/@vector_index/"idx"
    owner:/
```

The two `#hex` segments are exactly the issue's two classes: the root provider (unresolvable via `--db` because the registry is empty without the module) and the `vector_index` attachment (unresolvable even with the app loaded, because its registry is build-local).

## Perf impact

Measured with #2305's benches (criterion baseline saved at the #2305 tip, compared at this branch's tip; the fixture's end-to-end seeding run persists names automatically on this branch):

```
$ cargo bench -p cocoindex --features bench-support --bench state_store_read -- --baseline sdk2305 '^(detail_per_path|list_target_states)/'
detail_per_path/1k      time:   [5.5587 ms 5.5774 ms 5.5998 ms]
                 change: time:  [-10.703% -9.8453% -8.9369%] (p = 0.00 < 0.05)
                        Performance has improved.
detail_per_path/10k     time:   [5.8169 ms 5.8489 ms 5.8861 ms]
                 change: time:  [-12.006% -11.393% -10.733%] (p = 0.00 < 0.05)
                        Performance has improved.
detail_per_path/100k    time:   [29.268 ms 29.320 ms 29.372 ms]
                 change: time:  [-19.714% -19.025% -18.218%] (p = 0.00 < 0.05)
                        Performance has improved.
list_target_states/1k   time:   [3.9012 ms 3.9147 ms 3.9325 ms]
                 change: time:  [-16.248% -13.677% -10.759%] (p = 0.00 < 0.05)
                        Performance has improved.
list_target_states/10k  time:   [40.897 ms 41.068 ms 41.245 ms]
                 change: time:  [-4.2961% -3.1976% -2.1034%] (p = 0.00 < 0.05)
                        Performance has improved.
list_target_states/100k time:   [387.00 ms 390.42 ms 393.95 ms]
                 change: time:  [-9.9264% -8.8161% -7.7064%] (p = 0.00 < 0.05)
                        Performance has improved.
```

3–19% faster across the board: persisted names replace the `#hex` fingerprint rendering, which costs more than the extra name lookup. The fallback itself is free — with names deleted after seeding, this branch's resolver measures the same as the #2305 baseline (`resolver_prefix_sharing/detail/fanout1`: `names_off` 4.31 ms vs baseline 4.33 ms), while `names_on` drops it to 3.60 ms.

The second commit adds the `names on/off` axis so this stays measurable: the engine persists names during the fixture's E2E seeding run (names-on comes free), and a small `bench-support`-gated core hook prefix-deletes them afterwards for the names-off shape — a targeted deletion through the typed keyspace, not record simulation.

### DIFF
--- a/rust/core/src/engine/execution.rs
+++ b/rust/core/src/engine/execution.rs
@@ -792,13 +792,12 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
     // backed provider contribute no entries here (and no existence-check
     // reads at apply time). Deduped by fingerprint; the apply step skips
     // already-persisted entries.
-    let segment_names: Vec<(Fingerprint, StableKey)> = {
+    let segment_names: HashMap<Fingerprint, StableKey> = {
         let guard = declared_target_states.lock().await;
-        let mut seen: HashSet<Fingerprint> = HashSet::new();
-        let mut names = Vec::new();
+        let mut names = HashMap::new();
         for decl in guard.values() {
             decl.provider
-                .collect_provider_only_segment_names(&mut seen, &mut names);
+                .collect_provider_only_segment_names(&mut names);
         }
         names
     };
--- a/rust/core/src/engine/execution.rs
+++ b/rust/core/src/engine/execution.rs
@@ -781,6 +781,29 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
     if pending_retry {
         return Ok(PreCommitOutcome::PendingRetry);
     }
+
+    // Readable names for the provider segments (root → … → immediate
+    // provider) of every declared target-state path. Provider-only segments
+    // (root providers, attachments) have no owner-index/tracking record, so
+    // these write-once name entries are the only persisted source inspection
+    // can resolve them from (see `DbEntryKey::TargetSegmentName`). Deduped
+    // here by fingerprint; the apply step skips already-persisted entries.
+    let segment_names: Vec<(Fingerprint, StableKey)> = {
+        let guard = declared_target_states.lock().await;
+        let mut seen: HashSet<Fingerprint> = HashSet::new();
+        let mut names = Vec::new();
+        for decl in guard.values() {
+            let provider_path = decl.provider.target_state_path().as_slice();
+            for (fp, key) in std::iter::zip(provider_path.iter(), decl.provider.stable_key_chain())
+            {
+                if seen.insert(*fp) {
+                    names.push((*fp, key));
+                }
+            }
+        }
+        names
+    };
+
     let mut modified_old_owners: HashSet<StablePath> = HashSet::new();
     let previously_exists = tracking_info.is_some();
     if let Some(tracking_info) = &mut tracking_info {
@@ -1174,6 +1197,7 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
             self_path: stable_path.clone(),
             new_tracking_info: new_tracking_info_bytes,
             preempted_owner_updates,
+            segment_names,
         },
     })
 }
--- a/rust/core/src/engine/execution.rs
+++ b/rust/core/src/engine/execution.rs
@@ -782,24 +782,23 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
         return Ok(PreCommitOutcome::PendingRetry);
     }
 
-    // Readable names for the provider segments (root → … → immediate
-    // provider) of every declared target-state path. Provider-only segments
-    // (root providers, attachments) have no owner-index/tracking record, so
-    // these write-once name entries are the only persisted source inspection
-    // can resolve them from (see `DbEntryKey::TargetSegmentName`). Deduped
-    // here by fingerprint; the apply step skips already-persisted entries.
+    // Readable names for the provider-only segments (root providers,
+    // attachments) reachable from every declared target-state path. Such
+    // segments have no owner-index/tracking record, so these write-once name
+    // entries are the only persisted source inspection can resolve them from
+    // (see `DbEntryKey::TargetSegmentName`). Each chain walk stops at the
+    // first target-state-backed ancestor — its declaring component covers
+    // itself and everything above — so N sibling declarations under the same
+    // backed provider contribute no entries here (and no existence-check
+    // reads at apply time). Deduped by fingerprint; the apply step skips
+    // already-persisted entries.
     let segment_names: Vec<(Fingerprint, StableKey)> = {
         let guard = declared_target_states.lock().await;
         let mut seen: HashSet<Fingerprint> = HashSet::new();
         let mut names = Vec::new();
         for decl in guard.values() {
-            let provider_path = decl.provider.target_state_path().as_slice();
-            for (fp, key) in std::iter::zip(provider_path.iter(), decl.provider.stable_key_chain())
-            {
-                if seen.insert(*fp) {
-                    names.push((*fp, key));
-                }
-            }
+            decl.provider
+                .collect_provider_only_segment_names(&mut seen, &mut names);
         }
         names
     };
--- a/rust/core/src/engine/target_state.rs
+++ b/rust/core/src/engine/target_state.rs
@@ -10,7 +10,7 @@ use crate::{
 
 use cocoindex_utils::batching::{BatchQueue, Batcher, BatchingOptions, Runner};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     hash::{Hash, Hasher},
 };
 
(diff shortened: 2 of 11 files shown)


rust/core/src/engine/execution.rs
```diff
@@ -781,6 +781,29 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
...
+        let mut seen: HashSet<Fingerprint> = HashSet::new();
+        let mut names = Vec::new();
+        for decl in guard.values() {
+            let provider_path = decl.provider.target_state_path().as_slice();
+            for (fp, key) in std::iter::zip(provider_path.iter(), decl.provider.stable_key_chain())
+            {
+                if seen.insert(*fp) {
+                    names.push((*fp, key));
+                }
+            }
```
georgeh0: if I understand it correctly, we're accumulating the provider fingerprints & stable key for each child of the provider here.

e.g. in the chain of root provider -> table -> row (N)


If each row is declared by a separate component, the root provider's fingerprint & stable key will be pushed for each of the N component. And during commit, for each component, at least a read (for existence check) will happen.

I'm thinking there may be a lighter way. e.g. we only need to iterate each level of parent **until** the provider comes from a target state.

e.g. 
* For "row", for the row, its provider (table) comes from a target state for the table, so we can stop.
* in "table", its provider (the root provider) doesn't come from a target state, so we can push it (the root provider).
* For attachment cases, "root provider -> table -> attachment -> a specific attachment", the specific attachment's parent doesn't come from a target state, so we can push it (the attachment), but the attachment's parent provider (the table) come from a target state, so we can stop here.

WDYT?

junzh0u: Good call — done in cc0914a6. `pre_commit` now walks each declared state's provider chain only until the first ancestor backed by a declared target state: that segment resolves via its declaring component's owner-index/tracking records, and that component's own pre-commit covers the chain above it. Matches your examples: a row pushes nothing, a table pushes just the root, an attachment item pushes the attachment segment then stops at the table. An already-seen fingerprint is skipped but the walk continues, since the same segment key (e.g. an attachment type) can appear under different parents.

This needed a new `backed_by_target_state` flag on the provider, set only in `register_lazy` — by pre-commit time a lazily-created provider is otherwise indistinguishable from a root/attachment one (handler already fulfilled), and "backed by a target state" is a cross-component property. Alternatives considered:

- Checking the component-local declarations map: a component declaring only rows doesn't have the table in its map, so it would walk past it and re-push table + root names — the same N× cost.
- Consulting the persisted owner index: a DB read per segment per commit, the cost we're removing.
- Recording names at provider-registration sites (no flag, no walk): attachment providers are registered in `fulfill_handler` *after* the declaring component's commit, so their names would reliably miss the write plan.

Perf: no re-measure needed for the PR body's numbers — the read path is untouched, and the entries this change stops writing are exactly the ones the resolver never consulted (owner/tracking resolution wins before the name fallback); the commit path strictly does less work. Confirmed with a back-to-back A/B of this tip vs 7c2612da on the read benches: no regressions (detail_per_path 1k/10k statistically unchanged, the other cases at noise level or slightly favorable).

rust/core/src/engine/target_state.rs
```diff
@@ -291,6 +297,40 @@ impl<Prof: EngineProfile> TargetStateProvider<Prof> {
...
+    /// backed by a declared target state: that segment resolves via its
+    /// declaring component's owner-index/tracking records, and that
+    /// component's own pre-commit covers the segments above it. `seen`
+    /// dedups across calls; a seen fingerprint is skipped but the walk
+    /// continues, since providers at different depths can share a segment
+    /// (e.g. the same attachment type on two tables).
+    pub fn collect_provider_only_segment_names(
+        &self,
+        seen: &mut HashSet<utils::fingerprint::Fingerprint>,
+        out: &mut Vec<(utils::fingerprint::Fingerprint, StableKey)>,
```
georgeh0: Consider directly taking a `&mut HashMap<utils::fingerprint::Fingerprint, StableKey>`: cleaner interface, and callers only need to construct a single container (and probably can directly use `HashMap` on instead of `Vec` down the road in `execution.rs` etc.).

BTW, when using `HashMap`, the [entry API](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#entry-api) is usually best for "insert if not exist".

rust/core/src/engine/target_state.rs
```diff
@@ -291,6 +297,40 @@ impl<Prof: EngineProfile> TargetStateProvider<Prof> {
...
 
+    /// Collect segment-name entries (lone segment fingerprint → stable key)
+    /// for this provider and its ancestors, stopping at the first ancestor
+    /// backed by a declared target state: that segment resolves via its
+    /// declaring component's owner-index/tracking records, and that
+    /// component's own pre-commit covers the segments above it. `seen`
+    /// dedups across calls; a seen fingerprint is skipped but the walk
+    /// continues, since providers at different depths can share a segment
+    /// (e.g. the same attachment type on two tables).
+    pub fn collect_provider_only_segment_names(
```
georgeh0: This method is a little bit ad-hoc.  Consider restrict its visibility by `pub(crate)`.

georgeh0 (approved):

rust/core/src/engine/target_state.rs
junzh0u: Done in 54d174fd — the method now takes a single `&mut HashMap<Fingerprint, StableKey>` and dedups via the entry API (`or_insert_with`, so the clone only happens on first insertion). Also took the "down the road" part: `execution.rs` builds the `HashMap` directly and `PrecommitWritePlan.segment_names` carries it as-is through to the apply loop, so the `Vec` is gone entirely. Iteration order of the plan entries is now unspecified, which is fine for these write-once puts.

rust/core/src/engine/target_state.rs
junzh0u: Done in 54d174fd — now `pub(crate)`.

georgeh0 (approved):
Great, thanks!

PR #2318, opened 2026-07-26
fix(core): clear inverted target-owner index on component deletion

suju-droid:
## Summary

Whole-component deletion removed the component's `__track` row but left every `__target` owner row it held behind. Those rows point at a component whose tracking info no longer exists, so `iter_target_states` reports them as `dangling` and nothing ever reclaims them.

`build_commit_writes` short-circuited Delete mode to `(None, vec![])`:

```rust
if self.component_ctx.mode() == ComponentProcessingMode::Delete {
    return Ok((None, Vec::new()));   // <- no owner rows collected
}
```

`target_owners_to_delete` is the only mechanism that removes rows from the inverted index, and it was always empty on delete. There is no sweep or GC elsewhere that collects them (the Phase 5 GC handles child components via tombstones, not the owner index), so the leak is permanent for paths that are never re-declared.

## Root cause / execution trace

1. A component owning target-state paths is orphaned (source key removed) or dropped.
2. `Committer::commit` calls `build_commit_writes`, which returns `(None, vec![])` for `Delete`.
3. `AppStore::commit` sees `new_tracking_info == None` and deletes the `__track` row, then iterates an empty `target_owners_to_delete`.
4. The `__target` rows survive, now pointing at a component with no tracking info — precisely the state `TargetStateEntry::dangling` documents as "left behind by ... an interrupted cleanup".

## Fix

Delete mode now reads the component's own tracking info and collects the target-state paths it still owns (deduped across provider ids), so the owner rows are dropped in the same commit txn as the `__track` row.

The delete in `AppStore::commit` is guarded on the row still pointing at the deleting component. A component that preempted the path between our last run and this delete has already upserted itself as owner via `precommit_claim_targets`; deleting unconditionally would orphan that live claim. This ordering is reachable because the owner index is claimed during precommit while the delete plan is built from a component-scoped read.

## Behavior

| | before | after |
|---|---|---|
| `__target` row after component delete | retained, `dangling: true` | removed |
| path preempted by another component | n/a (never deleted) | preserved for the new owner |

Reclaiming a leaked path was already safe (`old_tracking_cache` misses → `prev_item = None` → `prev_may_be_missing = true`, the conservative branch), so this is an orphaned-state/unbounded-growth fix rather than a data-corruption one. It also removes spurious `[dangling]` entries from `cocoindex show --target-states`.

## Test plan

- [x] New e2e test `test_delete_dict_clears_target_owner_index` in `python/tests/core/test_component_target_states.py`, reusing the existing `_declare_dicts_data_together` fixture: deletes one dict component while the app lives on, asserts its owner rows are gone and none are `dangling`. **Fails before the fix** (`assert not any(entry.dangling ...)` → `assert not True`), passes after.
- [x] New `rust/core/tests/target_owner_cleanup.rs` covers the preemption guard — asserts a delete does not drop a row another component has claimed. Fails without the guard.
- [x] `cargo test -p cocoindex_core` — 84 passed, 0 failed
- [x] `uv run pytest python/tests/core/` — 520 passed, 1 skipped
- [x] `uv run pytest python/tests/cli/ python/tests/internal/` — 97 passed
- [x] `uv run mypy` — clean (278 files); `cargo fmt --check` clean; `ruff format/check` clean on touched files

### DIFF
--- a/python/tests/core/test_component_target_states.py
+++ b/python/tests/core/test_component_target_states.py
@@ -164,6 +164,39 @@ def test_dicts_data_together_delete_dict() -> None:
     ]
 
 
+@pytest.mark.asyncio
+async def test_delete_dict_clears_target_owner_index() -> None:
+    """Deleting a component drops its rows from the inverted owner index.
+
+    The `__track` row goes away with the component, so any `__target` owner
+    row it left behind is dangling and never reclaimed.
+    """
+    DictsTarget.store.clear()
+    _source_data.clear()
+
+    app = coco.App(
+        coco.AppConfig(
+            name="test_delete_dict_clears_target_owner_index", environment=coco_env
+        ),
+        _declare_dicts_data_together,
+    )
+
+    _source_data["D1"] = {"a": 1, "b": 2}
+    _source_data["D2"] = {"c": 3}
+    await app.update()
+
+    owners_before = [entry async for entry in coco_inspect.iter_target_states(app)]
+    assert any("D1" in entry.readable_path for entry in owners_before), owners_before
+    assert not any(entry.dangling for entry in owners_before)
+
+    del _source_data["D1"]
+    await app.update()
+
+    owners_after = [entry async for entry in coco_inspect.iter_target_states(app)]
+    assert not any("D1" in entry.readable_path for entry in owners_after), owners_after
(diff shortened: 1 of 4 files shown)


georgeh0 (approved):
Thanks a lot for the fix!

Issue #2263, opened 2026-07-07
coco.map documentation


badmonster0:
suggestion from @prrao87  - op: https://discord.com/channels/1314801574169673738/1345477421972852826/1521535129305219093

I think there's a gap in the documentation specifically with regard to how CocoIndex uses coco.map. I checked the v1 code base and there is a coco.map decorator that makes it convenient to map execution across async executors, and then collect the results. Sort of like a built-in "map-reduce API", without being an explicit map-reduce API in and of itself. 

The following pattern that maps a section formatting function across async executors and combines them into a single formatted document is very clean and natural - it fits seamlessly into a python developer's workflow. 
```
# Format each document concurrently within the current component.
formatted_sections = await coco.map(format_section, document.sections)

# Combine the results with ordinary Python.
result = "\n\n".join(formatted_sections)
```

I think it deserves some love in the documentation, it's so clean and Pythonic!
Right now, searching for "map" in the search box of the docs doesn't produce any results in the docs page. However, it does lead to the "Examples" section of the docs, and the main branch of CocoIndex v1 does clearly indicate that coco.map exists (meaning that it was documented and used at some point in the past), so I think this could be documented to let the user know how they can apply it this way.

rchiang:
I'll be happy to take this one. Seems like at least one good way to get started.

badmonster0:
assigned thanks ray! @rchiang

rchiang:
Sorry for the delay. It took me a while to figure out what was even going on. There is already documentation for the map() method in the section called "Processing helpers". What gets weird is what the search index finds vs. doesn't find. I couldn't figure out a way to use the search bar locally in my dev environment, so I had to test on the live website.

- Searching "map" returns results focuses solely on LiveMap
- Searching "coco.map" returns various places where that exact term is, but not the code block in "Processing helpers"
- Searching for phrases within the "Processing helpers" like "when to use map" or "map applies an async function" finds the correct header or sentence in the section, but the ranking isn't always high.

The documentation in that section seems fine, but with the period potentially being treated as a word separator, the LiveMap term often having higher ranking, and some coco.map calls not being found/indexed, I'm not sure I have a suggestion for how this is best fixed.

prrao87:
I wonder if the lack of search capability is a result of the algolia plugin settings? That's what caused me to ask for map() docs (because I couldn't find it when I searched 😅)

rchiang:
[...]

PR #2325, opened 2026-07-27
docs: Add error handling to "Processing helpers" section (#2263)

rchiang:
Added section on error handling. Also adjusted the header and some of the text in the map() section to improve searchability.

### DIFF
--- a/docs/src/content/docs/programming_guide/processing_component.mdx
+++ b/docs/src/content/docs/programming_guide/processing_component.mdx
@@ -57,7 +57,7 @@ And two sugar APIs that simplify common patterns:
 - **`mount_each()`** — mounts one component per item in a keyed iterable
 - **`mount_target()`** — mounts a target without an explicit subpath
 
-See also [`map()`](#map) for a utility API that operates within a component without creating new ones.
+See also [`map()`](#Using CocoIndex map to apply transformations) for a utility API that operates within a component without creating new ones.
 
 ### Automatic subpath derivation
 `mount()`, `use_mount()`, and `mount_each()` all accept an optional `ComponentSubpath` as their first argument. When omitted, the subpath is **auto-derived** from the function name using `Symbol(fn.__name__)`.
@@ -288,8 +288,8 @@ This pattern ensures that CocoIndex can track component relationships and target
 
 ## Processing helpers
 
-### `map()`
-`map()` applies an async function to each item in a collection, running all calls concurrently within the current processing component. Unlike [`mount()`](#mount) and [`mount_each()`](#mount_each), it does **not** create child processing components — it's purely concurrent execution (similar to `asyncio.gather()`).
+### Using CocoIndex `map()` to apply transformations
+The `map()` handler (also known as `coco.map()`) applies an async function to each item in a collection, running all calls concurrently within the current processing component. Unlike [`mount()`](#mount) and [`mount_each()`](#mount_each), it does **not** create child processing components — it's purely concurrent execution (similar to `asyncio.gather()`).
 
 ```python
 @coco.fn(memo=True)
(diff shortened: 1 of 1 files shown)


badmonster0:

rchiang:
> @georgeh0 can help take a look!

Weird. This was supposed to be tied to https://github.com/cocoindex-io/cocoindex/issues/2263 (which has some additional suggestions), so I'm not sure how it ended up as a separate pull request.

docs/src/content/docs/programming_guide/processing_component.mdx
```diff
@@ -288,8 +288,8 @@ This pattern ensures that CocoIndex can track component relationships and target
 
 ## Processing helpers
 
-### `map()`
-`map()` applies an async function to each item in a collection, running all calls concurrently within the current processing component. Unlike [`mount()`](#mount) and [`mount_each()`](#mount_each), it does **not** create child processing components — it's purely concurrent execution (similar to `asyncio.gather()`).
+### Using CocoIndex `map()` to apply transformations
+The `map()` handler (also known as `coco.map()`) applies an async function to each item in a collection, running all calls concurrently within the current processing component. Unlike [`mount()`](#mount) and [`mount_each()`](#mount_each), it does **not** create child processing components — it's purely concurrent execution (similar to `asyncio.gather()`).
```
georgeh0: I wouldn't introduce the term "handler" here, since it's just an API that users can call. A "handler" in a framework usually means something provided by users and called by the framework.

I think "API" is the proper term here.

docs/src/content/docs/programming_guide/processing_component.mdx
```diff
@@ -57,7 +57,7 @@ And two sugar APIs that simplify common patterns:
 - **`mount_each()`** — mounts one component per item in a keyed iterable
 - **`mount_target()`** — mounts a target without an explicit subpath
 
-See also [`map()`](#map) for a utility API that operates within a component without creating new ones.
+See also [`map()`](#Using CocoIndex map to apply transformations) for a utility API that operates within a component without creating new ones.
```
georgeh0: `#Using CocoIndex map to apply transformations` is not a valid anchor. You may start the docs server locally to see that.

docs/src/content/docs/programming_guide/processing_component.mdx
```diff
@@ -288,8 +288,8 @@ This pattern ensures that CocoIndex can track component relationships and target
 
 ## Processing helpers
 
-### `map()`
-`map()` applies an async function to each item in a collection, running all calls concurrently within the current processing component. Unlike [`mount()`](#mount) and [`mount_each()`](#mount_each), it does **not** create child processing components — it's purely concurrent execution (similar to `asyncio.gather()`).
+### Using CocoIndex `map()` to apply transformations
```
georgeh0, line 300: Everything in the doc is about CocoIndex, so we don't need to repeat "CocoIndex" here.

IMO something like this is sufficient:

```markdown
`map()`: transform items concurrently
```

docs/src/content/docs/programming_guide/processing_component.mdx
```diff
@@ -299,9 +299,35 @@ async def process_file(file: FileLike, table: postgres.TableTarget[DocEmbedding]
     await coco.map(process_chunk, chunks, file.file_path.path, id_gen, table)
 ```
 
-The first argument to the function receives each item; additional arguments are passed through to every call. `map()` returns a `list` of the results, in the same order as the input items.
+The first argument to the function receives each item; additional arguments are passed through to every call. The `coco.map()` handler returns a `list` of the results, in the same order as the input items.
 
-#### When to use `map()` vs `mount_each()`
+#### When to use the `map()` processor vs `mount_each()`
```
georgeh0: Maybe just call it "API" instead of "processor".

docs/src/content/docs/programming_guide/processing_component.mdx
```diff
@@ -299,9 +299,35 @@ async def process_file(file: FileLike, table: postgres.TableTarget[DocEmbedding]
...
+except Exception as e:
+    error_found = e
+```
+
+To handle multiple errors, any code must use the `list[ReturnT]` return value from the `coco.map()` handler.
+
+```python
+check = await coco.map(disallow_word, TEXT)
+failures = [c for c in check if not c.ok]
+```
```
georgeh0: I'm confused when reading this example. This example doesn't define another `disallow_word` from above, so both will make `coco.map` throw an exception on an exception on any item, right?

The way to raise and handle exception is very typical for normal Python code, and there's no extra support provided by `coco.map()` itself. IMO we don't have to have a dedicated section for exception handling.

rchiang: Okay, I'll remove the section on exception handling.

docs/src/content/docs/programming_guide/processing_component.mdx
rchiang: The two main reasons I went with a more "full English phrasing" in the header.
1. Using a full phrase is a bit more common in the "user's guide" documentation style. The shortened version you are suggesting is more common in "reference" documentation style.
2. I'm not sure if it will get fixed with search index configuration, but I think having too short phrasing in the headers is part of what makes the search bar matching a bit confusing in the search previews, especially combined with "Header 3" or "Header 4" style.

PR #2328, opened 2026-07-29
fix(sqlite): preserve schema consistency on failed column migrations

suju-droid:
## Summary

Fix a correctness issue in the SQLite target connector where migration failures during `ALTER TABLE ADD COLUMN` could be silently ignored.

Previously, `_apply_column_actions()` caught every `sqlite3.OperationalError` raised during column additions. This unintentionally treated genuine migration failures (such as adding a `NOT NULL` column to a populated SQLite table) the same as idempotent duplicate-column cases.

As a result, SQLite could reject the schema change while the migration continued, allowing schema reconciliation to proceed after a failed database migration.

## Root Cause

The implementation used:

```python
except sqlite3.OperationalError:
    pass
```

for every `ALTER TABLE ADD COLUMN` operation.

This swallowed all SQLite operational errors instead of distinguishing expected duplicate-column errors from real migration failures.

## Fix

- Preserve the original DDL, including `NOT NULL` constraints.
- Catch only duplicate-column errors.
- Re-raise all other SQLite migration failures with context using exception chaining.
- Prevent schema reconciliation from continuing after an unsuccessful migration.

## Tests

Added regression coverage verifying:

- unsupported SQLite schema migrations fail loudly
- duplicate-column operations remain idempotent
- migration failures no longer proceed silently

## Why

This preserves CocoIndex's declarative schema semantics by ensuring database schema changes either succeed completely or fail explicitly, rather than silently diverging from the intended schema.

### DIFF
--- a/python/cocoindex/connectors/sqlite/_target.py
+++ b/python/cocoindex/connectors/sqlite/_target.py
@@ -372,9 +372,11 @@ class TableSchema(Generic[RowT]):
         record_info = RecordType(record_type)
         columns: dict[str, ColumnDef] = {}
 
-        for field in record_info.fields:
-            override = column_overrides.get(field.name) if column_overrides else None
-            type_info = analyze_type_info(field.type_hint)
+        for rec_field in record_info.fields:
+            override = (
+                column_overrides.get(rec_field.name) if column_overrides else None
+            )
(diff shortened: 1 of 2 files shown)


georgeh0 (approved):
Thanks for the fix!

PR #2343, opened 2026-08-06
fix(core): stage target-state ownership transfers until Phase 4 commit

Sujit-1509:
### Problem Statement

During component submission (`submit()`), execution progresses through:
- **Phase 2 (`precommit`)**: Writes `__track` with a `pending_process_token`, claims target state ownership in `__target` (`upsert_target_state_owner`), and commits the write transaction to LMDB.
- **Phase 3 (`sink.apply()`)**: Executes target actions against external connectors (Postgres, LanceDB, Qdrant, SQLite, localfs, etc.).
- **Phase 4 (`commit`)**: Finalizes component tracking info, child existence tombstones, and clears stage tokens.

If **Phase 3 (`sink.apply()`) fails** (e.g., DB connection drop, network timeout, constraint error, or Python exception inside a target handler), `submit()` catches the error and calls `clear_stage_marker` to clear `pending_process_token`.

However, `clear_stage_marker` did not revert `__target` ownership claims written to disk during `precommit`. Consequently:
1. `__target` was left mapping the target path to the failed component.
2. The prior owner's tracking record was stripped during `precommit`.
3. The failed component never completed Phase 4 commit, so its committed tracking info did not contain the target state item.
4. On subsequent reprocessing or deletion, neither component would clean up or reconcile the target state, leaving it permanently orphaned in external sinks and internal state store indices.

### Proposed Solution

This PR implements staged target state ownership:
1. **Deferred Ownership Transfers**: `AppStore::precommit` no longer updates `__target`. Ownership transfers are collected in `target_owners_to_upsert` inside `CommitPlan` and applied atomically during Phase 4 `AppStore::commit` after `sink.apply()` succeeds.
2. **Prior Owner Preservation**: When a component claims a target state from a prior owner during precommit, the prior owner's item is retained with a `Deleted` pending state entry (`item.is_pending() == true`).
3. **Clean Failure Recovery**: If `sink.apply()` fails, `__target` still points to the previous owner. On its next run, the previous owner detects `is_pending() == true`, forces `prev_may_be_missing = true`, and reconciles against external storage, healing any partial mutations.

### Testing

Added comprehensive integration tests in `rust/core/tests/staged_ownership_preemption.rs`:
- `sink_failure_preserves_previous_owner_and_prevents_uncommitted_target_claim`: Verifies `__target` remains pointing to the previous owner when `sink.apply()` fails.
- `successful_commit_transfers_target_ownership`: Verifies `__target` transfers to the new owner upon Phase 4 commit success.
- `retry_after_sink_failure_recovers_and_transfers_ownership`: Verifies retries recover and transfer ownership cleanly without dangling records.

All Rust unit/integration tests and Python test suites pass cleanly.

### DIFF
--- a/rust/core/src/engine/execution.rs
+++ b/rust/core/src/engine/execution.rs
@@ -1119,6 +1119,14 @@ async fn pre_commit<'tracking, Prof: EngineProfile>(
             }
 
             // Delete: target state is no longer declared.
+            let is_preempted = item.states.iter().any(|(v, s)| *v == 0 && s.is_deleted())
+                || bulk_target_owners
+                    .get(&target_state_path_with_pid.target_state_path)
+                    .is_some_and(|owner| owner != stable_path);
+            if is_preempted {
+                continue;
+            }
+
             let Some(target_states_provider) = target_states_providers
                 .get(target_state_path_with_pid.target_state_path.provider_path())
             else {
--- a/rust/core/src/engine/execution.rs
+++ b/rust/core/src/engine/execution.rs
@@ -395,7 +395,7 @@ impl<Prof: EngineProfile> Committer<Prof> {
         // converge — produce final bytes the AppStore will write. Per-
         // component exclusivity means no concurrent writer can change
         // `__track` between this standalone read and `app_store.commit`.
-        let (new_tracking_info, target_owners_to_delete) =
+        let (new_tracking_info, target_owners_to_upsert, target_owners_to_delete) =
             self.build_commit_writes(curr_version).await?;
 
         let child_path_set = if self.demote_component_only {
@@ -415,7 +415,7 @@ impl<Prof: EngineProfile> Committer<Prof> {
 
         let plan = CommitPlan {
             new_tracking_info,
-            target_owners_to_upsert: Vec::new(),
+            target_owners_to_upsert,
             target_owners_to_delete,
             fn_memo_clear_all_first: fn_memo_plan.clear_all_first,
             fn_memo_writes: fn_memo_plan.writes,
@@ -464,7 +464,11 @@ impl<Prof: EngineProfile> Committer<Prof> {
     async fn build_commit_writes(
         &self,
         curr_version: Option<u64>,
-    ) -> Result<(Option<Vec<u8>>, Vec<TargetStatePath>)> {
+    ) -> Result<(
+        Option<Vec<u8>>,
+        Vec<(TargetStatePath, StablePath)>,
+        Vec<TargetStatePath>,
+    )> {
         if self.component_ctx.mode() == ComponentProcessingMode::Delete {
             // Whole-component deletion also drops the inverted `__target`
             // owner index for every path this component still owns. Without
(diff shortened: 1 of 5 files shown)


georgeh0:
Thanks for the detailed writeup and for digging into the submit lifecycle! “Defer ownership transfer to commit and revert it on failure” is a natural design to consider; we considered it when building this and deliberately chose the current roll-forward model.

After adding end-to-end coverage for both failure recovery and concurrent claims, I’m going to close this PR. The orphan scenario in the problem statement does not occur, while deferring the claim to commit breaks a concurrency property the current protocol depends on.

**The failed claimant retains enough information for recovery.** Step 3 of the problem statement—“the failed component never completed Phase 4 commit, so its committed tracking info did not contain the target state item”—isn’t correct. During precommit, the same transaction that claims `__target` also durably writes the item into the new owner’s tracking in multi-state form.

That tracking record is the recovery hook. On the next run, the new owner finds the item and re-reconciles it against the retained previous states. If the claimant has instead been unmounted, Delete-mode reconciliation finds the same tracking record and cleans up the external state. #2345 now covers both paths end to end.

This is why `clear_stage_marker` intentionally clears only the stage token. Recovery here means “converge on the next run,” not “revert on failure.” Because execution may be interrupted between any two phases, next-run reconciliation is required regardless; a graceful-failure revert would be a second recovery mechanism that covers only a subset of those interruptions.

**The ownership claim must be visible before sink application.** `__target` is not only a record of completed ownership; it is also the claim record read by the concurrency protocol. Once C2 claims a path during serialized precommit, a concurrent C3 sees `__target → C2` together with C2’s live pending token and waits instead of racing C2 at the sink.

#2346 adds [a deterministic end-to-end test](https://github.com/cocoindex-io/cocoindex/pull/2346/changes#diff-fd039e4d29b23f5f4a9f7a2db39b3dccc9d4743c4dd3ecde4f07b2cd3206d48b) for this exact window: it pauses C2 inside `sink.apply`, verifies that `__target` already points to C2, and then admits C3. The test passes on `main`. Against this PR, it fails at the critical assertion because `__target` still points to C1 while C2’s sink action is already in flight.
(Note:  I mistakenly conflate #2346 with another pending change, so the PR is mixed with unrelated changes) 

Deferring the owner upsert to Phase 4 therefore reopens the stale-owner, last-write-wins race that the pending-token protocol was designed to prevent.

The useful outcome of this PR is that it exposed missing end-to-end coverage around failure and concurrency during ownership transfer. Those gaps are now closed by #2345 and #2346. Given that the reported orphan is handled by the existing roll-forward model, while the proposed timing change breaks concurrent preemption, I’m going to close the PR.

Thanks again for engaging so deeply with this subtle part of the engine.

Sujit-1509:
Thanks for the detailed explanation and for taking the time to validate this with the end-to-end tests. The distinction between __target as a committed ownership record and as the concurrency claim used by the preemption protocol makes the design tradeoff much clearer. I also understand now why the existing model intentionally relies on roll-forward reconciliation rather than compensating the precommit on sink failure.

The additional coverage in #2345 and #2346 is especially useful, and I appreciate you pointing out the incorrect assumption in my original failure scenario. I'll close this out and take the concurrency/recovery behavior into account for future changes. Thanks again for the thorough review

PR #2352, opened 2026-08-10
fix(lancedb): apply column schema evolution under full reprocess

Sujit-1509:
## Summary

Fixes a gap left by #2346 in the LanceDB connector: schema evolution is silently skipped when the same update that changes the schema runs with `full_reprocess=True`.

`full_reprocess` marks the table's previous state as maybe-missing, which turns the table's own action into an `"upsert"` (`create_table` with `if_not_exists`). #2346 fixed the *reconcile* side to still compute `column_actions` for upserts, but the *apply* side in `_TableHandler._apply_actions` still gated in-place column changes on `main_action is None` — so `add_columns` was never executed.

Because the new tracking record is committed either way, the table is then permanently stuck on the old column set: subsequent row upserts fail with `ValueError: Field 'extra' not found in target schema`, and once that error surfaces, plain updates see `prev == desired` and silently "succeed" without ever converging.

## Changes

- `python/cocoindex/connectors/lancedb/_target.py`: apply column actions for `"upsert"` main actions as well as `None`; create the table first (`if_not_exists`) so the action also converges when the table itself was missing (e.g. dropped externally). This mirrors the ordering already used by the postgres connector.
- `python/tests/connectors/test_lancedb_target.py`: regression test `test_schema_evolution_under_full_reprocess` — add a column + rows with `full_reprocess=True`, assert the column lands on the real table and a follow-up plain update is a no-op. Verified it fails on current `main` (before this fix) with the schema-mismatch error above.

## Testing

- `uv run pytest python/tests/connectors/test_lancedb_target.py` — 41 passed
- `uv run mypy` — clean
- `uv run ruff format --check` / `uv run ruff check` — clean

### DIFF
--- a/python/cocoindex/connectors/lancedb/_target.py
+++ b/python/cocoindex/connectors/lancedb/_target.py
@@ -1012,8 +1012,22 @@ class _TableHandler(coco.TargetHandler[_TableSpec, _TableTrackingRecord, _RowHan
                     continue
 
                 spec = action.spec
+
+                if action.main_action in ("insert", "upsert", "replace"):
+                    await self._create_table(
+                        conn,
+                        key.table_name,
+                        spec.table_schema,
+                        if_not_exists=(action.main_action == "upsert"),
+                    )
+
+                # "insert" / "replace" just built the table from the desired
+                # schema, so its columns already match. "upsert" may have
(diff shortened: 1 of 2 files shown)


georgeh0 (approved):
Thanks for the fix!

PR #2359, opened 2026-08-21
fix(doris): re-raise genuine column DDL failures instead of swallowing them

### LINKED ISSUE #2358: Doris connector's column-DDL reconciliation swallows every ALTER TABLE failure, permanently desyncing tracking state from actual schema
**Describe the bug**

`_apply_table_actions` in `python/cocoindex/connectors/doris/_target.py` applies non-PK column `ADD COLUMN`/`DROP COLUMN` DDL wrapped in a bare `except Exception: pass`, with no re-raise and no logging (around line 1161 on current main). `_TableHandler.reconcile()` computes and returns the tracking record from the desired schema independent of whether the DDL actually succeeded, so on the next incremental run `statediff.diff()` sees `prev == desired` and never retries. A genuine ALTER TABLE failure (a permission error, a Doris schema-change job already in progress, a syntax error) is silently accepted as success, and the table's real columns never converge on the desired schema.

The `DROP TABLE` branch two lines above in the same function does log a warning on failure (though it also does not re-raise), so the ADD/DROP COLUMN branches having neither a log line nor a re-raise reads as an oversight rather than a deliberate choice.

**To Reproduce**

Live reproduction:
1. Configure a Doris target with a schema and run once so the table exists.
2. Change the schema so a non-PK column ADD or DROP is scheduled on the next run.
3. Make that specific `ALTER TABLE ... ADD/DROP COLUMN` fail for a reason unrelated to idempotency, e.g. revoke `ALTER` privilege on the table, or trigger it while a schema-change job is already in progress on that table (Doris returns an error for this case).
4. Observe: the run completes without error, but the table's actual columns never changed. No later run retries the change, because the committed tracking record already claims the desired state was reached.

Isolated control-flow reproduction (no live Doris cluster needed, since the bug is pure control flow inside `_apply_table_actions`): build a `_TableAction` whose `column_actions` includes an `"insert"`, monkeypatch `_execute_ddl_sync` to raise a generic `Exception`, and call `_apply_table_actions`. On current main it returns normally instead of propagating the failure.

**Expected behavior**

A genuine ALTER TABLE failure during column reconciliation should propagate out of `_apply_table_actions`, so the caller does not commit a tracking record for a schema change that never happened. A benign race (e.g. the column is already in the state the action wants, on an idempotent re-run) should still be tolerated, matching how the sqlite and lancedb connectors already distinguish the two cases for their own `ADD COLUMN`.

**CocoIndex Version**

Reproduced against current `main` (`aee7b27d`).

**Additional context**

Sibling connectors already handle this: sqlite's `_apply_column_actions` re-raises a `RuntimeError` for a genuine `ADD COLUMN` failure while swallowing only `duplicate column name`, and postgres avoids the ambiguity entirely with native `ADD COLUMN IF NOT EXISTS` / `DROP COLUMN IF EXISTS`. Doris' DDL dialect does not appear to support those `IF [NOT] EXISTS` column clauses, so the fix needs the same catch-and-classify approach sqlite uses rather than a syntax change.

I have a fix ready and will open a PR shortly.

AmirF194:
Fixes #2358

**Root cause**

`_apply_table_actions` in `python/cocoindex/connectors/doris/_target.py` applied non-PK column `ADD COLUMN`/`DROP COLUMN` DDL inside a bare `except Exception: pass`. Because `_TableHandler.reconcile()` commits the tracking record from the desired schema regardless of whether the DDL succeeded, a genuine failure (a permission error, a schema-change job already in progress, a syntax error) was silently accepted as success: the next incremental run saw `prev == desired` and never retried, leaving the tracked schema permanently out of sync with the real table.

**Fix**

Extracted the per-column apply loop into `_apply_column_actions` and added `_is_benign_column_ddl_error`, which distinguishes an idempotent no-op (the column is already in the state the action wants) from a genuine failure. A benign error is still ignored; anything else is logged and re-raised, so the caller's `reconcile()` does not commit a tracking record for a change that never happened. This mirrors how the sqlite connector already handles the identical failure mode for its own `ADD COLUMN`.

Doris' DDL dialect does not appear to support `ADD COLUMN IF NOT EXISTS` / `DROP COLUMN IF EXISTS` (unlike postgres, which sidesteps this ambiguity entirely with those clauses), so the fix classifies the error message rather than changing the SQL, the same approach sqlite's connector takes for `duplicate column name`.

**Verification**

This repo's own Doris tests need a live cluster (`DORIS_FE_HOST`/`DORIS_PASSWORD`), which I don't have, so I added `python/tests/connectors/test_doris_target_ddl_swallow.py`, which exercises the real control flow without one: it calls `_apply_column_actions` directly, and an end-to-end test calls the actual `_apply_table_actions` entry point with a stubbed `_execute_ddl_sync`, `ContextProvider`, and `ManagedConnection` (no Doris connection ever opens).

- All 10 new tests fail against pre-fix `main`, the end-to-end one with `DID NOT RAISE Exception` (confirming the swallow, not just missing symbols), and pass against this branch. Ran both ways in the same container (`python:3.11`, matching this repo's own CI matrix's standard leg) against `cocoindex==1.0.20`'s compiled core with this repo's Python source overlaid on top.
- `ruff check` and `ruff format --check` on both changed files: clean, pinned to this repo's own `v0.12.0`.
- `mypy` strict over the whole `python/` tree (`uv run mypy`): clean, 281 files.
- Not verified: an actual Doris cluster returning these specific error strings for a real `ADD`/`DROP COLUMN` conflict. I don't have one available, and this repo's own connector tests are gated on `DORIS_FE_HOST` for the same reason. I did not attempt a local `maturin develop` build (this repo's real `pytest` CI hook) since a full Rust workspace build was not practical in my environment; the tests above ran against the officially published wheel's compiled core with this branch's Python source substituted in, exercising the exact code path unmodified.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,7 @@ build-test = [
 ]
 format = ["ruff"]
 type-stubs = ["types-psutil", "asyncpg-stubs"]
-ci-enabled-optional-deps = ["pydantic>=2.11.9", "asyncpg>=0.31.0", "neo4j>=5.18.0"]
+ci-enabled-optional-deps = ["pydantic>=2.11.9", "asyncpg>=0.31.0", "neo4j>=5.18.0", "pymysql>=1.1.0", "aiohttp>=3.9.0"]
 examples = ["sentence-transformers>=3.0.0", "numpy"]
 
 ci = [
--- a/python/cocoindex/connectors/doris/_target.py
+++ b/python/cocoindex/connectors/doris/_target.py
@@ -1031,6 +1031,73 @@ class _TableAction(NamedTuple):
     column_actions: dict[str, statediff.DiffAction]
 
 
+def _is_benign_column_ddl_error(e: Exception, *, expect_exists: bool) -> bool:
+    """True if `e` reports the column DDL is already a no-op for the
+    direction being applied (e.g. ADD COLUMN on a column that is already
+    there, or DROP COLUMN on one that is already gone): a benign race or
+    an idempotent re-run. False for a genuine failure, which the caller
+    must re-raise rather than silently swallow.
+    """
+    msg = str(e).lower()
+    if expect_exists:
+        return "already exist" in msg or "duplicate column" in msg
+    return "not exist" in msg or "unknown column" in msg
+
+
+def _apply_column_actions(
+    config: DorisConnectionConfig,
+    table_name: str,
+    table_schema: TableSchema[Any],
+    column_actions: dict[str, statediff.DiffAction],
+) -> None:
+    """Apply non-PK column ALTER TABLE actions for one table.
+
+    A benign no-op (per `_is_benign_column_ddl_error`) is ignored. Any other
+    DDL failure is logged and re-raised, so the caller's reconcile() does
+    not commit a tracking record for a schema change that never actually
+    happened: mirrors the sqlite connector's own handling of the same bug
+    class.
+    """
+    for sub_key, col_action in column_actions.items():
+        if not sub_key.startswith(_COL_SUBKEY_PREFIX):
+            continue
+        col_name = sub_key[len(_COL_SUBKEY_PREFIX) :]
+        if col_name in table_schema.primary_key:
+            continue
+
+        col_def = table_schema.columns.get(col_name)
+        if col_action == "delete":
+            ddl = f"ALTER TABLE `{config.database}`.`{table_name}` DROP COLUMN `{col_name}`"
+            try:
+                _execute_ddl_sync(config, ddl)
+            except Exception as e:
+                if not _is_benign_column_ddl_error(e, expect_exists=False):
+                    _logger.warning(
+                        "Failed to drop column %s from table %s: %s",
+                        col_name,
+                        table_name,
+                        e,
+                    )
+                    raise
+        elif col_action in ("insert", "upsert") and col_def is not None:
+            nullable = "NULL" if col_def.nullable else "NOT NULL"
+            ddl = (
+                f"ALTER TABLE `{config.database}`.`{table_name}` "
+                f"ADD COLUMN `{col_name}` {col_def.type} {nullable}"
+            )
+            try:
+                _execute_ddl_sync(config, ddl)
+            except Exception as e:
+                if not _is_benign_column_ddl_error(e, expect_exists=True):
+                    _logger.warning(
+                        "Failed to add column %s to table %s: %s",
+                        col_name,
+                        table_name,
+                        e,
+                    )
+                    raise
+
+
 def _apply_table_actions(
     context_provider: ContextProvider,
     actions: Sequence[_TableAction],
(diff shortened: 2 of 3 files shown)


badmonster0:
@ZhiHanZ @tomz-alt  Tom could you help take a look at this fix?

AmirF194:
Bumping this in case it slipped by, no rush. Happy to answer questions on the DDL reconciliation fix or adjust the approach.

tomz-alt:
Hi @AmirF194 , I think it is LGTM with some minor fix expected as follow up.
Verified the mechanism rather than just the diff, and it holds. The part I wanted to be sure of is that re-raising actually un-commits: sink raises → cleanup_pending_token (rust/core/src/engine/execution.rs:1814) → the item keeps its multi-state shape → next diff_composite trips len(grouped_state.prev) < len(t.prev) and re-emits. Against a build of this branch:

after rolled-back run -> column_actions {'col:col1': 'upsert'}   # retried
steady state          -> column_actions {'col:col1': None}       # no churn

Also checked the matcher against a real Doris/VeloDB cluster instead of the strings in the tests, since those are circular by construction:

- ADD duplicate → Can not add column which already exists in base table: col1 → benign ✓
- DROP missing → Column does not exists: nosuchcol → benign ✓
- missing table → Unknown table '...' → re-raised ✓

I expected "not exist" to over-match the missing-table case; it doesn't.

One thing to fix before merge: these tests skip in CI. --group ci pulls ci-enabled-optional-deps (pydantic, asyncpg, neo4j), and pymysql/aiohttp only live in the doris extra, so DEPS_AVAILABLE=False and the module never runs — the green checks here don't cover it. Skipping is normal for python/tests/connectors/, but this is the first file there that doesn't need a live cluster, so adding pymysql + aiohttp to ci-enabled-optional-deps makes them real. (aiomysql is imported lazily in connect_async, so those two are enough.)

Two follow-ups, not blockers — the same bug class in paths this PR didn't touch:

1. "replace" still falls through _apply_column_actions with no DDL at all, while reconcile() commits the new tracking record. A VARCHAR(255) → VARCHAR(1024) change yields {'col:col1': 'replace'} and emits nothing — exactly the desync described here, reached by a column type change.
2. The DROP TABLE for main_action replace/delete is still except Exception: warning(...) with no re-raise.

Worth a line in the PR body that a genuine failure now aborts the rest of the batch rather than continuing silently. That's the right trade and rollback makes it safe, but it's a user-visible change.

AmirF194:
Thanks for verifying the mechanism directly, and for catching the CI gap. Added pymysql and aiohttp to ci-enabled-optional-deps (aiomysql is only used lazily in connect_async, so those two are enough); confirmed locally that the suite skips all 10 without them and passes all 10 with them, matching what CI will now install.

Left the two follow-ups (replace falling through with no DDL, and the swallowed DROP TABLE exception) out of this PR since they're a different code path than the one this fix touches, happy to open a separate issue for those if that's useful.

badmonster0:
merged thanks @AmirF194 !

PR #2366, opened 2026-08-28
feat: recognize msgspec.Struct as a record type

yukishize:
## Summary

This PR extends CocoIndex's record type detection to include `msgspec.Struct`.

`msgspec` is already a core dependency of CocoIndex, and the serialization layer already treats `msgspec.Struct` as a natively supported type. However, `is_record_type()` currently only recognizes:

- `dataclass`
- `NamedTuple`
- Pydantic `BaseModel`

As a result, connector APIs such as `TableSchema.from_class()` / `CollectionSchema.from_class()` reject `msgspec.Struct` subclasses even though users should be able to use them as row/record types.

## Changes

- Add `is_msgspec_struct_type()` helper in `_internal/datatype.py`.
- Include `msgspec.Struct` in `is_record_type()`.
- Make `RecordType.fields` introspect `msgspec.Struct` fields via `msgspec.structs.fields`, including default values.
- Update connector docstrings and error messages to mention `msgspec.Struct`.
- Update connector docs to mention `msgspec.Struct` as an accepted record type.
- Add unit tests covering:
  - `analyze_type_info()` recognizes `msgspec.Struct` as `RecordType`.
  - `RecordType.fields` returns correct field names, type hints, and defaults.

## Motivation

`msgspec.Struct` is already used internally (e.g. `cocoindex/connectorkits/statediff.py`) and documented as a natively supported serialization type. Allowing it as a record type makes the connector schema APIs consistent with the existing serialization support.

## Testing

Run the datatype unit tests:

```bash
uv run pytest python/tests/core/test_datatype.py -q

### CODE REVIEW

File: python/cocoindex/_internal/datatype.py [hunk 1]
```diff
@@ -158,6 +168,19 @@ def fields(self) -> Iterator[RecordFieldInfo]:
...
+        elif is_msgspec_struct_type(self.record_type):
+            for field in msgspec.structs.fields(self.record_type):
+                default = field.default
+                yield RecordFieldInfo(
+                    name=field.name,
+                    type_hint=type_hints.get(field.name, field.type),
+                    default_value=(
+                        inspect.Parameter.empty
+                        if default is msgspec.NODEFAULT
+                        else default
```
georgeh0, line 180: I think ideally we may also want to handle `default_factory`

python/cocoindex/_internal/datatype.py
```diff
@@ -158,6 +168,19 @@ def fields(self) -> Iterator[RecordFieldInfo]:
                     else inspect.Parameter.empty,
                     description=field_info.description,
                 )
+        elif is_msgspec_struct_type(self.record_type):
+            for field in msgspec.structs.fields(self.record_type):
+                default = field.default
+                yield RecordFieldInfo(
+                    name=field.name,
+                    type_hint=type_hints.get(field.name, field.type),
```
georgeh0, line 176: Now it resolves class annotations twice per call, `type_hints.get` plus `msgspec.structs.fields`' uncached annotation walk, and keeps an unreachable `field.type` fallback that diverges from the siblings' Any fallback.

georgeh0:
This is another place that may need to be updated:

https://github.com/cocoindex-io/cocoindex/blob/1a22bde0bca997ef66d50b2ff069854c768e94fb/python/cocoindex/_internal/memo_fingerprint.py#L352-L359

yukishize:
Thanks for the review,I've added `msgspec.Struct` support to `memo_fingerprint.py` so it is fingerprinted structurally instead of falling back to pickle.

### Changes

- Added `_is_msgspec_struct_instance()`
- Added `_canonicalize_msgspec()`
- Wired it into `_canonicalize()` after dataclass / Pydantic handling
- Added unit tests in `test_memo_fingerprint.py`

### Tests

- `uv run ruff check --fix python/cocoindex/_internal/memo_fingerprint.py python/tests/internal/test_memo_fingerprint.py` → **0 remaining**
- Syntax check passed
- `git diff --check` passed

I've pushed the changes in commit `7fdb9e2`. Anyone take a look?

PR #2370, opened 2026-08-29
feat(code_ast): wire tree-sitter Lua grammar

hardness1020:
## Summary

Adds the `tree-sitter-lua` grammar to `cocoindex_code_ast` and wires it into the existing `lua` language entry, so `.lua` files get syntax-aware splitting in `RecursiveSplitter` on par with the other grammar-backed languages. Previously Lua was registered with its seven extensions but `None` for `treesitter_info`, so it silently fell back to separator-based splitting: functions cut mid-body, chunks starting with no signature above them.

## Changes

- Add `tree-sitter-lua = "0.5.0"` dependency to `rust/code_ast`, exact-pinned like the other 35 grammar crates
- Attach `TreeSitterLanguageInfo` to the `lua` entry in `prog_langs.rs`; name and alias list untouched, so `detect_code_language()` is unchanged by construction
- Rust tests: `test_lua_has_treesitter` (prog_langs), `test_split_with_lua_language` and `test_split_lua_keeps_table_constructor_whole` (recursive splitter)
- Python tests: `detect_code_language` for `.lua`, plus the same two splitter tests through `RecursiveSplitter`
- Docs: add the Lua row to the syntax-aware table in `ops/text.mdx`, drop `"lua"` from the separator-fallback list, correct that list's count to 110
- Lock files: workspace `Cargo.lock` plus the 11 example locks via `dev/update_rust_extern_locks.sh`

The existing splitter tests for other languages assert only that the chunk list is non-empty, which passes identically with a grammar unwired (separator fallback also returns chunks). The new splitter tests use samples with no blank lines, so they fail if the grammar is ever unwired: one asserts every chunk is a whole `local function ... end`, one asserts a `setmetatable({ x = x, y = y }, Point)` constructor is not cut at the comma between its fields. Verified by unwiring the grammar and re-running.

## Testing

- `cargo test -p cocoindex_code_ast -p cocoindex_ops_text`: 42 + 36 passed
- `uv run pytest python/tests/ops/test_text.py`: 37 passed
- `ruff format --check .`: 283 files already formatted; `ruff check` clean on the touched file

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1978,6 +1978,7 @@ dependencies = [
  "tree-sitter-julia",
  "tree-sitter-kotlin-ng",
  "tree-sitter-language",
+ "tree-sitter-lua",
  "tree-sitter-md",
  "tree-sitter-pascal",
  "tree-sitter-php",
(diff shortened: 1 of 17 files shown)


georgeh0:
Thanks for adding Lua!

PR #2376, opened 2026-09-05
fix(cli): parse Windows app target paths

yhz5613813:
## Summary

The CLI splits app targets at the first colon. On Windows, that colon can be the drive separator, so targets such as `C:\project\main.py:app2` are parsed as module `C`.

Skip the drive-letter colon for absolute Windows paths, then locate the optional app-name separator.

## Testing

- Added parameterized coverage for a bare Windows path, an app name, and an app name with environment.
- Verified all three parser cases against the changed function.
- Ruff check, formatting check, and `compileall` on both changed files.


georgeh0 (approved):
Thanks for the fix!

PR #2381, opened 2026-09-06
fix(litellm): align embedding responses to inputs by index

hardness1020:
## Problem

`LiteLLMEmbedder._embed` mapped `response.data` back onto the input batch
positionally. LiteLLM passes the provider's `data` list through unsorted,
and each item may carry an `index`. A provider (or proxy) that returns
items out of order silently attached embeddings to the wrong texts. A count
mismatch was already rejected by the Rust batcher, so same-count reordering
was the only silent failure.

## Change

- New private `_aligned_embeddings(data, n)` in `ops/litellm.py`; `_embed`
  routes through it.
- Single pass over `data`: each item is placed into a preallocated list of
  `n` slots by its `index`. No sorting.
- `len(data) != n` raises `RuntimeError` up front, naming both counts. This
  moves the count check from the Rust batcher into Python. With the count
  fixed, a full placement is a permutation by construction, so no separate
  completeness scan is needed.
- Every item lacks `index` (missing or `None`): positional mapping, same
  vectors as before for providers that omit it.
- Any item has `index`: all must. Presence is taken from the first item;
  an item that disagrees raises `RuntimeError` ("mixes items with and
  without `index`").
- Indexed items must be `int` (not `bool`), in `0..n-1`, and unused.
  Duplicate, negative, out-of-range, or non-int indices raise `RuntimeError`
  naming the expected range and the received indices.
- Raised after the retry handler, so a malformed response is never
  converted to `RetryWithSmallerBatch` (splitting to size 1 would hide the
  provider bug). At the root call this fails the whole batch. Inside a
  sub-batch produced by an earlier `RetryWithSmallerBatch` it fails that
  sub-batch only; sibling sub-batches are unaffected.

## Not changed

- Memo `version` stays at 1. This avoids forcing a full re-embed for all
  users. Previously cached outputs that were misaligned, if any, will not
  be invalidated automatically.
- Memo key, `api_key` / `api_base` handling, and public API are untouched.

## Tests

`python/tests/ops/test_embedder_refactor.py`: helper unit tests for
positional, reorder, and each rejection case (partial index, duplicate,
count mismatch, negative, out-of-range, non-int); two tests through the
`_embed` body for realignment and for a bad index set surfacing as
`RuntimeError` rather than `RetryWithSmallerBatch`. Existing index-less
fakes are unchanged and cover the compatibility path.

### DIFF
--- a/python/cocoindex/ops/litellm.py
+++ b/python/cocoindex/ops/litellm.py
@@ -21,6 +21,7 @@ from collections.abc import Awaitable as _Awaitable
 from collections.abc import Callable as _Callable
 from typing import Any as _Any
 from typing import TypeVar as _TypeVar
+from typing import cast as _cast
 
 import litellm as litellm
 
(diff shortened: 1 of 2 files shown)


python/cocoindex/ops/litellm.py
```diff
@@ -167,6 +167,33 @@ async def _retry_litellm_call(
...
+            f"got {indices}"
+        )
+    elif any(type(i) is not int for i in indices) or sorted(indices) != list(range(n)):
+        raise RuntimeError(
+            "litellm embedding response indices are not a permutation of "
+            f"0..{n - 1}: got {indices}"
+        )
+    else:
+        ordered = sorted(data, key=lambda item: item["index"])
+    return [_np.array(item["embedding"], dtype=_np.float32) for item in ordered]
```
georgeh0: We can do the validations and sorting in a simpler and slightly more efficient way. e.g. currently we create a list with indices and traverse it multiple times, and call `sorted()` twice. 

Ideally I think we only need to traverse `data` once in a loop, extract the `index`, check it, and use the `index` value as subscription to set a target list. No `sorted()` needed.

georgeh0 (commented):
Good catch! Thanks for the PR!

python/cocoindex/ops/litellm.py
hardness1020: Done in d6ba5d9a: single loop that places each item into a preallocated list by its `index`, no `sorted()`. Duplicate / out-of-range / non-int indices are caught on placement, and the length is checked up front so a full placement is a permutation by construction. Updated PR body.

georgeh0 (approved):
Awesome! Thanks for the PR!

PR #2383, opened 2026-09-07
fix: preserve typed Python exceptions in component exception handlers

### LINKED ISSUE #2380: [FEATURE] Preserve original Python exceptions in component exception handlers
### What is the use case?

Production pipelines often need to route component failures by exception type: retry a transient `ConnectionError`, dead-letter a `ValueError`, or let a `DeadlineExceededError` propagate.

Today, a handler registered via `builder.set_exception_handler(...)` or `coco.exception_handler(...)` cannot reliably do that for component-originated failures and must instead inspect the error message.

The reason is the Rust/Python bridge. `build_on_error` in `rust/py/src/runtime.rs` formats the engine error with `format!("{err:?}")` and calls the Python callback with that string. On the Python side, `ComponentContext.resolve_exception_handler` in `python/cocoindex/_internal/component_ctx.py` reconstructs it as `RuntimeError(err_str)` before walking the handler chain.

As a result, the first handler sees `RuntimeError` for every component-originated failure, regardless of the exception the component actually raised:

```python
def handler(exc: BaseException, ctx: coco.ExceptionContext) -> None:
    type(exc)          # RuntimeError for a component-originated Python failure
    exc.__traceback__  # None; the traceback is embedded in str(exc)
```

The same bridge also flattens errors raised by the handler. `PyCallback::call()` already converts a Python exception into a `CError` backed by `HostedPyErr`, but `build_on_error` then wraps that error again with `Error::internal_msg(format!("{e:?}"))`.

This means a handler that raises `coco.DeadlineExceededError` or a user-defined exception loses its Python type before reaching `await handle.ready()`. It can also prevent the engine from preserving deadline/cancellation classification through that path.

The codebase already has the machinery to avoid this. `rust/py_utils/src/error.rs` uses `HostedPyErr` to tunnel Python exceptions through the engine, and `cerror_to_pyerr` restores them for Python callers. The `try_clone` comment describes the intended behavior: preserving the original error so it remains catchable by type with its Python traceback intact.

The `on_error` path appears to be the remaining path that still converts these errors to text.

### Describe the solution you'd like

Use the same Python exception conversion semantics for component exception handlers that foreground calls such as `use_mount` already use.

Proposed behavior:

* Python-originated component failures reach the handler as their original exception type, with the Python traceback preserved.
* Engine-native failures follow `cerror_to_pyerr`: internal errors remain `RuntimeError`, client errors become `ValueError`, and deadline/cancellation errors keep their existing Python types.
* An exception raised by a handler remains typed when propagated back through Rust, so `await handle.ready()` raises the exception the handler raised.
* `ExceptionContext` remains unchanged. No additional string field is needed because `str(exc)` provides the message and `traceback.format_exception(exc)` provides the traceback.

A minimal implementation seems possible using the existing machinery:

1. In `rust/py/src/runtime.rs`, have `build_on_error` convert the incoming `CError` with the same `cerror_to_pyerr` / `IntoPyResult` semantics used elsewhere and pass the resulting Python exception to the callback instead of `format!("{err:?}")`.
2. When the callback raises, propagate the `CError` returned by `PyCallback::call()` directly instead of converting it to `Error::internal_msg`. It already contains a `HostedPyErr` via `from_py_result()`.
3. In `python/cocoindex/_internal/component_ctx.py`, change the callback to accept `BaseException` and remove the `RuntimeError(err_str)` reconstruction. The no-handler fallback can log with `exc_info=exc`.
4. In `python/cocoindex/_internal/live_component.py`, pass the existing exception object directly from `report_exception()` instead of formatting it to text first.
5. Update the corresponding `handler_callback` annotations in `core.pyi` and document the exception semantics in `exception_handlers.mdx`.

Regression coverage could verify:

* a Python `ValueError` reaches the handler as `ValueError` with its traceback intact;
* handlers can route failures with `isinstance`;
* exceptions raised by handlers preserve their type through `handle.ready()`;
* a handler-raised `DeadlineExceededError` preserves deadline semantics through `handle.ready()`;
* engine-native failures retain the existing `cerror_to_pyerr` mapping.

The existing `test_global_exception_handler_invoked_for_background_mount` and `test_background_mount_failure_surfaces_python_traceback` tests currently encode the flattened behavior and would be updated accordingly.

### Additional context

This changes observable behavior for handlers that currently depend on the flattened representation:

* A handler checking `isinstance(exc, RuntimeError)` will see the original type for Python-originated failures.
* A handler expecting the full traceback inside `str(exc)` would instead use `traceback.format_exception(exc)` or logging with `exc_info=exc`.

Given that `HostedPyErr` already establishes typed exception preservation elsewhere in the bridge, I think treating this as a consistency/correctness fix is preferable to adding a compatibility option. Happy to adjust if maintaining access to the flattened representation is important.

I’d like to work on this if the direction sounds right.

Version checked: `main` at `3b4e54c3` (after v1.0.20).

hardness1020:
Closes #2380.

Component exception handlers now receive the component's original Python exception instead of a `RuntimeError` wrapping a formatted string, and an exception raised by a handler keeps its Python type through `await handle.ready()`.

## Changes

- `build_on_error` converts the engine error with `cerror_to_pyerr` and passes the exception object to the handler callback. A raising handler's error propagates as-is, so `handle.ready()` re-raises it with its Python type intact.
- Tunneled Python exceptions pass through unchanged, the same as `use_mount` and `handle.ready()` already do. Engine-native failures map to `ValueError` / `DeadlineExceededError` / `RuntimeError`.
- `resolve_exception_handler` and `report_exception` take `BaseException`. The no-handler fallback logs the exception summary in the message and the traceback via `exc_info`.
- Docs describe the exception semantics; `core.pyi` stubs updated.

Rebased onto main with #2385. The earlier revision's Python-side workaround for a retained exception pinning a finished component (`_CoreProcessorCtxSlot`, `del comp_ctx`, `StateHandle` holding the `ComponentContext`) is dropped: the engine now guarantees that, and `test_component_ctx_release.py` on main covers it.

## Behavior change

Handlers that relied on the flattened representation will see a difference:
- `isinstance(exc, RuntimeError)` now sees the original type for Python-originated failures.
- The traceback is no longer inside `str(exc)`; use `traceback.format_exception(exc)` or `logging(..., exc_info=exc)`.
- Engine-side context around a Python-originated failure (for example which `use_state` key failed to serialize) is no longer in the delivered exception, matching `use_mount`.

## Tests

New or updated in `python/tests/core/test_exception_handlers.py`, `test_live_component.py`, `test_use_state.py`:
- `ValueError` reaches the handler as `ValueError` with its traceback intact (build and live mode).
- Handler-raised exceptions, including `DeadlineExceededError`, preserve their type through `handle.ready()`.
- Engine-side client error reaches the handler as `ValueError`.

Verified locally: `cargo test --workspace`, `uv run mypy`, `uv run pytest python/` (1095 passed, 303 skipped), `prek run --all-files`.

### DIFF
--- a/python/cocoindex/_internal/component_ctx.py
+++ b/python/cocoindex/_internal/component_ctx.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import inspect
 import logging
@@ -192,6 +193,10 @@ class ComponentContext:
                     if inspect.isawaitable(ret):
                         await ret
                     return  # Handler swallowed → don't propagate.
+                except asyncio.CancelledError:
+                    # Cancellation is not a handler failure: never feed it
+                    # to outer handlers (which could swallow it).
+                    raise
                 except BaseException as handler_exc:
                     current_exc = handler_exc
                     source = "handler"
--- a/python/cocoindex/_internal/component_ctx.py
+++ b/python/cocoindex/_internal/component_ctx.py
@@ -132,10 +132,15 @@ class ComponentContext:
         stable_path: str,
         processor_name: str | None,
         mount_kind: MountKind,
-    ) -> Callable[[str], Awaitable[None]]:
+    ) -> Callable[[BaseException], Awaitable[None]]:
         """Build the exception-handler resolver for a child mounted under this context.
 
-        Returns a callable that takes a stringified error and:
+        Returns a callable that takes the failure as a Python exception
+        (a Python-originated failure arrives as its original exception
+        object with traceback intact; engine-native failures arrive as
+        ``RuntimeError`` / ``ValueError`` / ``DeadlineExceededError``, the
+        same mapping ``use_mount`` uses; the engine filters cancellation
+        before calling this) and:
         - walks this context's handler chain (innermost first);
         - if a handler raises, calls the next outer handler with the
           new exception;
@@ -158,18 +163,18 @@ class ComponentContext:
         # captured implicitly. `self._core_path.to_string()` in particular
         # is a PyO3 → Rust → string allocation we don't want to pay on
         # every mount.
-        async def _run(err_str: str) -> None:
+        async def _run(exc: BaseException) -> None:
             node = self._exception_handler_chain
             if node is None:
                 # No handlers registered — log directly without building
                 # the ExceptionContext metadata at all. Don't propagate.
-                _logger.error("component build failed:\n%s", err_str)
+                _logger.error("component build failed: %s", exc, exc_info=exc)
                 return
 
             env_name = self._env.name
             parent_stable_path = self._core_path.to_string()
-            original_exc: BaseException = RuntimeError(err_str)
-            current_exc: BaseException = original_exc
+            original_exc = exc
+            current_exc = exc
             source: Literal["component", "handler"] = "component"
             while node is not None:
                 ctx = ExceptionContext(
--- a/rust/py_utils/src/error.rs
+++ b/rust/py_utils/src/error.rs
@@ -85,7 +85,13 @@ impl cocoindex_utils::error::HostError for HostedPyErr {
     }
 }
 
-fn cerror_to_pyerr(err: CError) -> PyErr {
+/// Convert an engine error into the Python exception surfaced to callers.
+///
+/// Tunneled Python exceptions (`HostedPyErr`) pass through as the original
+/// object, including type, traceback and `__cause__`. Engine-native errors
+/// map to `asyncio.CancelledError` / `DeadlineExceededError` / `ValueError`
+/// (client) / `RuntimeError` (internal).
+pub fn cerror_to_pyerr(err: CError) -> PyErr {
     let inner = err.without_contexts();
     if let CError::HostLang(host_err) = inner {
         // Pass through tunneled Python errors as-is — preserves the
--- a/docs/src/content/docs/advanced_topics/exception_handlers.mdx
+++ b/docs/src/content/docs/advanced_topics/exception_handlers.mdx
@@ -144,6 +144,29 @@ ExceptionHandler = Callable[
 ]
 ```
 
+### What `exc` is
+
+Handlers receive the same exception object that a foreground call such as `await coco.use_mount(...)` would raise for the same failure:
+
+- **Python-originated failures** (your component, or a target connector, raised) arrive as the original exception. `type(exc)` is what was raised, `exc.__traceback__` is intact, and `exc.__cause__` / `exc.__context__` are preserved. Route by type with `isinstance`; render the traceback with `traceback.format_exception(exc)` or `logger.error(..., exc_info=exc)`.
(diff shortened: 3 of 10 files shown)


rust/py_utils/src/error.rs
```diff
@@ -95,7 +101,11 @@ fn cerror_to_pyerr(err: CError) -> PyErr {
         // below; return the original.
         let any: &dyn Any = host_err.as_ref();
         if let Some(hosted_py_err) = any.downcast_ref::<HostedPyErr>() {
-            return Python::attach(|py| hosted_py_err.0.clone_ref(py));
+            return Python::attach(|py| {
+                let py_err = hosted_py_err.0.clone_ref(py);
+                add_context_notes(py, &py_err, &err);
```
georgeh0: `add_context_notes` permanently mutates the caller's exception object, which `HostedPyErr::try_clone` deliberately shares (by `clone_ref`) across every replica of one failure, so engine context strings accumulate on it and can leak between unrelated recipients.

python/cocoindex/_internal/component_ctx.py
```diff
@@ -154,22 +205,28 @@ def resolve_exception_handler(
...
+        env = self._env
+        core_path = self._core_path
+
+        async def _run(exc: BaseException) -> None:
+            node = chain
             if node is None:
                 # No handlers registered — log directly without building
                 # the ExceptionContext metadata at all. Don't propagate.
-                _logger.error("component build failed:\n%s", err_str)
+                _logger.error("component build failed", exc_info=exc)
```
georgeh0: We're losing all error text here.

georgeh0 (commented):
Note that after #2385 is merged, some changes (e.g. mostly from commit 2 and 3) can be dropped

rust/py_utils/src/error.rs
hardness1020: Agreed, mutating the shared object was wrong. Removed `add_context_notes`; tunneled exceptions now pass through untouched, same as `use_mount`. Handlers no longer see the use_state key for serialization failures; test and docs updated.

python/cocoindex/_internal/component_ctx.py
hardness1020: `exc_info=exc` renders the full traceback with the default formatter, but the message field itself was bare. Now `"component build failed: %s", exc, exc_info=exc`, so the summary is in the message and the traceback renders once.

hardness1020:
Rebased onto main with #2385. Dropped the slot/StateHandle changes, the `del comp_ctx` workaround and the duplicated pinning tests; the engine guarantees that now. PR is back to the handler bridge change only.

python/cocoindex/_internal/component_ctx.py
georgeh0, line 195: I think we need to add this branch now:

```py
except asyncio.CancelledError:
  raise
```

hardness1020: Added in f4c94130. `CancelledError` raised while a handler is awaited now re-raises immediately instead of being forwarded to outer handlers. Also added `test_cancellation_in_handler_skips_outer_handlers`, which fails without the branch (outer handler received the `CancelledError`).

PR #2403, opened 2026-09-11
fix: preserve StableKey variants through MessagePack decoding

### LINKED ISSUE #2401: [BUG] StableKey MessagePack decoding conflates arrays, bytes, and strings, breaking target-state ownership
**Describe the bug**

`StableKey` values embedded in the LMDB state store's MessagePack values do not reliably retain their variant when read back. This affects component paths in `TargetStateOwnerInfo` and persisted target-segment names.

For example, `coco.component_subpath("process", (1, 2))` is serialized correctly as an array segment, but reads back as bytes: `/"process"/[1,2]` becomes `/"process"/b"\x01\x02"`. Empty arrays and arrays containing only integers in `0..=255` are affected, including nested arrays. Separately, bytes containing valid UTF-8 decode as strings, and `StableKey::Fingerprint` fails to deserialize.

Consequences verified against current `main`:

1. **Owner rows leak.** `AppStore::commit()` deletes an owner row only if its decoded component path equals the deleting component's path. Tuple and UTF-8-bytes paths fail that comparison, leaving the row behind.
2. **The ownership guard can delete another component's row.** If the current owner is `/"process"/[1,2]` and the deleting component is `/"process"/b"\x01\x02"`, decoding makes the paths compare equal. The guard incorrectly deletes the current owner's row. This was reproduced through the public storage API using the existing owner-cleanup test harness.

Additional consequences traced through current code:

- **Ownership transfer and concurrency checks use the wrong owner path.** `PrecommitSession::precommit_claim_targets()` uses it to fetch the previous owner's tracking record. `pre_commit()` uses that record for previous-state transfer and pending-operation detection. The lookup can miss the real owner or select a different component at the mis-decoded path. Full sink/concurrency scenarios still need regression tests.
- **Inspection displays the wrong owner and can mark a live owner row dangling.** Readable target-key lookup uses the mis-decoded owner path. It falls back to a fingerprint when neither tracking nor persisted segment names resolve the label; fallback is not unconditional. Persisted segment names also use the affected decoder.

Tagged storekey encoding preserves these variants, so the original tracking-record keys remain distinct. The codec defect is in MessagePack-embedded values, but its effects are not confined to display or leaked metadata because ownership decisions consume those values.

**To reproduce**

This uses the same serializer and deserialization wrapper as the state store:

```rust
use cocoindex_core::state::stable_path::StableKey;
use cocoindex_utils::deser::from_msgpack_slice;

let key = StableKey::Array(vec![StableKey::Int(1), StableKey::Int(2)].into());
let bytes = rmp_serde::to_vec_named(&key).unwrap();
assert_eq!(bytes, [0x92, 0x01, 0x02]);
let back: StableKey = from_msgpack_slice(&bytes).unwrap();
assert_eq!(back, key); // Fails: back is StableKey::Bytes([1, 2].into()).
```

Current-source diagnostics produced:

| Written value | Read value |
| --- | --- |
| `Array([])` | `Bytes([])` |
| `Array([Int(1), Int(2)])` | `Bytes([1, 2])` |
| `Array([Int(0), Int(255)])` | `Bytes([0, 255])` |
| `Array([Str("process"), Array([Int(1), Int(2)])])` | `Array([Str("process"), Bytes([1, 2])])` |
| `Bytes(b"")` | `Str("")` |
| `Bytes(b"abc")` | `Str("abc")` |
| `Fingerprint([7; 16])` | Deserialization error |

Arrays containing `-1` or `256` retain their outer array variant; nested elements must still be checked independently. Sampled null, bool, `i64::MAX`, UUID, symbol, and invalid-UTF-8 bytes round-tripped correctly.

Using the helpers in `rust/core/tests/target_owner_cleanup.rs`:

- Upsert an owner at `/"process"/[1,2]`, read it back, then call `commit_owner_deletes()` for that original path. The read path is corrupted and the row survives deletion. A bytes segment containing `abc` also leaks.
- Upsert an owner at `/"process"/[1,2]`, then call `commit_owner_deletes()` from `/"process"/b"\x01\x02"`. The current owner's row is incorrectly removed.

The original report also observed Python create → unchanged update → unmount leaving a dangling tuple-owner row after target data and tracking were removed. On the reviewed commit, the Python/CLI path was traced through source rather than separately rerun.

**Expected behavior**

Every `StableKey` variant should round-trip through the state store's MessagePack encoding unchanged. Cleanup must remove rows still owned by the deleting component and preserve rows claimed by another component, including when path segments differ only by array/bytes/string type. Ownership transfer must resolve the actual previous owner. Inspection should render tuple segments as `[1,2]` and retain bytes syntax for bytes segments.

**Execution path and root cause**

References below are against `main` at `3b4e54c3b76a3c633a878c6ea1ae9f4ef8d15ce5`:

- `python/cocoindex/_internal/api.py:396`: `mount()` builds the child path. `rust/py/src/stable_path.rs:48`: `PyStableKey::extract()` converts tuple/list input to `StableKey::Array`.
- `rust/core/src/state/stable_path.rs:18`: `Serialize` preserves native MessagePack array/bin/string shapes. At line 48, `Deserialize` uses the untagged `Repr` enum, trying `Str`, then `Bytes`, before `Array`.
- `rust/core/src/state_store/app_store.rs:672`: owner reads use `from_msgpack_slice`; owner writes at line 682 use `rmp_serde::to_vec_named`. Segment-name writes at line 710 use the same encoding.
- `rust/core/src/engine/execution.rs:518`: `build_commit_writes()` obtains target paths to delete from tracking. `rust/core/src/state_store/submit_session.rs:500`: `AppStore::commit()` applies the owner-equality guard.
- `rust/core/src/state_store/submit_session.rs:261`: `precommit_claim_targets()` reads prior owners and their tracking records. `rust/core/src/engine/execution.rs:824` and `:949`: `pre_commit()` uses them for pending-operation detection and previous-state transfer.
- `rust/core/src/inspect/db_inspect.rs:160`: `TargetKeyResolver::load_owner_items()` looks up tracking using the decoded owner. Listing at line 671 derives the dangling marker; `python/cocoindex/cli.py:922` prints the result.

Under serde's untagged buffering, an array becomes a sequence that `serde_bytes` accepts before `Array` is tried. `String` likewise accepts valid-UTF-8 bytes before `Bytes` is tried. The buffered deserializer also reports a human-readable context, causing `Fingerprint`'s `IfIsHumanReadable<Base64, Bytes>` adapter (`rust/utils/src/fingerprint.rs:39`) to expect a base64 string instead of binary bytes.

The ordering comment explicitly favors JSON bytes round-trips. JSON cannot distinguish bytes from integer arrays with the current serialization shape; MessagePack can, but this implementation discards that distinction.

No current producer of `StableKey::Fingerprint` was found, so that failure is latent. The legacy `FunctionMemoizationEntry.child_components` field also embeds stable paths, but nonempty entries are already invalidated in `rust/core/src/engine/context.rs:441`.

**Proposed direction**

Fix `StableKey::deserialize()` locally in `rust/core/src/state/stable_path.rs`. For binary formats, use a visitor driven by `deserialize_any`:

- bytes → `Bytes`; sequences → recursive `Array`; strings → `Str`;
- null/bool/integer → their existing variants, accepting unsigned integers only when representable as `i64`;
- maps → `Uuid` / `Fingerprint` / `Symbol`, decoding typed values directly through `MapAccess::next_value` so the underlying format context is preserved.

Keep `Serialize`, storekey encoding, fingerprints, transaction handling, and ownership guards unchanged. Moving `Bytes` after `Array` alone leaves the bytes/string collision and fingerprint failure unresolved.

Two compatibility decisions must be explicit before implementation:

1. **Human-readable formats.** The minimum compatibility-preserving option is to retain existing human-readable decoding and use the visitor for binary formats. A universal visitor is simpler, but makes JSON integer arrays decode as `Array` instead of `Bytes`. No production JSON decoding call site was identified, but this changes public Rust deserialization semantics and two existing tests; it is not merely a test adjustment. Serialization shape would remain unchanged.
2. **Tagged maps.** Current decoding ignores unknown fields and chooses recognized variants by enum priority, not field order. For example, a valid `uuid` plus `sym` decodes as UUID. Specify whether to preserve this acceptance or deliberately reject ambiguous/extra fields. Avoid introducing first-key-wins behavior. Cover duplicate fields, missing tags, null payloads, and malformed typed values.

**Persistence and scope**

The decoder-only fix needs no storage-schema, format, or cache-version bump: the existing serializer already writes distinguishable MessagePack values. Existing correctly encoded rows become readable without rewriting them.

It does not automatically remove rows whose tracking records were already deleted, recreate incorrectly removed owner rows, or undo prior target-state effects. Redeclaring the affected target state and subsequently deleting it can reclaim a leaked row; simply remounting an empty component is insufficient. Successful full app drop also clears the app database (`rust/core/src/engine/app.rs:265`). A repair/GC sweep is separate work, not part of this fix.

**Tests**

- `rust/core/src/state/stable_path.rs`: MessagePack round trips through `from_msgpack_slice` for all variants; empty/singleton/repeated/reordered arrays; integer boundaries `0`, `255`, `-1`, `256`, `i64::MIN/MAX`; nested and heterogeneous arrays; empty, ASCII, Unicode, NUL-containing and invalid-UTF-8 bytes; null/bool/string/UUID/symbol/fingerprint; root path versus an empty-array segment. Reject overflowing integers, floats, malformed/truncated input, and invalid tagged payloads. Include fixtures encoded with the unchanged serializer.
- `rust/core/tests/target_owner_cleanup.rs`: owner round-trip and own-row deletion for tuple and bytes paths. Extend the existing preempted-owner guard regression with tuple/bytes and bytes/string collision pairs.
- `python/tests/core/test_component_target_states.py`: create → unchanged update → unmount, checking target data, tracking paths, and `list_target_state_owners`; include memoized execution and reopening persisted state.
- `python/tests/core/test_ownership_transfer.py`: ordered transfer and blocked-sink concurrent claims using affected paths; check previous-state transfer, pending-operation handling, and preservation of the new owner's row.
- `rust/core/src/inspect/db_inspect.rs` and `python/tests/cli/test_cli.py`: correct tuple/bytes owner rendering, no false dangling markers, and persisted provider-segment names read without loading the app module.
- Keep or intentionally revise `stable_key_serde_json_shape` and `stable_path_serde_json_shape` according to the chosen JSON policy. Add tagged-map acceptance/error cases according to the chosen map policy.

The five existing stable-path tests pass on the reviewed commit. Their coverage misses this defect: storekey round trips exercise another codec, and `serde_msgpack_bytes_roundtrip` covers invalid-UTF-8 bytes and heterogeneous arrays. `test_nested_container_keys` checks target data after one update, not persisted component ownership.

**Related**

- #2218 added native MessagePack bytes handling and its regression test; the current coverage misses small-integer arrays and valid-UTF-8 bytes.
- #2318 / #2319 concern owner cleanup and its inspection assertions; affected path types need additional coverage.
- #2297 concerns CLI parsing of stable path strings, a different code path.

**CocoIndex version and verification**

Reviewed `main` at `3b4e54c3b76a3c633a878c6ea1ae9f4ef8d15ce5`. The original report referenced `459c9156`; the affected deserializer is still present. Current-source Rust diagnostics confirmed the decoding failures and both owner-guard failure modes against LMDB. No implementation changes were made during verification.

I'd like to work on the decoder fix and these regressions, with the JSON and tagged-map compatibility policy settled first.

hardness1020:
Fixes #2401.

## Problem

`StableKey::deserialize` used an untagged `Repr` enum. Serde's untagged buffering erases distinctions MessagePack already encodes:

- arrays are consumed by `serde_bytes` before `Array` is tried, so `/"process"/[1,2]` read back as `/"process"/b"\x01\x02"`
- valid-UTF-8 bytes are consumed by `String` before `Bytes`, so `b"abc"` read back as `"abc"`
- the buffer reports a human-readable context, so `Fingerprint` demands base64 and fails to deserialize at all

The state store embeds `StableKey` in its MessagePack values, so the effects were not limited to display:

- `AppStore::commit()` drops an owner row only when its decoded component path equals the deleting component's path. Tuple and UTF-8-bytes paths failed that comparison and leaked their rows.
- Two components whose paths differ only by segment type compared equal, so one could delete the other's owner row.
- Ownership transfer, the pending-operation check and inspection all resolve the previous owner through the same decoded path, so they could miss the real owner or mark a live row `[dangling]`.

## Fix

`rust/core/src/state/stable_path.rs` is the only production code change. `StableKey::deserialize` now decodes straight off the format with a visitor for every format instead of buffering. The untagged `Repr` enum is deleted, and with it the `serde_bytes` dependency.

`Serialize`, storekey encoding, fingerprints and the ownership guards are unchanged, and rows written by earlier versions stay readable, so there is no storage-schema or cache-version bump.

## Compatibility decisions

The issue asked for these two to be settled explicitly.

1. **One decode path for every format.** Round trip is guaranteed only for formats that natively distinguish array / bin / str, which is what the state store uses (MessagePack). JSON has no bytes type, so `Bytes` serializes as an integer sequence and decodes as `Array`. Nothing in production deserializes a `StableKey` from JSON, so no second set of decoding rules is kept for it. `stable_key_serde_json_shape` / `stable_path_serde_json_shape` keep their serialization assertions and now assert that decoded form explicitly.
2. **Tagged maps are strict under every format:** exactly one entry, keyed `uuid`, `fp` or `sym`. Unknown keys, extra entries, duplicate keys, empty maps and null or malformed payloads are rejected rather than resolved by enum priority (the old behaviour let a valid `uuid` plus a `sym` silently decode as a UUID). No first-key-wins fallback. The serializer only ever writes one entry, so nothing previously written is affected.

## Tests

Covers the issue's test plan across six files. The tuple and bytes segments under test are shared constants in `python/tests/common/target_states.py`.

- `rust/core/src/state/stable_path.rs`: round trips through `from_msgpack_slice` for every variant; empty/singleton/repeated/reordered arrays; integer boundaries `0`, `255`, `-1`, `256`, `i64::MIN/MAX`; nested and heterogeneous arrays; empty, ASCII, Unicode, NUL-containing and invalid-UTF-8 bytes; root path versus an empty-array segment; hard-coded wire fixtures including the issue's `[0x92, 0x01, 0x02]`, asserted against the unchanged serializer; the tagged-map accept/reject matrix; rejection of `u64` above `i64::MAX`, floats, every truncation prefix and empty input; JSON shape, including `Bytes` decoding as `Array`.
- `rust/core/tests/target_owner_cleanup.rs`: owner round trip and own-row deletion for tuple, UTF-8-bytes and empty-tuple paths; the preempted-owner guard extended over tuple/bytes, bytes/tuple, str/bytes and bytes/str pairs.
- `rust/core/src/inspect/db_inspect.rs`: tuple/bytes owner rendering, no false dangling marker, and segment-name-only resolution for paths with no owner or tracking record.
- `python/tests/core/test_component_target_states.py`: create, unchanged re-run under `@coco.fn(memo=True)`, reopening the persisted store in a fresh `Environment`, then unmount; asserts target data, tracking paths and `list_target_state_owners`.
- `python/tests/core/test_ownership_transfer.py`: ordered transfer between tuple-segment and bytes-segment components, asserting the previous state reaches the new owner as a single upsert carrying `prev`. The blocked-sink concurrent-claim test is parametrized two ways: over string segments and over tuple/bytes segments (where the reclaimer's segment is exactly the bytes a collapsing codec would produce from the owner's tuple), and over resuming in the same update versus exhausting the pending-transfer retries. The exhaustion variant is what proves precommit observed the in-flight owner and refused the takeover; it also checks the next update completes the transfer with no further errors.
- `python/tests/cli/test_cli.py` plus a new `python/tests/cli/typed_key_app.py`: `show --target-states` renders `[1,2]` and `b"abc"` with their owners, no `#fingerprint` and no `[dangling]`, both with the app module loaded and from the database alone via `--db/--app-name`. The flat in-memory store both CLI fixture apps use moved to `python/tests/cli/target_store.py` so it accepts non-string keys.

Every new test was checked to fail against the previous untagged decoder and pass with the visitor. The one exception is the malformed-input rejection test, which holds on both paths; it covers input validation the issue asked for rather than the regression itself.

## Verification

`prek run --all-files` passes in full, which includes `cargo test`, `cargo check` for the Rust examples and benchmarks, `mypy`, `ruff format` and the whole `pytest` suite.

## Not in scope

The decoder fix does not recreate owner rows that were already incorrectly removed, nor remove rows already orphaned by a previous run. As the issue notes, a repair or GC sweep is separate work.

### DIFF
--- a/rust/core/src/state/stable_path.rs
+++ b/rust/core/src/state/stable_path.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
 use std::{fmt::Write as FmtWrite, io::Write};
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -45,8 +45,96 @@ impl Serialize for StableKey {
     }
 }
 
+/// Decodes one `StableKey` straight off a self-describing format, keeping the
+/// array / bin / str distinction the format already carries.
+struct StableKeyVisitor;
+
+impl<'de> de::Visitor<'de> for StableKeyVisitor {
+    type Value = StableKey;
+
+    fn expecting(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("a stable key")
+    }
+
+    fn visit_unit<E: de::Error>(self) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Null)
+    }
+
+    fn visit_bool<E: de::Error>(self, v: bool) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Bool(v))
+    }
+
+    fn visit_i64<E: de::Error>(self, v: i64) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Int(v))
+    }
+
+    fn visit_u64<E: de::Error>(self, v: u64) -> std::result::Result<StableKey, E> {
+        i64::try_from(v)
+            .map(StableKey::Int)
+            .map_err(|_| de::Error::custom(format!("integer {v} out of range for i64")))
+    }
+
+    fn visit_str<E: de::Error>(self, v: &str) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Str(Arc::from(v)))
+    }
+
+    fn visit_bytes<E: de::Error>(self, v: &[u8]) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Bytes(Arc::from(v)))
+    }
+
+    fn visit_seq<A: de::SeqAccess<'de>>(
+        self,
+        mut seq: A,
+    ) -> std::result::Result<StableKey, A::Error> {
+        let mut items = Vec::new();
+        while let Some(item) = seq.next_element::<StableKey>()? {
+            items.push(item);
+        }
+        Ok(StableKey::Array(Arc::from(items)))
+    }
+
+    /// Tagged variants are written as a one-entry map. Values are read through
+    /// `next_value` so `Uuid`/`Fingerprint` see the real format context instead
+    /// of a buffer that claims to be human-readable.
+    fn visit_map<A: de::MapAccess<'de>>(
+        self,
+        mut map: A,
+    ) -> std::result::Result<StableKey, A::Error> {
+        const TAGS: &[&str] = &["uuid", "fp", "sym"];
+        let Some(tag) = map.next_key::<String>()? else {
+            return Err(de::Error::invalid_length(0, &"a one-entry tagged map"));
+        };
+        let key = match tag.as_str() {
+            "uuid" => StableKey::Uuid(map.next_value()?),
+            "fp" => StableKey::Fingerprint(map.next_value()?),
+            "sym" => StableKey::Symbol(Arc::from(map.next_value::<String>()?)),
+            other => return Err(de::Error::unknown_field(other, TAGS)),
+        };
+        if map.next_key::<de::IgnoredAny>()?.is_some() {
+            return Err(de::Error::custom(
+                "tagged stable key must have exactly one entry",
+            ));
+        }
+        Ok(key)
+    }
+}
+
 impl<'de> Deserialize<'de> for StableKey {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> std::result::Result<Self, D::Error> {
+        // Binary formats (msgpack, the state-store encoding) distinguish
+        // array / bin / str natively, so decode directly off the format.
+        // Untagged buffering erases that: a sequence is taken by `serde_bytes`
+        // before `Array` is tried, valid UTF-8 bytes are taken by `String`
+        // before `Bytes`, and the buffer reports a human-readable context that
+        // makes `Fingerprint` demand base64. Human-readable formats keep the
+        // untagged path below, where those ambiguities are unavoidable anyway.
+        // This check needs the original format context: an outer untagged or
+        // flatten adapter can lose it before reaching this method. Persisted
+        // stable keys must therefore be deserialized without such buffering.
+        if !deserializer.is_human_readable() {
+            return deserializer.deserialize_any(StableKeyVisitor);
+        }
+
         #[derive(Deserialize)]
         #[serde(untagged)]
         enum Repr {
@@ -54,11 +142,10 @@ impl<'de> Deserialize<'de> for StableKey {
             Bool(bool),
             Int(i64),
             Str(String),
-            // Intentionally before Array to preserve StableKey::Bytes roundtrip in formats
-            // (like JSON) where bytes can be represented as a sequence of integers.
-            // `serde_bytes` makes this consume a native byte string too (msgpack
-            // `bin`, the state-store format) — a plain `Vec<u8>` only deserializes
-            // via `deserialize_seq`, so a `bin` would fail to match any variant.
+            // Intentionally before Array to preserve StableKey::Bytes roundtrip in
+            // human-readable formats (like JSON) where bytes serialize as a sequence
+            // of integers. Binary formats never reach this enum; see the
+            // `is_human_readable` branch above.
             #[serde(with = "serde_bytes")]
             Bytes(Vec<u8>),
             Uuid {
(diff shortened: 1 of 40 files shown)


rust/core/src/state/stable_path.rs
```diff
@@ -45,20 +45,107 @@ impl Serialize for StableKey {
...
+        // before `Array` is tried, valid UTF-8 bytes are taken by `String`
+        // before `Bytes`, and the buffer reports a human-readable context that
+        // makes `Fingerprint` demand base64. Human-readable formats keep the
+        // untagged path below, where those ambiguities are unavoidable anyway.
+        // This check needs the original format context: an outer untagged or
+        // flatten adapter can lose it before reaching this method. Persisted
+        // stable keys must therefore be deserialized without such buffering.
+        if !deserializer.is_human_readable() {
+            return deserializer.deserialize_any(StableKeyVisitor);
+        }
```
georgeh0: I think we want to go through the same deserialize path with the new visitor regardless of human readable or not.

Note that the following claim

> Human-readable formats keep the untagged path below, *where those ambiguities are unavoidable anyway*

isn't accurate. Whether or not ambiguities are unavoidable for a format depends on the specific format itself, not simply human-readable or not.

In reality rountrip will work for a subset of formats (like msgpack) and we'll use these formats storage. For remaining ones, we don't expect to call the deserializer, so we also don't need to complicate the implementation by another branch with different deserializing rules.

hardness1020: Done in d0d9ba8b. `deserialize` now always goes through `StableKeyVisitor`; the `is_human_readable()` branch, the untagged `Repr` enum and the `serde_bytes` dependency are removed. The two JSON shape tests keep their serialization assertions and now assert explicitly that `Bytes` decodes as `Array` there; the JSON lenient tagged-map tests are dropped and their uncovered cases folded into the msgpack strict matrix. 

Also corrected the comment: round trip is a property of the specific format (msgpack distinguishes array/bin/str), not of human-readable vs not. PR description updated to match.

rust/core/src/state/stable_path.rs
```diff
@@ -45,56 +45,90 @@ impl Serialize for StableKey {
...
+
+    fn visit_bytes<E: de::Error>(self, v: &[u8]) -> std::result::Result<StableKey, E> {
+        Ok(StableKey::Bytes(Arc::from(v)))
+    }
+
+    fn visit_seq<A: de::SeqAccess<'de>>(
+        self,
+        mut seq: A,
+    ) -> std::result::Result<StableKey, A::Error> {
+        let mut items = Vec::with_capacity(seq.size_hint().unwrap_or(0));
```
georgeh0: Let's use `Vec::new()` here, since `Vec::with_capacity()` may allocate a huge buffer on corrupted header.

georgeh0 (approved):

rust/core/src/state/stable_path.rs
hardness1020: Done in 12ace7a1.

georgeh0:
Thanks a lot for the fix!

PR #2413, opened 2026-09-13
feat(connectors): add Apache Pulsar target connector

maohsiah:
Part of #2125 — target connector only; the source connector remains follow-up work.

## Summary

- add `cocoindex[pulsar]` with the bounded `pulsar-client>=3.13.0,<4` dependency
- add public APIs for declaring and mounting keyed, user-managed Pulsar topic targets
- fingerprint message values so identical app updates produce no message
- emit native null-value Pulsar tombstones by default, with optional custom deletion payloads
- persist canonical value fingerprints and future deletion payloads in independent flat root states
- run sequential client/topic groups in one worker operation, with one producer per group
- defer cancellation until the active native worker closes its producer, without closing the caller's client
- document setup, ownership, string-key requirements, deletion behavior, and public APIs

## Design notes

The current `pulsar.asyncio.Producer.send()` path schema-encodes its payload and rejects `None`. The synchronous `pulsar.Producer.send(None, ...)` path was verified against Pulsar 4.0.13 to emit a real null-valued message (`Message.has_null_value() == True`). Therefore the target uses the synchronous producer inside one `asyncio.to_thread()` call per action batch. Producer creation, ordered sends, and close all happen in that worker operation; the caller-owned `pulsar.Client` remains open.

The topic handler is a handle-only namespace using the current child-slot target API. Each real message is independently tracked by `(ContextKey name, topic, message key)` through the private flat root provider `cocoindex/pulsar/message`. This allows parent declaration removal and `App.drop()` to reconstruct and publish deletions after the topic child slot disappears.

Desired strings and bytes are compared as canonical UTF-8 wire bytes. The deletion callback is evaluated on every desired reconciliation and its normalized result is persisted. Changing only that policy commits metadata without publishing a duplicate value. Conflicting persisted policies fail before broker I/O; uncertain empty history emits the policy-independent null tombstone.

Publication is at least once. SDK failures surface to CocoIndex, tracking remains uncommitted, and an application rerun replays the full batch, including messages already acknowledged before the failure. Groups run sequentially in first-seen order; ordering is promised only within a topic.

Cancellation is also cleanup-safe: it remains the primary caller outcome, but
the target waits for the already-started synchronous worker and producer close.
The SDK producer send timeout and the caller-configured client operation
timeout bound their respective interrupted native calls.

## Verification

Pinned broker:

```text
Image: apachepulsar/pulsar:4.0.13
Digest/Image ID: sha256:f923d4ccad4c79b5de9a2c88f47f73c8128223e8e11a95a3cbd81229c4f4dd9e
Architecture/OS: arm64/linux
Persistent mount: cocoindex-pulsar-2125-data:/pulsar/data
AutoRemove: false
Health command: docker --context colima-cocoindex-2125 exec cocoindex-pulsar-2125 bin/pulsar-admin brokers healthcheck
Health result: ok
Client: pulsar-client==3.13.0
```

Live Pulsar tests were explicitly enabled; none of the 60 Pulsar tests skipped:

```sh
PULSAR_TEST_SERVER=1 \
PULSAR_SERVICE_URL=pulsar://localhost:6650 \
PULSAR_ADMIN_URL=http://localhost:8080 \
  uv run pytest python/tests/connectors/test_pulsar_target.py -q -rs
# 60 passed in 16.36s; zero skips
```

Public-App acceptance covered these broker sequences:

```text
create:v1 -> unchanged:no-message -> update:v2 -> custom-delete -> recreate:empty-non-null
parent removal -> null tombstone -> recreate on the same user-managed topic
App.drop -> null tombstone -> direct producer confirms topic remains usable
process A:create -> process B:no-message -> process C:persisted custom-delete
value -> null tombstone -> compaction -> no retained compacted key
partitioned topic: Unicode keys -> one-key update -> other-key null tombstone
broker ack 08531000 -> child crash -> full replay 08531001 -> final no-message
durable A:v1 (08261000) -> broker restart -> same ID redelivered
  -> unchanged:no-message -> A:v2 (082e1000)
```

Additional checks:

```sh
uv run mypy
# Success: no issues found in 297 source files

uv run --group format ruff check \
  python/cocoindex/connectors/pulsar \
  python/tests/connectors/test_pulsar_target.py
# All checks passed!

uv lock --check
# Resolved 208 packages

cd docs && npm run build
# 99 page(s) built; check-agent-output: all agent-facing artifacts OK

uv run --extra pulsar --extra kafka --extra iggy pytest \
  python/tests/connectors/test_kafka_source.py \
  python/tests/connectors/test_kafka_target.py \
  python/tests/connectors/test_iggy_source.py -q -rs
# 39 passed in 0.95s

env -u CONDA_PREFIX TESTCONTAINERS_RYUK_DISABLED=true \
PULSAR_TEST_SERVER=1 PULSAR_SERVICE_URL=pulsar://localhost:6650 \
PULSAR_ADMIN_URL=http://localhost:8080 uv run pytest python/ -q -rs
# 1172 passed, 279 skipped, 2 warnings in 100.70s

env -u CONDA_PREFIX \
TESTCONTAINERS_RYUK_DISABLED=true \
PULSAR_TEST_SERVER=1 \
PULSAR_SERVICE_URL=pulsar://localhost:6650 \
RUST_TEST_THREADS=1 \
  uv run prek run --all-files
# Every hook passed with the documented macOS/Colima compatibility settings.
```

The clean ABI3 wheel was installed and exercised through public target create
and tombstone deletion on standard CPython 3.11.15, 3.12.9, 3.13.12, and
3.14.5, always resolving `pulsar-client==3.13.0`. A separately built 3.14t
base wheel imported successfully; installing its Pulsar extra failed solely
because the native SDK has no `cp314t` wheel, as documented.

The manual disconnect probe observed `b, a, b, a`: the first send was
acknowledged, the broker pause caused a `pulsar.Timeout`, and the healthy retry
replayed the same entire batch and converged. Cancellation, repeated
cancellation, one-in-flight backpressure, and 20 repeated producer-lifecycle
cycles are covered deterministically. The persistent-broker harness also
restarted the same container and proved unacknowledged message-ID redelivery.

Tests cover reconciliation, UTF-8 string payloads, true tombstones versus empty bytes, custom deletion payloads, child-slot fulfillment, ordered batching, off-event-loop calls, cancellation, bounded sends, producer cleanup, create/send/close failures, crash replay, persistent restart, partitions, compaction, and public-App retry after a failed send.

Free-threaded Python remains supported for base CocoIndex but not for the
Pulsar extra until the native SDK publishes a compatible wheel.

With the native `pulsar` import deliberately blocked, all 53 non-live target
tests passed and only the seven explicitly broker-backed tests skipped.

## Prior work and scope

Credit to @thatditsyboy and #2161 for the earlier target design and live-testing work. This implementation was rewritten against current `main`, the current child-slot API, and the verified native-client tombstone behavior.

Not included here: Pulsar source support, topic administration, authentication helpers, a generic messaging abstraction, or Rust SDK support.


georgeh0:
Thanks for the careful work here!

One structural change before anything else. Currently messages are registered under a second root provider instead of as children of the topic, so that deletions can still be published after the topic declaration is gone. That treats the topic as still *alive* after we've stopped declaring it.

The semantics we currently support for message-broker targets (and what kafka and iggy do) is the opposite: a topic is user-managed, so when its declaration is removed, or on `App.drop()`, CocoIndex **abandons** it: it releases its claim on the topic and stops writing to it entirely, leaving whatever messages are in it as they are. Deletions are published only while the topic is alive, i.e. still declared: removing an individual message from a declared topic tombstones it as before. The reason is safety, not just consistency: once we've released a topic, writing to it again can resurrect a topic the user deleted, clobber keys another producer now owns, and can only apply a deletion policy that no longer exists. A topic that stays alive after this app stops declaring it (a shared topic, where we'd tombstone only our own keys) is a real use case, but it's a different semantics that needs engine support, and we don't support it yet.

Please match kafka's shape: `_TopicHandler._apply_actions` fulfils a per-topic `_MessageHandler` holding the resolved client (`context_provider.get(key, pulsar.Client)`) and the live `deletion_value_fn`, and `PulsarTopicTarget.declare_target_state` goes through `self._provider.target_state(key, value)`. Since the handler is now per-topic, it's also the natural place to hold one producer for the topic's lifetime rather than creating one per batch.

This is a net simplification: the persisted `deletion_payload` and its refresh path, `_MessageAction.publish`, the `ValueError` at 254, `_TopicHandle`, the now-unused `_provider`, the `isinstance` guard at 271, the `_MessageSpec` wrapper, and the (client, topic) grouping in `_apply_actions` all go away, along with the stale-payload issues that only exist because of the split. `test_parent_removal_emits_delete`, `test_app_drop_emits_delete` and their live counterparts flip to assert that nothing is published.

